PR #73, opened 2015-04-10
[JRO] Update all views with consistent markup

jayroh:
Taking the markup from the one theme#show view and placing them into the
appropriate views, individually, gets things to a little more consistent
state.


jayroh:
Merged this in after a whole bunch of fixes, revisions and polishing.

PR #74, opened 2015-04-11
[JRO] Use a single Dockerfile

jayroh:
I have a feeling this is an anti-pattern.

It works, but postgres is not my forte

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,36 @@
+FROM ruby:2.2.1
+
+ENV APP_HOME /thredded
+RUN mkdir $APP_HOME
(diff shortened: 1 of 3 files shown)


jayroh:
Questions re: the usage of a Dockerfile:

* Is this bad practice -- multiple servers/services in one dockerfile? I'd guess yes, but this feels much more straightforward than using docker-compose **for development**.
* What's the best way to update an image built by a Dockerfile? Can that be done _**IN**_ and/or _**WITH**_ the Dockerfile? 
* Or should there be a change inside a running container, then committed? If that's the case what good is the Dockerfile?

PR #75, opened 2015-04-11
[JRO] Try it out with Docker Compose

jayroh:
This reduces the size of the Dockerfile tremendously but there's a little more overhead with regards to getting things up and running.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM ruby:2.2.2
+
+RUN apt-get update -qq && \
+  apt-get install -y build-essential libpq-dev \
+  libxml2-dev libxslt1-dev nodejs libpq-dev
+
+WORKDIR /tmp
+RUN mkdir -p /tmp/lib/thredded
+ADD ./lib/thredded/version.rb /tmp/lib/thredded/
+ADD thredded.gemspec /tmp/
+ADD Gemfile /tmp/
+ADD Gemfile.lock /tmp/
+RUN bundle install
+
+ENV APP_HOME /thredded
+RUN mkdir $APP_HOME
+WORKDIR $APP_HOME
+ADD . $APP_HOME
--- a/spec/dummy/config/database.yml
+++ b/spec/dummy/config/database.yml
@@ -1,15 +1,14 @@
 defaults: &defaults
-  host: localhost
+  host: <%= ENV.fetch('THREDDED_DB_1_PORT_5432_TCP_ADDR', 'localhost') %>
+  port: <%= ENV.fetch('THREDDED_DB_1_PORT_5432_TCP_PORT', 5432) %>
   adapter: <%= ENV['DB'] || 'postgresql' %>
   encoding: utf8
   min_messages: WARNING
 
 development:
   <<: *defaults
-  host: localhost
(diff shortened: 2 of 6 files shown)


shanesveller:
Of this and #74, I would strongly lean towards this approach as more idiomatic from the Docker side. For #74, people typically use something like `supervisord` or `runit` to spool up multiple child processes in a single container.

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ruby:2.2.1
+
+ENV APP_HOME /thredded
+RUN mkdir $APP_HOME
+WORKDIR $APP_HOME
+ADD . $APP_HOME
+
+RUN apt-get update -qq && \
```
shanesveller: Typically you'll want to do steps like this one, that change very infrequently and also take a lot of time and bandwidth, earlier in the Dockerfile. In this ordering, any time you change any file in the repo, this `apt-get install` will have to be re-run. If you place it before the `ADD` line instead, the results of the `apt-get` command will be cached and re-used, until you change a preceding line or add to the list of required APT packages.

spec/dummy/config/database.yml
```diff
@@ -1,15 +1,14 @@
 defaults: &defaults
-  host: localhost
+  host: <%= ENV.fetch('THREDDED_DB_1_PORT_5432_TCP_ADDR', 'localhost') %>
```
shanesveller: This will definitely work as-is. It might still be helpful to know that the container *link* you have defined in `docker-compose.yml` above will also [create a fake hostname](https://docs.docker.com/userguide/dockerlinks/#updating-the-etchosts-file) that refers to the linked container. Within your `web` container, any running applications can connect to a hostname named `DB` which will point at the defined `db` container's IP. That feature is kind of awkward to summarize, so please let me know it's still unclear.

jayroh: So, @shanesveller, does this mean pointing to `db` here, theoretically would work?

The one snag that jumps out at me would be that, outside of using Docker, I would have to define `db` in my own host `/etc/hosts` file ... whereas here it looks for that ENV and falls back to localhost, which we know everyone has (pretty much).

Dockerfile
jayroh: You, sir, are a font of knowledge and wisdom.  Should have a revision of this up in the next few hours after I play with it a little bit

jayroh:
@shanesveller any chance I could get your :eyes: on this one more time? My gut feel when I was trying out both approaches was the same as yours - go with compose and do it the right way. After I got things smoothed out a bit, I really quite liked how simple it was to get things up and running.

:heart: Thank you, dude :beers:

spec/dummy/config/database.yml
shanesveller: This is a great point, so your solution is probably preferable here. :100:

shanesveller:
Belated :+1: from me!

TheDudeWithTheThing:
and I"ll :+1:  too to make @jayroh feel awesome

jayroh:
Thank you Shane! :beers:

And thank YOU, Shaun ... for all your hard work! :eggplant:

PR #83, opened 2015-05-03
[JRO] Add placeholder avatars for online users

jayroh:
Empty image src tags look a little off so might as well put some fake
stuff in there to flesh it out a little more.


kylefiedler:
:+1:

PR #98, opened 2015-06-05
[JRO] Not finding messageboard is not Exception'al

jayroh:
Current situation:

1. you call `messageboard` from a view
2. `def messageboard` looks for it via the slug in the params hash
3. it has no param, or that messageboard passed in via param does not exist
4. it says "whoops. doesn't exist. I'm raising an exception"

This causes an issue when you're at the root of the thredded engine and your layout uses some code that's looking for a messageboard ... resulting in an infinite redirect loop.

Bummer.

This PR changes that behavior to only return nil on that scope and let people deal with it accordingly where it is being called.

### DIFF
--- a/app/commands/thredded/ensure_role_exists.rb
+++ b/app/commands/thredded/ensure_role_exists.rb
@@ -1,6 +1,6 @@
 module Thredded
   class EnsureRoleExists
-    def initialize(params={})
+    def initialize(params = {})
       @user = params.fetch(:user)
       @messageboard = params.fetch(:messageboard)
     end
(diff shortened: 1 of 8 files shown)


kylefiedler:
Looks fine to me. It's still failing though…

jayroh:
@kylefiedler do you mean failing on the tests or failing while browsing around, period?

kylefiedler:
@jayroh I meant the CI tests

jayroh:
Gotcha. I'll get those situated before I jump into anything new.

PR #99, opened 2015-06-08
[JRO] Active public and private topic nav(s)

jayroh:
* When you are on the *public* topics list the navigation should be visually "active".
* When you are on the *private* topics list the navigation should be visually "active".

Up until now only the public topics link was active.

* Also only show the topic navigation links if there is a `messageboard`
  object available. When you are on the main messageboard listing you
  should not be able to see those two links.

* * *

Added another commit that will make someone a member of a messageboard, if they are not already, when they create a new topic, private or public.

### DIFF
--- a/spec/dummy/app/themes/default/assets/stylesheets/layout/_topic-navigation.scss
+++ b/spec/dummy/app/themes/default/assets/stylesheets/layout/_topic-navigation.scss
@@ -24,9 +24,17 @@
 
 }
 
-.topic-navigation--active {
+%topic-navigation--active {
   border-bottom: 1px solid $action-color;
   margin-bottom: -2px;
+}
+
+#thredded_topics_index .topic-navigation--public {
+  @extend %topic-navigation--active;
+}
+
+#thredded_private_topics_index .topic-navigation--private {
+  @extend %topic-navigation--active;
 }
 
 .topic-navigation--unread {
--- a/spec/dummy/app/themes/default/views/thredded/shared/_topic_nav.html.erb
+++ b/spec/dummy/app/themes/default/views/thredded/shared/_topic_nav.html.erb
@@ -8,17 +8,19 @@
     </li>
   </ul>
 
-  <ul class="topic-navigation">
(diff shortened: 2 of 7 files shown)


spec/dummy/app/themes/default/assets/stylesheets/layout/_topic-navigation.scss
```diff
@@ -24,11 +24,19 @@
...
   margin-bottom: -2px;
 }
 
+#thredded_topics_index .topic-navigation--public {
+  @extend %topic-navigation--active;
+}
+
+#thredded_private_topics_index .topic-navigation--private {
+  @extend %topic-navigation--active;
+}
```
jayroh: @kylefiedler I know the desired method for this is probably to have a class applied on the view itself but what would have required some heavy'ish conditional branching in the view. Instead I reckon leaning on the page's body id is the best approach.

Is this sufficient or is there a mo betta way? School me :)

spec/dummy/app/themes/default/views/thredded/shared/_topic_nav.html.erb
```diff
@@ -8,17 +8,19 @@
...
+          <% unless unread_private_topics_count == 0 %>
+            <span class="topic-navigation--unread"><%= unread_private_topics_count %></span>
+          <% end %>
         <% end %>
-      <% end %>
-    </li>
-  </ul>
+      </li>
+    </ul>
+  <% end %>
```
jayroh: without this `if messageboard` conditional the main messageboard listing page pukes and goes belly up.

spec/dummy/app/themes/default/assets/stylesheets/layout/_topic-navigation.scss
kylefiedler: Keep it simple and just do:
```scss
#thredded_topics_index .topic-navigation--public,
#thredded_private_topics_index .topic-navigation--private {
  border-bottom: 1px solid $action-color;
  margin-bottom: -2px;
}

PR #100, opened 2015-06-09
[JRO] Toggle active users div on over/out/touch

jayroh:
https://github.com/jayroh/thredded/issues/86

Thought about just using css to toggle this div on mouseover/out but reconsidered when thinking about the mobile experience. Using javascript we can more accurately manipulate the behavior based on the various events we have available.

Looking at PPK's touch table here -- http://quirksmode.org/mobile/tableTouch.html -- I went with the best options available.

The css and how it looks needs some work still but this will do in a pinch and as a starting point.

### DIFF
--- a/spec/dummy/app/themes/default/assets/stylesheets/components/_currently-online.scss
+++ b/spec/dummy/app/themes/default/assets/stylesheets/components/_currently-online.scss
@@ -1,8 +1,17 @@
 .currently-online {
-  @include position(fixed, null 0 0 null);
+  @include position(fixed, null 0 -1.25rem null);
   background-color: $base-border-color;
   padding: $base-spacing;
   width: rem(260);
+
+  &.is-expanded {
+    @include position(fixed, null 0 0 null);
+
+    .currently-online--users {
+      height: auto;
+      visibility: visible;
+    }
+  }
 }
 
 .currently-online--title {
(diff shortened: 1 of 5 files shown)


spec/dummy/app/themes/default/assets/stylesheets/components/_currently-online.scss
```diff
@@ -1,8 +1,17 @@
...
   width: rem(260);
+
+  &.is-expanded {
+    @include position(fixed, null 0 0 null);
+
+    .currently-online--users {
+      height: auto;
+      visibility: visible;
+    }
+  }
```
jayroh: I know this is ugly and I know this could use a little TLC in the form of some transitions or something but ... derp.

spec/dummy/app/themes/default/views/layouts/thredded.html.erb
```diff
@@ -23,5 +23,9 @@
 
     <%= javascript_include_tag 'thredded' %>
     <%= yield :thredded_javascript %>
+    <script>
+      thredded = (typeof thredded === 'undefined') ? new Thredded : thredded;
+      thredded.currentlyOnline.init();
```
jayroh: if `thredded` isn't defined, instantiate it and call the currently online method init.

Issue #95, opened 2015-05-31
Breadcrumbs aren't working


kylefiedler:
Right now they are just for show but they should probably work.

PR #104, opened 2015-06-12
[JRO] Get breadcrumbs nice and situated. re #95

jayroh:
Breadcrumbs are easy in this theme. There are two states:

1. From the top-level, the messageboards listing
2. Inside a messageboard itself - you see a link to the top-level
   messageboard listing, or a link to the topics listing for the
   messageboard.

That's it.

Abstracting it out into partials and feeding them in through yield tags
allows them to be sprinkled around where they need to be.

* Add some CSS to add an active state to the "new" form pages for topics
  and private topics.

### DIFF
--- a/spec/dummy/app/themes/default/assets/stylesheets/layout/_topic-navigation.scss
+++ b/spec/dummy/app/themes/default/assets/stylesheets/layout/_topic-navigation.scss
@@ -24,7 +24,9 @@
 
 }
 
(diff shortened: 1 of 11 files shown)


jayroh:
@kylefiedler any thoughts on this approach? yielding a block of html in the view allows for customization as desired. DRY'ing it up into partials kept me from copy-pasta'ing the same html across many views.

PR #127, opened 2015-08-08
[JRO] Get rid of "themes". Stick to Rails conventions

jayroh:
re: https://github.com/jayroh/thredded/issues/123

Yeah. The "theme" idea had developed into a rube goldberg machine of
unnecessary proportions. As a result - all of the "theme" files should
all remain within the standard rails view and asset path conventions.
This commit moves all of those files and simplifies everythign that is
done.

* move `spec/dummy/app/themes/**/*` to where they need to be ... in `app/views` and `app/assets`
* Check for user class presence
* With all of these changes the readme needs updating.
* Add bourbon and es6 gems as a direct dependency of the engine. See:
  https://github.com/jayroh/thredded/issues/123#issuecomment-128777705
  @glebm makes a great point.
* The generator installs the initializer and has an option to also
  install the views and assets.


app/assets/stylesheets/thredded.scss
```diff
@@ -0,0 +1,33 @@
...
+@import "thredded/components/form-list";
+@import "thredded/components/topic-header";
+@import "thredded/components/post";
+@import "thredded/components/post-form";
+@import "thredded/components/currently-online";
+@import "thredded/components/empty";
+@import "thredded/components/chosen";
+@import "thredded/components/flash-message";
+
+@import "thredded/components/main-section";
```
jayroh: I moved all of these sass assets to a namespaced dir.

thredded.gemspec
```diff
@@ -33,11 +33,12 @@ Gem::Specification.new do |s|
   s.add_dependency 'rinku'
   s.add_dependency 'sanitize'
   s.add_dependency 'unf'
+  s.add_dependency 'bourbon'
+  s.add_dependency 'neat'
+  s.add_dependency 'bitters'
+  s.add_dependency 'sprockets-es6'
```
jayroh: This should fix https://github.com/jayroh/thredded/issues/116 indefinitely

glebm:
Awesome! :+1:

PR #128, opened 2015-08-12
use the user path instead of the old user.to_s style link

saturnflyer:


jayroh:
sonofa ...

thank you Jim!

PR #129, opened 2015-08-16
Do not root the database

glebm:
* Connect to the database(s) as `thredded` and not `root`.
* A script that creates the database users at `script/create-db-users`.
* Moved `bin/migrate` to the more idiomatic `script/migrate`.
  Gem `bin/` are the binaries exposed by the gem.
* `script/migrate` now also invokes `rake dev:seed` and creates the test
  database from the schema dump.
* All scripts now support running on a single database when the
  `DB` environment variable is set.
* Fix Dockerfile (was not running for me at all due to missing
  docker_command.sh). Avoid conflict with the host DB by running the docker one at a different port.


docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 db:
-  image: postgres:9.4.1
+  image: postgres:9.4.4
   ports:
-    - "5432:5432"
+    - "5433:5432"
```
jayroh: Good move. This bothered me but I didn't take the opportunity to fix it at the time. Thank you for doing this!

docker-compose.yml
```diff
@@ -9,7 +9,14 @@ web:
...
   working_dir: /thredded
-  command: ./bin/docker_command.sh
+  environment:
+    DB: postgresql
+    DB_USERNAME: root
+    DB_PASSWORD:
+  command: >
+    bash -c 'script/wait-for-tcp $THREDDED_DB_1_PORT_5432_TCP_ADDR $THREDDED_DB_1_PORT_5432_TCP_PORT &&
+             script/migrate &&
+             rake dev:server'
```
jayroh: Fancy. Wait until the host and port are accepting connections before running migrate and serve? Damn fancy. Love it.

glebm: Yes. Ideally, there would be a way for a container to tell the linked containers that its services are ready, but there isn't according to this thread on docker/compose about it: https://github.com/docker/compose/issues/374.

docker-compose.yml
glebm: I didn't even realize the ports in the container could clash with the host ports until they did. Not quite like a VM :)

jayroh:
Looking awesome, @glebm. Could you squash this first? Then we'll do the business

glebm:
Squashed.

Issue #115, opened 2015-08-02
non-specific scoping in query for message board


saturnflyer:
I'm working on https://github.com/saturnflyer/purple_door/tree/forum and ran into this error where the `name` column seems to be not scoped to the appropriate table.

```
PG::UndefinedColumn: ERROR:  column "name" does not exist
LINE 1: ...thredded_roles"."messageboard_id" = $1 AND (lower(name) IN (...
                                                             ^
[...]

PR #130, opened 2015-08-16
Fix user name scoping throughout

glebm:
Also fix case-insensitive search for postgresql. Refs #115

### DIFF
--- a/app/commands/thredded/authorize_superadmin.rb
+++ b/app/commands/thredded/authorize_superadmin.rb
@@ -18,11 +18,7 @@ module Thredded
     private
 
     def user
-      @user ||= begin
-        column = Thredded.user_name_column
-        klass = Thredded.user_class
-        klass.where("#{column} = '#{username}'").first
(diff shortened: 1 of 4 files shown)


app/commands/thredded/authorize_superadmin.rb
```diff
@@ -18,11 +18,7 @@ def run
     private
 
     def user
-      @user ||= begin
-        column = Thredded.user_name_column
-        klass = Thredded.user_class
-        klass.where("#{column} = '#{username}'").first
-      end
+      @user ||= Thredded.user_class.where(Thredded.user_name_column => username).first
```
TheDudeWithTheThing: I've lately tried to avoid where.first in favor of find_by, thoughts?

glebm: No strong preference either way.

jayroh:
This looks great. Thank you @glebm

Issue #117, opened 2015-08-02
User link not generating correct path


saturnflyer:
thredded is creating a link to my user at `http://localhost:3000/users/Admin%20Admin` which sets the id to the first and last name "Admin Admin".

the code in this PR provides a scenario for this https://github.com/saturnflyer/purple_door/pull/44

saturnflyer:
I realized that I should post some context around this.
Here's what config/routes.rb looks like
```ruby
  authenticate do
    mount Thredded::Engine => '/forum', as: :forum
    delete '/users/sign_out', as: :session, to: 'devise/sessions#destroy'
  end
```
We're using the `authenticate` routing helper from devise. I had to add the `delete` line to get the sign out link in thredded to work.
But it seems that thredded _should_ be creating a link to `http://localhost:3000/forum/users/Admin%20Admin`
Even though that user id is incorrect, I'm pointing out that it doesn't have `forum` in the path to the thredded profile.
If you pull that branch at purple_door, you can create message board, but then somehow the message board is `nil` when you go to the correct thredded profile path at `http://localhost:3000/forum/users/1`

jayroh:
Great context there, sir. So ... here's the quick solution (I hope) to this.

https://github.com/jayroh/thredded/blob/master/lib/generators/thredded/install/templates/initializer.rb#L15

That config will let you customize that path. Since the engine doesn't hold the responsibility for displaying user information - delegating that to the parent application - I didn't want to scope it to inside of the engine.

That being said - was this config option a little buried? Would it have helped if it was bubbled up to the readme?

Talk to me Jim!

glebm:
BTW, the app's routes can be accessed in the lambda via the global `Rails.application.routes.url_helpers`.

saturnflyer:
Thanks @jayroh and @glebm. that was helpful. 
We have been rushing to get things together so I didn't take the time to study the initializer and just went with the defaults.

I suggest that you change the default to something like
```
Thredded.user_path = ->(user) { "/users/#{user.id}" }
```
or 
```
Thredded.user_path = ->(user) { "/users/#{user.to_param}" }
```

On thing that comes to mind with this is using the lambda in a different context.
For example if you use `instance_exec` then I would be able to use the lambda like the way it feels it should be used.
In the topics controller, for example, it uses the lambda like this
```ruby
user_path = Thredded.user_path(post.user)
```
[...]

PR #131, opened 2015-08-17
User links and timestamps

glebm:
* Execute user_path in the view context. Provide intelligent default. #117
* Move HTML rendering of user links and timestamps into partials.
  This simplifies the code and the tests, and makes customization easier.
  Not sure if I am overlooking some benefit of doing this in the decorator?

### DIFF
--- a/lib/thredded.rb
+++ b/lib/thredded.rb
@@ -42,6 +42,7 @@ module Thredded
   self.queue_inline = false
   self.email_reply_to = -> postable { "#{postable.hash_id}@#{Thredded.email_incoming_host}" }
 
+  # @return [Class] the user class from the host application.
   def self.user_class
     if @@user_class.is_a?(Class)
       fail 'Please use a string instead of a class'
@@ -56,9 +57,12 @@ module Thredded
     end
   end
 
-  def self.user_path(user)
+  # @param view_context [Object] context to execute the lambda in.
+  # @param user [Thredded.user_class]
+  # @return [String] path to the user evaluated in the specified context.
+  def self.user_path(view_context, user)
     if @@user_path.respond_to? :call
-      @@user_path.call(user)
+      view_context.instance_exec(user, &@@user_path)
     else
       '/'
     end
--- a/spec/lib/thredded_spec.rb
+++ b/spec/lib/thredded_spec.rb
@@ -46,27 +46,27 @@ describe Thredded, '.queue_inline' do
 end
 
 describe Thredded, '.user_path' do
-  after do
-    Thredded.user_path = nil
-  end
-
   it 'returns "/" if lambda is not set' do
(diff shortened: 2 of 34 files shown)


lib/thredded.rb
```diff
@@ -56,9 +57,12 @@ def self.user_class
     end
   end
 
-  def self.user_path(user)
+  # @param context [Object] context to execute the lambda in.
+  # @param user [Thredded.user_class]
+  # @return [String] path to the user evaluated in the specified context.
+  def self.user_path(context, user)
```
jayroh: Considering the several (few, but distinct) ways that the term `context` is used in this code base I'm wondering if you would consider a bit of a name shift for "context" when we talk about the view instance we're passing around. When I see "context" in isolation - like, per method - I'm not sure I completely grok what it means. What do you think about `view_instance` or `view_context` or something to that effect?

jayroh:
And if you haven't been notified by GitHub yet, Gleb - I added you to the project. It was obviously way overdue and I had no idea you *weren't* with the commit bit. In any case - thank you and welcome aboard :)

spec/lib/thredded_spec.rb
```diff
@@ -46,27 +46,27 @@
...
 
 describe Thredded, '.user_path' do
-  after do
-    Thredded.user_path = nil
-  end
-
   it 'returns "/" if lambda is not set' do
-    expect(Thredded.user_path(nil)).to eq '/'
+    Thredded.user_path = nil
+    expect(Thredded.user_path(nil, nil)).to eq '/'
```
jayroh: What do you think about adding some assignments of/for the passed params to aid in clarity of what's being passed into the method?  something like

```ruby
Thredded.user_path(_view_context=nil, _user=nil)
```

As I'm reading through these tests it's a little harder to see what parameter is what. There are some where you can infer pretty easily but sometimes it's a little more work.

Thoughts?

spec/views/thredded/user/link_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    HTML
+  end
+
+  it 'with nil user' do
+    render_partial nil
+    expect(rendered).to eq <<-HTML.strip_heredoc
+      <em>Deleted user</em>
+    HTML
+  end
+end
```
jayroh: This is flipping great. :sparkles: :star2: :sparkles:

lib/thredded.rb
glebm: Done.

spec/lib/thredded_spec.rb
glebm: Done.

glebm:
Awesome, thanks! Addressed comments and squashed the commits.

PR #135, opened 2015-08-23
A more realistic dummy app setup

glebm:
Apologies for the large PR.

This sets up the dummy app set up to render thredded in app layout and use turbolinks, exposing and fixing a number of issues:
* Some CSS selectors are too generic (`h1, ...`). Didn't fix all of these, just the most visible ones.
* JS: removed all inline script tags, setup now done in `thredded.es6`.
* Theme preview page had some issues. Fixed most.
* Seed data is now more diverse (using the Faker gem).
* User preferences test was generating all of the seed data. Fixed.
* `thredded.html.erb` did more than just a `<%= yield %>` making it impossible to use a different layout without duplicating the code from `thredded.html.erb`. Fixed by introducing a named `yield` for the content  and making the `main_app` routes available directly if the layout is not a thredded one (`standalone_layout?`). Moved `thredded.html.erb` to `layouts/thredded/application.html.erb`. Change the demo app to use thredded in the demo app's layout.
* Fixed a number of smaller issues.


glebm:
TODO: switch to ES6 imports, perform initialization in the individual modules.

app/assets/javascripts/thredded.es6
```diff
@@ -1,18 +1,59 @@
...
-class Thredded {
-  constructor() {
-    this.topicForm = new ThreddedTopicForm();
-    this.privateTopicForm = new ThreddedPrivateTopicForm();
-    this.postForm = new ThreddedPostForm();
-    this.kbShortcuts = new ThreddedKeyboardShortcuts();
-    this.timestamps = new ThreddedTimeStamps();
-    this.searchBar = new ThreddedSearchBar();
-    this.currentlyOnline = new ThreddedCurrentlyOnline();
+(function() {
```
jayroh: I've seen this approach in more than a few places but never took the time to investigate why. Now that I see it in here I did a little searching and from what I can tell it's to wrap everything in an isolated context so as not to pollute the global js namespace, is that right?

And yes, I do realize my ignorance is showing through via my obvious disdain for javascript. I really have to figure out how to get through that.

app/assets/javascripts/thredded.es6
```diff
@@ -1,18 +1,59 @@
...
+
+  const pageComponents = [
+    ['thredded-edit-post', 'postForm'],
+    ['thredded-new-private-topic', 'postForm', 'privateTopicForm'],
+    ['thredded-new-topic', 'postForm', 'topicForm'],
+    ['thredded-posts', 'postForm'],
+    ['thredded-private-topics-index', 'privateTopicForm'],
+    ['thredded-theme', 'topicForm', 'privateTopicForm', 'postForm', 'searchBar'],
+    ['thredded-topics-index', 'topicForm']
+  ];
```
jayroh: I like this a lot! Awesome approach :). I wonder, though, if it could be made a little more clear.

```javascript
const pageComponents = {
  { pageId: 'thredded-edit-post',            components: ['postForm'] },
  { pageId: 'thredded-new-private-topic',    components: ['postForm', 'privateTopicForm'] },
  { pageId: 'thredded-new-topic',            components: ['postForm', 'topicForm'] },
  { pageId: 'thredded-posts',                components: ['postForm'] },
  { pageId: 'thredded-private-topics-index', components: ['privateTopicForm'] },
  { pageId: 'thredded-theme',                components: ['topicForm', 'privateTopicForm', 'postForm', 'searchBar'] },
  { pageId: 'thredded-topics-index',         components: ['topicForm'] },
}
```

This could be splitting hairs a little bit but I did have to sit and think for a second before I realized that there was a fundamental difference between items at index 0 vs 1+ ... so it's coming from a place of communicating intent, not pedantry ;)

app/assets/stylesheets/thredded/base/_typography.scss
```diff
@@ -1,4 +1,4 @@
-body {
+.thredded--main-container {
```
jayroh: awesome

app/assets/javascripts/thredded.es6
glebm: What I would like even more is if the components are independent of the pages they are on, so the post form has a `data-thredded-post-form` attribute etc, and the initialization is performed by the component itself (`jQuery($) { $('[data-thredded-post-form]').each(node => new ThreddedPostForm(node)) }`). Is it OK if we do this after this PR is merged?

app/assets/javascripts/thredded.es6
glebm: Yes, this is a self-executing anonymous function, so that the global template is not polluted. A better approach is to use [ES6 modules](http://exploringjs.com/es6/ch_modules.html), but I haven't figured out how to make them work with sprockets-es6 yet.

app/assets/javascripts/thredded.es6
glebm: Just pushed an amended commit that fixes this, and reverts changes to schema files (not sure how that happened).

jayroh: > What I would like even more is if the components are independent of the pages they are on, so the post form has a data-thredded-post-form attribute etc, and the initialization is performed by the component itself (jQuery($) { $('[data-thredded-post-form]').each(node => new ThreddedPostForm(node)) }). Is it OK if we do this after this PR is merged?

Absolutely!

> Just pushed an amended commit that does the {pageId, components} thing, and reverts changes to schema files (not sure how that happened).

Awesome, dude. Thank you.

jayroh:
Everything looks fantastic, dude. 



Merge at your leisure

TheDudeWithTheThing:
> 787 additions and 821 deletions.

glebm:
Awesome! As I said, sorry about the size! :smile_cat:

PR #139, opened 2015-08-27
Run thredded migrations automatically

glebm:
Based on

http://blog.pivotal.io/labs/labs/leave-your-migrations-in-your-rails-engines


glebm:
Looking at the comments on the blog, this approach seems to have issues.

PR #140, opened 2015-08-28
Let JS components handle their own initialization

glebm:
Also:
* Remove unused JS components.
* Scope currently-online and post-form CSS.
* Use the same JS component for private-topic-form and topic-form.
* For the topic forms, un-expand if the form is empty and clicked
  outside the form.


app/assets/stylesheets/thredded/components/_currently-online.scss
```diff
@@ -1,4 +1,8 @@
-.currently-online {
+.thredded--currently-online {
+  @include media($mobile) {
```
glebm: Fine.

app/assets/stylesheets/thredded/components/_currently-online.scss
```diff
@@ -6,36 +10,32 @@
...
-}
+  &--users {
+    height: 0;
+    visibility: hidden;
+  }
 
-.currently-online--users {
-  height: 0;
-  visibility: hidden;
+  &.is-expanded &--users {
```
glebm: It seems to me that having the two states of the same component next to each other is more readable.

app/assets/stylesheets/thredded/components/_post-form.scss
```diff
@@ -1,20 +1,20 @@
...
   }
-}
 
-.post-form--wrapper {
-  margin-top: $base-spacing;
-  padding-top: $large-spacing;
-  border-top: $base-border;
-}
+  &--wrapper {
+    margin-top: $base-spacing;
```
glebm: Fine, bike-shedding CI.

app/assets/stylesheets/thredded/components/_currently-online.scss
jayroh: lol

jayroh:
@glebm this looks on point. Feel free to merge at your leisure, of course, but just have one question to pick your brain on. The use of `local_assigns` in the partials, is there a compelling reason to use this instead of referencing the getter method that is provided to the partial? Speed? Clarity? Combination of both? It's not a pattern I've seen *often*, rather rarely actually, but would love to know the use cases behind it for my own personal usage.

Thanks again for everything, dude :)

glebm:
If a local `a` is not provided, just calling `a` will blow up. One workaround is to do `a ||= nil` before using it, another is to use `local_assigns[:a]` directly (it's simply a hash with all the locals).

Issue #137, opened 2015-08-23
Avatar support in Who's online


glebm:
Avatars currently rely on Gravtastic that we have to put on some model, that is currently Post, meaning we can't display these where there is no Post. Need to use a gem that doesn't require associating with a model, e.g. https://github.com/Wixel/ruby-gravatar

jayroh:
Hey, I really like that! Much better, indeed

glebm:
Done.

jayroh:
🌪GLEB-NADO🌪

PR #144, opened 2015-09-02
Fix avatars #137

glebm:


lib/generators/thredded/install/templates/initializer.rb
```diff
@@ -21,7 +21,7 @@
 # User avatar URL. Thredded uses Gravatar via the gravtastic gem by default.
```
TheDudeWithTheThing: probably change / remove the comments since gravtastic is being removed.

lib/generators/thredded/install/templates/initializer.rb
```diff
@@ -21,7 +21,7 @@
...
 # Visit the gravtastic project - https://github.com/chrislloyd/gravtastic#usage
 # for further instructions on how to customize.
-Thredded.avatar_url = ->(_user, post) { post.gravatar_url(default: 'mm') }
+Thredded.avatar_url = ->(user) { Gravatar.src(user.email, 128, 'mm') }
```
TheDudeWithTheThing: should `128` be configurable somewhere?  what if I want HUGE avatars on my site??

glebm: The entire call is configurable.

TheDudeWithTheThing: whoops, missed that this is in the initializer file, carry on!

lib/generators/thredded/install/templates/initializer.rb
glebm: Done

TheDudeWithTheThing:
the return of the might @glebm !

welcome back!

PR #145, opened 2015-12-25
Correct location of template

mlaco:
`app/views/layouts/thredded.html.erb` does not exist, but
`app/views/layouts/thredded/application.html.erb` does. This change allows `rails
generate thredded:install --theme` to work.

It is possible that I did something wrong while trying to integrate Thredded into my app, and that this change is completely misguided. Thanks in advance for taking a look at this PR.

### DIFF
--- a/lib/generators/thredded/install/install_generator.rb
+++ b/lib/generators/thredded/install/install_generator.rb
@@ -23,7 +23,7 @@ module Thredded
         return unless options.theme?
 
         copy_file \
-          'app/views/layouts/thredded.html.erb',
+          'app/views/layouts/thredded/application.html.erb',
           'vendor/views/layouts/thredded.html.erb'
 
         directory \


glebm:
@mlaco Thanks!
@jayroh I don't think this should go into vendor. Merging for now anyhow.

jayroh:
> I don't think this should go into vendor. Merging for now anyhow.

I could be convinced either way. On the one hand, it's much more visible in `/app` but at the same time, this gem *is* a "vendor" provided bunch of code.

glebm:
Well, the vendor code is supposed to unmodified (copied byte-for-byte from some vendor), but as I understand this generator is for when a user wants to customize the theme.

glebm:
Also, copying all the files will make updates very difficult. Perhaps we should just describe how to override individual files so that users do this only for the files they need to customize (http://guides.rubyonrails.org/engines.html#overriding-views).

jayroh:
These are all excellent points Gleb. With that clarification you are
completely right.

mlaco:
@glebm my pleasure

PR #146, opened 2016-01-07
Removes the last instance of the old timeago

glebm:
Also, locks select2-rails at 3.5 as the newly released 4.0.1 has issues.


glebm:
Merging, as this fixes the build.

PR #148, opened 2016-01-08
[JRO] Small fixes/optimizations for 2 shell scripts

jayroh:
The create-db-users script does not take into account the fact that the postgresql daemon may be running as another user account on the system so this commit defaults it to `postgres` but allows being overridden by the PG_DAEMON_USER environment variable. This use case is semi-common if a developer is using something like [Postgres.app][] from Heroku.

Both `script/migrate` and `script/create-db-users` were logging out to STDOUT with escape codes that were not being interpreted, so this commit also adds descriptive ENV variables denoting that they will come out colored green.

`database.yml` was defaulting to the postgres port and would not work if using mysql, so pulling those into variables in the yaml file makes it a little easier to discern what the values are that are being set.

Lastly, some accompanying README edits to make things more clear (possible options, etc).

[Postgres.app]: http://postgresapp.com/

### DIFF
--- a/README.mkdn
+++ b/README.mkdn
@@ -131,10 +131,12 @@ To be more clear - these are some dependencies for when you are working on *this
 
 * PostgreSQL & MySQL - `brew install postgres mysql`
 
-Then, create the `thredded` database users for both MySQL and Postgres:
+Then, create the `thredded` database users for both MySQL and Postgres. Optional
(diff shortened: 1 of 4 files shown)


jayroh:
@glebm would love your eyes on this since shell scripting isn't exactly my tip-top forté

Thank you :)

glebm:
Looks great! :+1:

PR #143, opened 2015-09-02
Removed vendorized dependencies, add gems

glebm:
Replace vendorized JS with gems, avoiding the problem if the parent app is using some of the same JS dependencies.

Additionally:
* Switch from chosen to select2, a modern fork of chosen.
* Removed Bundler.require from the dummy application.rb,
  to catch problems with missing requires.
* Autosize post form textarea as well.

PR #150, opened 2016-03-03
Move requiring turbolinks from thredded to dummy app

cgunther:
Thredded itself does not require turbolinks (best I can see), only the dummy app used for testing required it.

Trying to use thredded on a project that does not use turbolinks itself fails when trying to require the turbolinks related libraries because they may not be installed.

#143 originally moved the require from the dummy app to thredded itself.


jayroh:
Nice @cgunther! Thanks for reporting and even more thanks for the commit to fix it. Merging it in now :)

cgunther:
Thanks for the merge and for thredded!

PR #151, opened 2016-03-04
Mark #signed_in? as a helper method in the application controller

cgunther:
The _top_nav partial relies on #signed_in? however it was not previously available to the view.

https://github.com/jayroh/thredded/blob/e2dc1e1e4a94bb7b98ea80b635e55b4aee5e5dd2/app/views/thredded/shared/_top_nav.html.erb#L8

### DIFF
--- a/app/controllers/thredded/application_controller.rb
+++ b/app/controllers/thredded/application_controller.rb
@@ -7,7 +7,8 @@ module Thredded
       :active_users,
       :messageboard,
       :preferences,
-      :unread_private_topics_count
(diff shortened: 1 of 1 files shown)


cgunther:
The build failures seem to be due to a missing SASS variable which should be unrelated to my change.

cgunther:
Rebased after 368c35c4 to solve the build failures.

glebm:
And the tests didn't catch this because Dummy::ApplicationController defines its own signed_in? helper method. Thanks! :+1:

PR #153, opened 2016-03-04
Only show the button to create a new messageboard when permitted to

cgunther:
The controller already authorizes creating a messageboard, so showing the button when the user is not permitted to create a messageboard is misleading to users.

I opted to use a fresh ability model rather than the `Thredded::Ability` model in the test so that the view could simply grant the abilities we want to test against, rather than worry about getting the view spec's state into a scenario where the ability would normally be granted (such as making the user a superadmin).

### DIFF
--- a/app/views/thredded/messageboards/index.html.erb
+++ b/app/views/thredded/messageboards/index.html.erb
@@ -12,7 +12,9 @@
     <%= render @messageboards %>
   </section>
 
-  <div class="messageboard--create">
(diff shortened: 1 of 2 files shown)


cgunther:
Rebased after 368c35c4 to solve the build failures.

glebm:
Nice! :+1:

jayroh:
Love it :). Thank you @cgunther

cgunther:
Thanks for the merge!

PR #155, opened 2016-03-07
Move SeedDatabase class out of an autoloaded path since it's for development only

cgunther:
app/commands is added to the Rails autoload path, however when eager_load is true, all files in this folder are eagerly loaded, which causes an error when factory_girl_rails is required since it likely isn't available in an environment that eager loads, like production.


glebm:
Thanks! :+1:

cgunther:
Thanks for the merge!

PR #156, opened 2016-03-07
Apply button styling to elements with the "button" class

cgunther:
368c35c4 stopped applying the button styling to elements with the "button" class.

bourbon's `$all-buttons` variable only includes native button-like elements, not any general element to be styled like a button:
https://github.com/thoughtbot/bourbon/blob/ac307d9981cce5ae4d3aadbf41e9b15503b68bb4/core/bourbon/library/_buttons.scss#L50-L57

### DIFF
--- a/app/assets/stylesheets/thredded/base/_buttons.scss
+++ b/app/assets/stylesheets/thredded/base/_buttons.scss
@@ -1,3 +1,4 @@
+.button,
 #{$all-buttons} {
   appearance: none;
   -webkit-font-smoothing: antialiased;


jayroh:
Oh hell yes. This is a good idea

cgunther:
Thanks for the quick merge!

Should have mentioned, the "button" class is used for elements like the link to create a new message board to create your first private topic.

glebm:
Thanks! I've finished the styling changes and now this works as expected.

PR #158, opened 2016-03-22
Add the ability for superadmins to delete topics

cgunther:
This adds a Delete Topic button below the form to add a new post when viewing a single topic.


glebm:
A much needed feature for public forums! We're planning on adding a moderation system in future but this is great for now! :+1:

cgunther:
Thanks for the quick merge!

Issue #138, opened 2015-08-23
Should private topics be tied to Messageboards?


glebm:
Seems like the private topics should not need a Messageboard.

jayroh:
Great question. I'll answer this with the context of someone involved in a single community (read: a single messageboard within a domain/site) amongst several other isolated messageboards. On thredded.com there was a period where there were 5 pretty active boards, with 400-500 users signed up, and about 150 active users. If private topics were open across all users, across the entire site, there would/could have been some abuse (I know, I know - abuse?) from one side to another. It's weird.

You could simulate behavior like you're talking about if you have a single messageboard on a domain, and split participation in verticals among categories.

glebm:
Ah, I see. How about this: a user can add participants only among the participants of the messageboards the user has access to.

E.g., assuming the messageboard has a `writeable_by_user(scope)` and a `users` relationship:

```ruby
users = Mesageboard.writeable_by_user(current_user).flat_map(&:users).uniq
```

jayroh:
Interesting thought! So this would, in effect, bubble everything up to the same level as the messageboards themselves. Effectively pulling it out of the context of being *within* a messageboard.

Hmmm. I could probably get behind that, actually

glebm:
Done

Issue #126, opened 2015-08-04
add support for managing permissions in the container app


saturnflyer:
We've yet to implement permissions to control the thredded forum in https://github.com/saturnflyer/purple_door/ but my expectation is that I'll need to have a lot of knowledge of the internal implementation of thredded to get my own prefrences form to hook things up.

Although I haven't tried yet, I'd love to be able to have some kind of convention to follow in my own preferences form. For example if i pass in the values `forum_access: true` (or something) that my controller or model could see that value and have a standard way to add someone to thredded.

Here's an idea of something I'd want: `Thredded.add_user(user) if params[:forum_access]`
I don't know exactly how that fits in. Perhaps `add_admin` or something might work too.

I'm not sure if the code is the best example, but the gist of what I'd want is: I've added thredded and already have a user edit screen, now what? How do I provide access to create message boards to my users?

jayroh:
@saturnflyer I see what you're saying with regards to providing an intuitive API for doing this. Question for you on a related note - would it be easier if a roles resource was exposed allowing you to set that admin, or moderator, bit?

Right now there's nothing available in terms of a UI to do that and I'm thinking it's definitely necessary

glebm:
Related: Instead of storing a roles table, it would be nice to have a more flexible way of doing this. E.g. let the user specify the scopes and the methods to check for permission. I would imagine most forums do not need per-user permissions.

jayroh:
I really like that idea. I went way too far with the authorization code in here and tried to anticipate the different combinations one might want to restrict or allow access.

Follow-up question to this would be, of course, how would that interface look? Two or three scopes for a messageboard resource - participate, moderate, admin? That could cover most of the bases for most of the use cases. Could open up a project-defined proc per each of those three that take a user and messageboard and return a bool?

glebm:
We could require 3 methods and a scope on the user:

```ruby
class User
  # Users that can read and post in the given messageboards
  scope :thredded_messageboards, -> (messageboards) { 
    # for example
    all
  }

  # @return [Array<Thredded::Messageboard>, ActiveRecord::Relation]
  #    messageboards that the user can read and post in
  def thredded_messageboards
    # for example
    Thredded::Messageboard.all
  end

  # @return [Array<Thredded::Messageboard>, ActiveRecord::Relation]
  def thredded_moderates_messageboards
    # for example 
    moderator? ? Thredded::Messageboard.all : Thredded::Messageboard.none
  end

[...]

PR #161, opened 2016-03-24
Permissions and private topics

glebm:
1. Migrates to the new permissions model. #126
2. Un-ties private topics from messageboards. #138
3. Splits out `PrivatePost` from `Post`.
4. Splits the views and mailers of private topics. They just point to
   the common topic views, but the user can now override them separately.

Not in this PR:

1. Notification preferences for private topics.
2. Permissions documentation.

Also in this PR:

1. Minore style tweaks.
2. A `rake test_all_gemfiles` task (works as advertised).

The gist of the new permissions system is in `app/models/thredded/user_permissions/`.

Sorry about a huge PR. In my defence, ripping out the previous roles system was an involved job.
@jayroh @cgunther


jayroh:
Just pulled down the branch and gave things a run through and, first comment - this is looking *phenomenal*. Really really really damn nice. Thank you @glebm for all this work. Thank you @cgunther for the work as of late. I'm super impressed with where things have gotten recently.

@glebm, I have some updates I've made locally to get `db:truncate` working (some recently added foreign keys necessitate our needing a `CASCADE` amidst our `TRUNCATE` statement), and `dev:seed` working again.  Would you prefer a pull request on top of your branch or should I just push a commit up to this branch? I ask so that I may have as good git(hub) etiquette as possible :smile: 

I haven't looked through the implementation with a fine-grained set of eyeballs yet, and will do so within the next 24 hours. A first pass, though, and I'm confident things will be ready to go sooner rather than later

glebm:
@jayroh Since this is a huge PR, a good GitHub etiquette would be a separate PR targeting this branch, but you can just push it on top of permissions in this case. :+1:

lib/thredded/seed_database.rb
```diff
@@ -51,15 +52,24 @@ def create_topics(count: 26)
...
+        end
+      end
+    end
+
+    def create_private_posts(count: (0..30))
+      @private_posts = private_topics.flat_map do |topic|
+        (count.min + rand(count.max + 1)).times do
+          FactoryGirl.create(:private_post, postable: topic, user: users.sample)
+        end
       end
```
jayroh: `create_posts` and `create_private_posts` are similar enough that they *could* be merged into a single method but the subtle differences were enough that I thought a little duplication was more beneficial than getting too clever with DRY'ing it up.

glebm: Looks good!

spec/support/features/page_object/authentication.rb
```diff
@@ -9,11 +9,8 @@ def signs_out
       visit '/users/sign_out'
     end
 
-    def signs_in_as(name)
-      ::User.where(name: name).first_or_create!(email: "#{name}@example.com")
-      visit '/sessions/new'
-      fill_in 'name', with: name
-      click_button 'Sign in'
+    def signs_in_as(user)
+      PageObject::User.new(user).log_in
```
jayroh: Love it :)

lib/generators/thredded/install/templates/initializer.rb
```diff
@@ -21,6 +21,18 @@
...
 Thredded.avatar_url = ->(user) { Gravatar.src(user.email, 128, 'mm') }
 
+# ==> Permissions Configuration
+# By default, thredded uses a simple permission model, where all the users can post to all message boards,
+# and admins and moderators are determined by a flag on the users table.
+
+# The name of the moderator flag column on the users table.
+Thredded.moderator_column = :admin
+# The name of the admin flag column on the users table.
+Thredded.admin_column = :admin
```
jayroh: I'm assuming the columns here are booleans - is that correct? Oh, looks like it. Just saw it here `20160308104204_add_admin_to_users.rb`. I'd recommend updating the comments above to say something like:

`# The name of the moderator flag column, a boolean column, on the users table.`
and
`# The name of the admin flag column, a boolean column, on the users table.`

db/schema.postgresql.rb
```diff
@@ -137,18 +151,6 @@
...
-    t.integer  "user_id"
-    t.integer  "messageboard_id"
-    t.datetime "last_seen"
-    t.datetime "created_at"
-    t.datetime "updated_at"
-  end
-
-  add_index "thredded_roles", ["messageboard_id"], name: "index_thredded_roles_on_messageboard_id", using: :btree
-  add_index "thredded_roles", ["user_id"], name: "index_thredded_roles_on_user_id", using: :btree
-
```
jayroh: This makes me happy :)

db/migrate/20160323025225_split_private_posts.rb
```diff
@@ -0,0 +1,51 @@
...
+        Thredded::PostNotification.create!(email: old_private_notification.email, post: private_post)
+      end
+      old_private_notifications.delete_all
+    end
+    old_private_posts.delete_all
+  ensure
+    # Re-enable timestamp handling
+    ActiveRecord::Base.record_timestamps = true
+    ActiveRecord::Base.send(:define_method, :touch, original_touch)
+  end
```
jayroh: gnarly

spec/spec_helper.rb
```diff
@@ -27,6 +28,7 @@
   config.before(:suite) do
     DatabaseCleaner.strategy = :transaction
     DatabaseCleaner.clean_with(:truncation)
+    TestAfterCommit.enabled = true
```
jayroh: So, I read through the gem README but wasn't *entirely* sure ... what does `test_after_commit` do for us?

lib/generators/thredded/install/templates/initializer.rb
glebm: Done

spec/spec_helper.rb
glebm: I've changed post_common.rb to use after_commit instead of after_save in one of the hooks:

```ruby
after_commit :notify_at_users, on: [:create, :update]
```

This is because `notify_at_users` uses the `id`, which is not available before commit on create, so the background job might have failed once in a while previously:

```ruby
def notify_at_users
   AtNotifierJob.queue.send_at_notifications(self.class.name, id)
end
```

However, the tests are currently run in a transaction, so the after_commit hooks do not fire in tests. This gem makes after_commit work in tests.

jayroh: Very nice! Now I get it. Thank you Gleb

TheDudeWithTheThing:
This is pretty epic, A+ Gleb!

glebm:
Merged! If there are any other concerns, I can address them later.

jayroh:
I'll be pulling down thredded.com and running it against master to see if
there are any edges that need to be smoothed over. Once I do that it should
be pretty obvious what/where needs attention.

On Sat, Mar 26, 2016 at 12:08 PM, Gleb Mazovetskiy <notifications@github.com

PR #162, opened 2016-03-26
Tweaks for better responsive UI

jayroh:
Looking at the demo app w/chrome inspector in mobile mode leaves a little to be desired. In fact, it doesn't "do responsive" at all right now. Here's before and after (left and right)


glebm:
:+1:

PR #166, opened 2016-03-27
Small clean-up of `_topic_nav.html.erb`

jayroh:
* Remove trailing returns from conditional blocks (beginning and end)
* Change regular anchor tags to `link_to`
* Remove unused variable assignment `(unread_count = unread_private_topics_count)`

### DIFF
--- a/app/views/thredded/shared/_topic_nav.html.erb
+++ b/app/views/thredded/shared/_topic_nav.html.erb
@@ -2,19 +2,21 @@
   <%= yield :thredded_breadcrumbs %>
 
   <ul class="thredded--topic-navigation">
-    <% if messageboard.try(:persisted?) %>
+    <% if messageboard.try(:persisted?) -%>
       <li class="thredded--topic-navigation--public">
-        <a href="<%= messageboard_topics_path(messageboard) %>">Topics</a>
+        <%= link_to 'Topics', messageboard_topics_path(messageboard) %>
       </li>
-    <% end %>
+    <% end -%>
+
     <li class="thredded--topic-navigation--private">
-      <a href="<%= private_topics_path %>">Private Topics
-        <% if (unread_count = unread_private_topics_count) > 0 %>
-            <span class="thredded--topic-navigation--unread">
-              <%= unread_private_topics_count %>
-            </span>
-        <% end %>
-      </a>
+      <%= link_to private_topics_path do %>
+        Private Topics
+        <% if unread_private_topics_count > 0 -%>
+          <span class="thredded--topic-navigation--unread">
+            <%= unread_private_topics_count %>
+          </span>
+        <% end -%>
+      <% end -%>
     </li>
   </ul>
 </div>
--- a/app/controllers/thredded/application_controller.rb
+++ b/app/controllers/thredded/application_controller.rb
@@ -35,7 +35,7 @@ module Thredded
     private
 
     def unread_private_topics_count
-      Thredded::PrivateTopic
+      @unread_private_topics_count ||= Thredded::PrivateTopic
         .joins(:private_users)
         .where(
           thredded_private_users: {


app/views/thredded/shared/_topic_nav.html.erb
```diff
@@ -2,19 +2,21 @@
   <%= yield :thredded_breadcrumbs %>
 
   <ul class="thredded--topic-navigation">
-    <% if messageboard.try(:persisted?) %>
+    <% if messageboard.try(:persisted?) -%>
```
TheDudeWithTheThing: What's the scenario where you don't have a persisted messageboard?

jayroh: It's a new, unsaved record - like on messageboards#new

app/views/thredded/shared/_topic_nav.html.erb
```diff
@@ -2,19 +2,21 @@
...
+
     <li class="thredded--topic-navigation--private">
-      <a href="<%= private_topics_path %>">Private Topics
-        <% if (unread_count = unread_private_topics_count) > 0 %>
-            <span class="thredded--topic-navigation--unread">
-              <%= unread_private_topics_count %>
-            </span>
-        <% end %>
-      </a>
+      <%= link_to private_topics_path do %>
```
glebm: Why is `link_to` better? I try to avoid `link_to` as it's an extra abstraction with no clear benefit.

app/views/thredded/shared/_topic_nav.html.erb
```diff
@@ -2,19 +2,21 @@
...
+        Private Topics
+        <% if unread_private_topics_count > 0 -%>
+          <span class="thredded--topic-navigation--unread">
+            <%= unread_private_topics_count %>
```
glebm: Would be better to use `unread_count` here to avoid an extra query here (it's cached in the query cache, but still).

jayroh: Ah I knew I forgot something. I meant to memoize the value in the
controller method to accomplish the reduction of extra method calls.

Also, using the `unread_count` in here is an extra level of redirection
that I feel is a bit unnecessary.

app/views/thredded/shared/_topic_nav.html.erb
TheDudeWithTheThing: I know what persisted is checking for, I mean how would a user get here?

Makes sense from a new perspective since this is a shared nav.

jayroh: No, I was certain you had no idea what persisted did. Are you sure?

:troll:

TheDudeWithTheThing: WHY I OUGHTA

app/views/thredded/shared/_topic_nav.html.erb
jayroh: If anything at this point, it's mostly about consistency and muscle memory. `grep -r "link_to" app/views` shows a whole lot of `link_to`'s, so ... I would err on the side of consistency. Not to be flippant, but - isn't half of rails an excessive abstraction? haha

Last but not least - vim/syntastic also yells at me when there are path helpers inside anchor tags :stuck_out_tongue:

app/views/thredded/shared/_topic_nav.html.erb
jayroh: Added commit c28aff6 w/some memoization

app/views/thredded/shared/_topic_nav.html.erb
glebm: Alright, not a big difference either way :+1:

app/views/thredded/shared/_topic_nav.html.erb
glebm: Great!

PR #169, opened 2016-03-29
[JRO] Explicitly require user permissions modules

jayroh:
When pulling this into thredded.com I ran into some issues with rails'
auto-loading of the user permissions modules. By explicitly requiring
the files we reduce the work rails (sometimes incorrectly) will do to
pull in the right files.

Also, aliasing `thredded_anonymous?` to `anonymous?` in the NullUser
class fixes an issue where the null user will be substituted for a real
user class and will be asked if it's anonymous.

### DIFF
--- a/app/models/thredded/null_user.rb
+++ b/app/models/thredded/null_user.rb
@@ -1,3 +1,9 @@
+require_relative './user_permissions/read/all'
+require_relative './user_permissions/write/none'
+require_relative './user_permissions/message/readers_of_writeable_boards'
+require_relative './user_permissions/moderate/none'
+require_relative './user_permissions/admin/none'
+
 module Thredded
   class NullUser
     include ::Thredded::UserPermissions::Read::All
--- a/app/models/thredded/user_extender.rb
+++ b/app/models/thredded/user_extender.rb
@@ -1,3 +1,9 @@
+require_relative './user_permissions/read/all'
+require_relative './user_permissions/write/all'
+require_relative './user_permissions/message/readers_of_writeable_boards'
+require_relative './user_permissions/moderate/if_moderator_column_true'
+require_relative './user_permissions/admin/if_admin_column_true'
+
 module Thredded
   module UserExtender
     extend ActiveSupport::Concern


app/models/thredded/null_user.rb
```diff
@@ -34,6 +40,8 @@ def thredded_anonymous?
       true
     end
 
+    alias_method :anonymous?, :thredded_anonymous?
```
glebm: In the real User class, this is also `thredded_anonymous?`: https://github.com/thredded/thredded/blob/8dedc824487063e2cc989c29af0b231c9c9485cd/app/models/thredded/user_extender.rb#L23

jayroh: Throughout the thredded engine we make use of the null object pattern and return either `super` (from the inherited ApplicationController), or OUR (thredded's) NullUser. See [here](https://github.com/thredded/thredded/blob/8dedc824487063e2cc989c29af0b231c9c9485cd/app/controllers/thredded/application_controller.rb#L91-L93). The issue with this is that the implementer *does not* own that NullUser class. Therefore the parent application controller might be sending methods to an object it does not control.

Here's an example ApplicationController snippet from my parent class.

```ruby
class ApplicationController < ActionController::Base
  def user_signed_in?
    current_user.present? && !current_user.anonymous?
    # I could really easily fix this if I change `anonymous` to `thredded_anonymous?`
  end
end
```

I suppose it's easy enough to implement `thredded_anonymous?` onto my parent User class but I think the less onus we put on the implementer, the better. Also - asking the parent class to lean heavily on a namedspaced method in a dependency feels a little weird, no?

What do you think? (By the way - I don't know if I was entirely clear--I may have rambled a bit there--so let me know if I'm missing any details.)

glebm: ll the methods thredded ever calls on a User object should be prefixed with `thredded_` so there is no chance of conflict, as `anonymous?` may mean something else for a particular application.

> The issue with this is that the implementer does not own that NullUser class. Therefore the parent application controller might be sending methods to an object it does not control.

Thredded `NullUser` should only ever be exposed to `Thredded` controllers, not the parent app controllers. I think there is an issue in the parent application (I can have a look once you push it to github).

> Also - asking the parent class to lean heavily on a namedspaced method in a dependency feels a little weird, no?

The parent class should never need to use this method.

jayroh: Ahhhh yup, I agree with everything you point out. I think this might be a case where we're defining and calling `current_user` from inside thredded - which is usually defined in the parent application. We should probably rename `current_user` to `thredded_current_user`

glebm: :+1: to renaming `current_user`.

jayroh:
Ok, I pulled out that extra method alias and will start an issue pointing towards renaming `current_user` to `thredded_current_user`

PR #170, opened 2016-03-29
[JRO] Pull notification preferences out to partial

jayroh:
At the (literal) top of the thredded views there's only one thing I
would initially want to pull out to the parent app's layout, or even
just hide - and that's the notification preferences link. Everything
else up there is just fine as it is.

By pulling this chunk of html out to a partial I can safely override it
in my parent app views and move it somewhere else (in my layout, for
instance) where it makes sense. For example:


glebm:
:+1:

Issue #167, opened 2016-03-27
Squash database migrations into a single migration


jayroh:
As mentioned in issue #141. Squashing migrations would certainly give us a smaller, cleaner start on projects. @glebm could you talk to me about some of the other benefits to doing this? I could go either way but would love to know your thoughts.

glebm:
Migrations are copied into the parent project. This is to avoid polluting the parent project with several dozen files we have currently.

When we do this, the squashed migration can have the same number as the last unsquashed migration, so there are no issues for the few existing installations.

Going forward, I like the ComfortableMexicanSofa approach:

1. One migration file for new installations:
https://github.com/comfy/comfortable-mexican-sofa/tree/master/db/upgrade_migrations

2. "Upgrade" migrations for tagged releases:
https://github.com/comfy/comfortable-mexican-sofa/tree/master/db/upgrade_migrations

3. Concise instructions:
https://github.com/comfy/comfortable-mexican-sofa/wiki/Upgrading-ComfortableMexicanSofa

jayroh:
This is EXACTLY what I was hoping you would respond with! Thank you @glebm!

glebm:
Depends on #182 for database-agnostic FTS/CI index / CI column creation

jayroh:
This has been merged in and pushed up with version 0.2.1

@glebm you're the man

Issue #142, opened 2015-09-01
Not installable when using SQLite3


Quintasan:
Hi,

When trying to install thredded in my Rails 4.2.3 application I get the following error:
`/home/vagrant/.rvm/gems/ruby-2.1.5/bundler/gems/thredded-7e4699fecddd/lib/thredded/case_insensitive_string_finder.rb:25:in 'use_adapter!': Please define CaseInsensitiveStringFinder for sqlite3 (RuntimeError)`

If this is intended then please mention it in the documentation.

glebm:
Hi @Quintasan, this gem currently only supports MySQL 5.6+ and Postgres, and this should definitely be mentioned in the docs.

The gem requires that the database provides full text search and case-insensitive exact match. It seems that the recent versions of SQLite3 also [provide](https://www.sqlite.org/fts3.html#section_3) Full Text Search. Would you like to improve thredded by implementing sqlite3 support?

glebm:
Resolved by #167

PR #173, opened 2016-03-29
Use db_text_search #172

### LINKED ISSUE #172: Split DB text stuff into its own gem
Already started on in https://github.com/thredded/db_text_search
The case insensitive string-in-set search bit is done.
Now need to do the FTS bit.

glebm:
It's happening! :watermelon:

Just released the first version of [db_text_search](https://github.com/thredded/db_text_search).

Resolves #172.

---

Next up:
1. Squash the migrations, removing any database-specific code from the resulting migration, and remove the schema files from version control, resolving #167.
2. Test on sqlite, as I think we fully support it now (db_text_search does), resolving #142.


app/models/thredded/post.rb
```diff
@@ -23,7 +23,7 @@ def private_topic_post?
 
     # @return [ActiveRecord::Relation<Thredded.user_class>] users from the list of user names that can read this post.
     def readers_from_user_names(user_names)
-      CaseInsensitiveStringFinder
+      DbTextSearch::CaseInsensitiveEq
         .new(Thredded.user_class.thredded_messageboards_readers([messageboard]), Thredded.user_name_column)
```
jayroh: Thoughts on pulling out this first param to a descriptive variable?

jayroh:
First and foremost - kudos (that word isn't strong enough) on the work pulling out the search stuff to db_text_search. Like, holy shit.

Second - question on how this effects thredded as to whether it maybe doesn't go far enough. As far as I can tell, we would still be supporting both mysql and postgresql for development purposes if this were to be merged right now. Would it be possible to take a shot even farther and pull all the search functionality into adapter gems that would act as a conduit to db_text_search? Example:  `thredded-mysql_search`, `thredded-pg_search`, `thredded-sqlite_search` ... etc?  My thought is that a default (or two?) would be good, in addition to a documented interface for anyone who would want to plug in another adapter (elasticsearch, sphinx, solr)

If you're familiar with how omniauth or griddler forked out their adapters to external gems, I was thinking something along those lines.

As I'm writing this I'm thinking - this work is too good to let go, so what do you think about pulling this in and continuing the conversation as to how this could evolve?

When all is said and done, this is awesome and this project is fortunate to have you kicking the crap out of it :).

app/models/thredded/post.rb
jayroh: I know that's not part of the diff but this *just* caught my eye due to proximity :)

db/migrate/20160329231848_text_search_indices.rb
```diff
@@ -0,0 +1,12 @@
+class TextSearchIndices < ActiveRecord::Migration
+  def change
+    DbTextSearch::CaseInsensitiveEq.add_index connection, :thredded_categories, :name, name: :thredded_categories_name_ci
+    # Recreate previously existing MySQL index in favour of consistent names
```
jayroh: Instead of the comment here what do you think about extracting the conditional and its contents into some descriptive methods like ...

```ruby
remove_old_indexes if mysql_adapter?
```

db/migrate/20160329231848_text_search_indices.rb
```diff
@@ -0,0 +1,12 @@
...
+    if connection.adapter_name.downcase =~ /mysql/
+      remove_index :thredded_topics, :title
+      remove_index :thredded_posts, :content
+    end
+    DbTextSearch::FullTextSearch.add_index connection, :thredded_topics, :title, name: :thredded_topics_title_fts
+    DbTextSearch::FullTextSearch.add_index connection, :thredded_posts, :content, name: :thredded_posts_content_fts
```
jayroh: So just above, if it's mysql it's going to remove the `:thredded_topics` and `:thredded_posts` indexes. If it's pq does it not need to remove those indexes too?  Or are those JUST a mysql thing .. and don't exist with pg because pg is superior in every way?

Oops is my bias showing

glebm: pg is superior in every way, except not being the most popular.

1. The index previously didn't exist at all on pg.
2. The index did exist on MySQL, but under a different name.

So, this:

1. Removes the previous indexes from mysql.
2. Adds the indexes everywhere.

db/migrate/20160329231848_text_search_indices.rb
glebm: I could, but well this is all going away once the migrations are squashed.

app/models/thredded/post.rb
glebm: Can do but need a good name (can't think of one)

glebm:
> As far as I can tell, we would still be supporting both mysql and postgresql for development purposes if this were to be merged right now. 

Yes, and additionally we would be supporting SQLite as well. This also enables us to do the development in a database-agnostic way, including the migrations. 

> Would it be possible to take a shot even farther and pull all the search functionality into adapter gems that would act as a conduit to db_text_search? Example: thredded-mysql_search, thredded-pg_search, thredded-sqlite_search ... etc? My thought is that a default (or two?) would be good, in addition to a documented interface for anyone who would want to plug in another adapter (elasticsearch, sphinx, solr)

For per-database gems, I think it's not necessary because db_text_search supports all of them. Adding a generic interface (for non-database search engines) is difficult because the unified interface needs to add indices, operate on ActiveRecord scopes, and manage data synchronization between the main database and the search database. And even if manage to do this, it would probably complicate the installation process of thredded quite a bit. I don't think it's a must-have feature for release to the wider open source community, but is something that can be added later on by other contributors (though it needs to be very well-designed to address the pain-points above).

app/models/thredded/post.rb
jayroh:
```ruby
def readers_from_user_names(user_names)
  messageboard_readers_scope = Thredded.user_class.thredded_messageboards_readers([messageboard])
  column_to_search = Thredded.user_name_column

  DbTextSearch::CaseInsensitiveEq
    .new(messageboard_readers_scope, column_to_search)
    .find(user_names)
end
```

??

db/migrate/20160329231848_text_search_indices.rb
jayroh: Duly noted!

Moving right along ...

jayroh:
> Yes, and additionally we would be supporting SQLite as well. This also enables us to do the development in a database-agnostic way, including the migrations.

Ahhh ... yes that's right! Now I got it.  DB agnostic migrations => use whatever db you want. No matter. Travis can be just the main state of authority for making sure the big ones are run through (the migrations and general integration) and sane.

> Adding a generic interface (for non-database search engines) is difficult because the unified interface needs to add indices, operate on ActiveRecord scopes, and manage data synchronization between the main database and the search database. 

Excellent point!

jayroh:
PUT 'ER THERE, GLEB

glebm:
:tada:

Issue #171, opened 2016-03-29
move `current_user` to `thredded_current_user`


jayroh:
There can be, and often will be, a clash between the parent application base controller class and the thredded application controller when it comes to the `current_user` method. Thredded should stay in its lane and not use what could potentially be a method owned and controlled by the parent application. Therefore we should move towards renaming `current_user` in thredded to `thredded_current_user`.

PR #176, opened 2016-04-01
[JRO] Rename current_user to thredded_current_user

jayroh:
There's no confusing what context we're in now.

This DOES reach into `current_user`, which is owned by the parent
application though. Possible TODO item(s) here is/are either allowing
configuration of the method called or expecting the parent application
to implement a `current_user` method.

This addresses issue #171

### DIFF
--- a/app/controllers/thredded/application_controller.rb
+++ b/app/controllers/thredded/application_controller.rb
@@ -29,7 +29,7 @@ module Thredded
       end
 
     def signed_in?
-      !current_user.thredded_anonymous?
+      !thredded_current_user.thredded_anonymous?
     end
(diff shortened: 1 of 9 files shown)


jayroh:
Thinking we may be able to sneak this into 0.2.0 before pushing that version up to Rubygems.

glebm:
:+1:

glebm:
A parent application *could* just implement a `thredded_current_user` method (similar to how Devise works), but because we require a NullUser object if a user is not signed in, I think it would be better to provide an option like `current_user_method` so that we can do `|| NullUser.new`.

Issue #160, opened 2016-03-22
Replace the Q gem with native Rails ActiveJob


glebm:
Q has not had any commits in two years, and Rails has supported [background jobs](http://edgeguides.rubyonrails.org/active_job_basics.html) since 4.2, and there is a [backport](https://github.com/ankane/activejob_backport) for Rails 4.0 and 4.1.

jayroh:
Closed with #177

PR #177, opened 2016-04-02
Replace Q gem with ActiveJob (and backport)

jayroh:
For several reasons it's probably a good idea to move from using the Q gem to
ActiveJob. This commit pulls out the dependencies for both Q and the threaded
in-memory queue that is used in conjunction with Q ... in favor of the built-in
ActiveJob *OR* the activejob backport.

Determining whether to add the backport is done with a little inspection of the
gemspec set. If the parent app is using a version of rails less than 4.2 then
it'll include the backport.

The backport only adds ActiveJob support, and does not monkeypatch
ActionMailer. ActionMailer got `deliver_now` and `deliver_later` in Rails 4.2
so this is somewhat of a notable omission. As a result we have to ask what
version of rails we're using in order to determine which version of the deliver
method we need to use -- `deliver` or `deliver_later`.

This PR addresses issue #160

### DIFF
--- a/thredded.gemspec
+++ b/thredded.gemspec
@@ -11,7 +11,7 @@ Gem::Specification.new do |s|
   s.homepage    = 'https://www.thredded.com'
   s.summary     = 'A messageboard engine'
   s.license     = 'MIT'
-  s.description = 'A messageboard engine for Rails 4.0 apps'
+  s.description = 'A messageboard engine for Rails 4.1+ apps'
 
   # backend
   s.add_dependency 'bbcoder', '~> 1.0'
@@ -22,7 +22,6 @@ Gem::Specification.new do |s|
   s.add_dependency 'htmlentities'
   s.add_dependency 'kaminari'
   s.add_dependency 'nokogiri'
-  s.add_dependency 'q'
   s.add_dependency 'rails', '>= 4.1.0'
   s.add_dependency 'rb-gravatar'
 
--- a/README.mkdn
+++ b/README.mkdn
@@ -60,19 +60,6 @@ Include thredded JavaScripts in your `application.js`:
 
 [initializer]: https://github.com/thredded/thredded/blob/master/lib/generators/thredded/install/templates/initializer.rb
 
-
-## Background Job Requirements
-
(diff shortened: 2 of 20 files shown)


thredded.gemspec
```diff
@@ -22,9 +23,9 @@ Gem::Specification.new do |s|
   s.add_dependency 'htmlentities'
   s.add_dependency 'kaminari'
   s.add_dependency 'nokogiri'
-  s.add_dependency 'q'
   s.add_dependency 'rails', '>= 4.0.0'
   s.add_dependency 'rb-gravatar'
+  s.add_dependency 'activejob_backport' if Thredded::RailsVersion.less_than_4_2?
```
jayroh: I have a weird feeling that either this could pose a problem or someone would consider this an issue. Just a hunch ... I'm not sure though.

jayroh:
Full disclosure - this was more satisfying than it probably should have been.

glebm:
Let's just drop support for 4.1. Rails 5 is coming out soon, meaning 4.1 will stop receiving bugfixes except security ones. The upgrade path from for 4.1 to 4.2 is very smooth.

thredded.gemspec
glebm: Yeah, it doesn't work like this, gem dependencies are determined at the time of building the gem

jayroh: Ha! Yeah I knew something didn't feel right. Should have listened to my gut
instead of my 2-beer-deep brain

glebm:
Rails gem downloads stats:
* Latest [4.2](https://rubygems.org/gems/rails/versions/4.2.6) release downloads: 235,338
* Latest [4.1](https://rubygems.org/gems/rails/versions/4.1.15) release downloads: 23,592

jayroh:
If anyone asks for it we can keep an unsupported branch for 4.0, and 4.1 in
git. No biggie

thredded.gemspec
```diff
@@ -11,7 +11,7 @@ Gem::Specification.new do |s|
   s.homepage    = 'https://www.thredded.com'
   s.summary     = 'A messageboard engine'
   s.license     = 'MIT'
-  s.description = 'A messageboard engine for Rails 4.0 apps'
+  s.description = 'A messageboard engine for Rails 4.0+ apps'
```
glebm: This should be 4.1+, and the Rails dependency below at 4.1+ as well, because db_text_search doesn't actually support 4.0. 4.0 also no longer receives bugfixes, so nobody should be using it.

jayroh: Ah-ha. Ok. Will make that change.

jayroh:
Alright @glebm. Barring anything else that's glaring this should be alright. I'm thinking it might not be a candidate for pulling into master yet ... certainly not until a `0.2.0` is tagged and released.

glebm:
@jayroh Looks really great, let's merge it before v0.2.0!

jayroh:
Alright!

Issue #159, opened 2016-03-22
Install generator should place all views and assets in `/app`, not `/vendor`


jayroh:
This relates to issue #157. Right now they end up in `/vendor` and should not.

jayroh:
@blimey85 if you would like to handle this issue I could point you to the file in question right here :) :point_right: https://github.com/jayroh/thredded/blob/master/lib/generators/thredded/install/install_generator.rb

:smile:

glebm:
@jayroh I don't think there should be a generator that copies everything! This makes updates very difficult, and is almost never ever necessary.

glebm:
To clarify, I don't think there should be a generator that copies anything at all. Just proper instructions on overriding thredded views and styles.

jayroh:
Before I say "I respectfully disagree, @glebm, my good sir :smile:" I want to make sure I'm understanding the approach you would suggest. Instead of a generator are you suggesting someone would have to manually find the files on their filesystem, in their gempath, and copy them over? Or, maybe download the raw file from github?

If that's the case then I can't say I agree. While I think that's doable, I think that's something we can make a heck of a lot easier with a generator (or part of a generator).

Is there a compromise, somewhere in between maybe? A flag for the respective parts of the UI, maybe? js, views, css, images, etc?

glebm:
I was thinking along the lines of pointing to [Rails Guides: Overriding Engine Views](http://edgeguides.rubyonrails.org/engines.html#overriding-views), warning against doing this (noting that updates will become very difficult), and documenting a command like this:
```sh
# Copy the post view into the application to customize it:
mkdir -p app/views/thredded/posts && cp "$(bundle show thredded)/$_/_post.html.erb" "$_"
```

blimey85:
What about how Devise does it, as an example?

Configuring views

We built Devise to help you quickly develop an application that uses authentication. However, we don't want to be in your way when you need to customize it.

Since Devise is an engine, all its views are packaged inside the gem. These views will help you get started, but after some time you may want to change them. If this is the case, you just need to invoke the following generator, and it will copy all views to your application:

rails generate devise:views
If you have more than one Devise model in your application (such as User and Admin), you will notice that Devise uses the same views for all models. Fortunately, Devise offers an easy way to customize views. All you need to do is setconfig.scoped_views = true inside the config/initializers/devise.rb file.

[...]

PR #178, opened 2016-04-02
Theming: Update readme, remove generators #159

### LINKED ISSUE #157: Thredded.theme setting in initializer.rb not working for me or I'm doing something wrong.
When trying to enable the theme initializer so I can customize the theme I get the following error:

`/Users/gary/Projects/thredded/config/initializers/thredded.rb:90:in `<top (required)>': undefined method `theme=' for Thredded:Module (NoMethodError)`

I might just be doing something wrong but just by uncommenting `Thredded.theme = :default` I'm getting this error. By default it's pulling theme files from within the gem, and obviously I don't want to make changes to those files, although I did change one template due to errors with devise paths. But I figured once I get it setup using a local copy of the template that wouldn't be a problem.

glebm:
As per the discussion in #159, removes the generators, documents styles customization, and individual view customization. Resolves #157.

### DIFF
--- a/README.mkdn
+++ b/README.mkdn
@@ -28,13 +28,10 @@ Add the gem to your Gemfile:
 gem 'thredded', github: 'thredded/thredded'
 ```
 
-Add a thredded initializer to your parent app by running the install generator.
-[The initializer] has detailed descriptions of all of the configuration options.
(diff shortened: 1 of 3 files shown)


jayroh:
Awesome. Absolutely needed. Thanks Gleb!

PR #184, opened 2016-04-02
Remove file_storage and asset_root configuration

glebm:
* `file_storage` is no longer used.
* Instead of a custom `asset_root`, just use `Rails.application.config.action_controller.asset_host`


jayroh:
Yes siiirrrrr

Issue #187, opened 2016-04-03
Images and videos should have size constraints to fit into the post DIVs


jayroh:
When testing out master against live data there are many images, and some videos that will blow out the post divs. The images can be fixed relatively easily - with some css for max-width - but videos will need something like [fitvids.js](http://fitvidsjs.com).

glebm:
How are there any videos at all? The sanitization filter doesn't allow them: https://github.com/thredded/thredded/blob/master/app/models/concerns/thredded/post_common.rb#L55

glebm:
If we only allow the `video` HTML5 tag, it can also be styled with pure CSS!

jayroh:
I guess that was something I set up on thredded.com!

Stuff like YouTube and Vimeo - what do you think about auto-embedding those?

This might then reach into the issue detailing the switch from bbcode OR
markdown to just markdown and *some* bbcode.

jayroh:
Got it - that's issue #68

jayroh:
re: vimeo and youtube - 

* https://github.com/dlackty/html-pipeline-vimeo Boom! :boom: 
* https://github.com/st0012/html-pipeline-youtube Boom! :boom:

jayroh:
Image issue fixed with #188

PR #188, opened 2016-04-03
[JRO] img's shouldn't be larger than container div

jayroh:
Fixes issue #187

### DIFF
--- a/app/assets/stylesheets/thredded/components/_post.scss
+++ b/app/assets/stylesheets/thredded/components/_post.scss
@@ -59,4 +59,8 @@
   hr {
     @extend %thredded--hr;
   }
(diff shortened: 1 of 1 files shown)


glebm:
As per how we do it in Bootstrap, it should also have `height: auto`: https://github.com/twbs/bootstrap/blob/v4-dev/scss/mixins/_image.scss#L10-L14

jayroh:
Good idea - will make that change and squash. Be right back

jayroh:
> how we do it in Bootstrap

Some day ... I'm buying you a beer and we're debating the merits of bootstrap.

glebm:
:+1: :smile:

Issue #189, opened 2016-04-03
Emojis need to be copied over to parent app


jayroh:
Otherwise the emoji part of html-pipeline references images that don't exist.

Example:
![](https://www.evernote.com/l/AUqGuCegHlhEAKY04uWCFEm_LYnfvOu1FPIB/image.png)

jayroh:
After running rake task in parent app

![](https://www.evernote.com/l/AUoIOYBvm5BCo60pfO7Z11kB8dV_sbwSL5AB/image.png)

glebm:
Is it possible to have the html-pipeline use regular Rails assets instead of copying the files to public? I prefer this approach since it doesn't bloat the app's git repository, and allows for updating the images (with fixed paths it's impossible to do consistenly with cache headers). The gemoji gem warns that this may slow down asset lookups considerably, but I haven't found it to be the case in practice on Ruby 2.x.

glebm:
Merged for now, hopefully later on there will be a solution that allows for both.

jayroh:
I hope so too. Looks like Gemoji suggests writing a helper? Might need to
look further into the adapter in html-pipeline

jayroh:
And thank you Gleb 😀

jayroh:
Fixed with pr #190

PR #190, opened 2016-04-03
[JRO] Rake task to install emoji images

jayroh:
`rake thredded:install:emoji` does the trick and copies images from
gemoji to the parent app's `./public/emoji` directory.

* Update README w/install instructions

Fixes issue #189

### DIFF
--- a/README.mkdn
+++ b/README.mkdn
@@ -25,13 +25,13 @@ Planned features:
 [Discourse]: http://www.discourse.org/
 [Forem]: https://www.github.com/radar/forem
 
-Thredded works with SQLite, MySQL (v5.6.4+), and PostgreSQL.
-Thredded has no infrastructure dependencies other than the database and, if configured in the parent application,
-the ActiveJob backend dependency such as Redis.
-Currently only MRI Ruby 2.2+ is supported. We would love to support JRuby and Rubinius as well.
+Thredded works with SQLite, MySQL (v5.6.4+), and PostgreSQL. Thredded has no infrastructure
+dependencies other than the database and, if configured in the parent application, the ActiveJob
+backend dependency such as Redis. Currently only MRI Ruby 2.2+ is supported. We would love to
+support JRuby and Rubinius as well.
 
-If you're looking for variations on a theme - see [Forem] and [Discourse].
-Forem is also an engine, while Discourse is a full app.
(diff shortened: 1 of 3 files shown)


README.mkdn
```diff
@@ -25,13 +25,13 @@ Planned features:
 [Discourse]: http://www.discourse.org/
 [Forem]: https://www.github.com/radar/forem
 
-Thredded works with SQLite, MySQL (v5.6.4+), and PostgreSQL.
-Thredded has no infrastructure dependencies other than the database and, if configured in the parent application,
-the ActiveJob backend dependency such as Redis.
-Currently only MRI Ruby 2.2+ is supported. We would love to support JRuby and Rubinius as well.
+Thredded works with SQLite, MySQL (v5.6.4+), and PostgreSQL. Thredded has no infrastructure
```
glebm: Markdown is usually formatted to 120 columns. Does Thredded use a different line length?

jayroh: You make a good point. If that's the usual norm I need to update my vimrc to adhere to this.

lib/tasks/thredded_tasks.rake
```diff
@@ -5,4 +5,14 @@ namespace :thredded do
   end
 
   task :nuke, [:slug] => :destroy
+
+  namespace :install do
+    desc 'Copy emoji to the Rails `public/emoji` directory'
+    task :emoji do
+      require 'emoji'
```
glebm: The gemoji gem [already provides a rake task](https://github.com/github/gemoji#gemoji) for this, so we can just do:

```ruby
load 'tasks/emoji.rake'
Rake::Task['emoji'].invoke
```

jayroh: It certainly does. Here's the crappy part - it places the images in the wrong place. I originally just wrapped the original task in the thredded namespace and let it rip but after trying that, the images ended up in `public/images/emoji` but then referenced them in `public/emoji`.

README.mkdn
glebm: :+1:

lib/tasks/thredded_tasks.rake
glebm: Oh, that's unfortunate :(

Issue #191, opened 2016-04-03
Categories on the main topic index need styling


jayroh:
Example:

![](https://www.evernote.com/l/AUpCj45icdtHUp3IvP5fLvPtyA27LgVRjsMB/image.png)

jayroh:
Fixed with PR #192

PR #192, opened 2016-04-03
[JRO] Style topics' categories

jayroh:
Previously the categories looked like an unstyled unordered list, so this
commit makes things render inline, just after the topic title link.

Also include a slight adjustment to the topics count style

This addresses issue #191


jayroh:
here's the result:

screenshot of categories

glebm:
Nice! :+1:

PR #193, opened 2016-04-03
[JRO] Closing div should be *inside* conditional

jayroh:
Otherwise it causes the pagination to be incorrectly scoped/styled.

Check it out:

Versus how it should look:


glebm:
:+1:

PR #194, opened 2016-04-04
Add gemspec files

glebm:
Looks like it was accidentally removed in:
3c257dabb71468d69799e9a47a4660eff72544bf

So the gems built since then have been empty.

### DIFF
--- a/thredded.gemspec
+++ b/thredded.gemspec
@@ -13,6 +13,9 @@ Gem::Specification.new do |s|
   s.license     = 'MIT'
   s.description = 'A messageboard engine for Rails 4.1+ apps'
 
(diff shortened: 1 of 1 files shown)


jayroh:
Oh ffs ... /facepalm

jayroh:
@glebm I presume the only recourse at this point is to push a couple of other versions with this fix -

1. With that's essentially 0.2.0 (but 0.3.0)
2. and 0.2.1 => 0.3.1

glebm:
We could just yank 0.2.0 and 0.2.1 from rubygems. As longs as there are tags in the repository, it's OK, since in the Readme we've been instructing people to install from the Github repository.

PR #196, opened 2016-04-05
Add Email Previews

glebm:
Adds email previews to the dummy app. Uses [a gem](https://github.com/glebm/rails_email_preview) I wrote, which is also a Rails engine.

It works by rendering emails with mock data provided by plain Ruby classes. If the users happens to use the same gem or a similar one, they can easily include Thredded emails to their list of previews.

Modifies the mailer code to optionally accept ActiveRecord instances as well, so that the preview classes can supply mock instances.

thredded-emails-list
thredded-email-html
thredded-email-raw


jayroh:
Yes! This is fantastic!

PR #197, opened 2016-04-05
Styles: more variables, and a dummy theme switcher

glebm:
My eyes are a bit tired of all the white, so I've added a dark theme to the dummy app. Having another theme helps us validate the variables of the default theme: that they are used everywhere they should be, and that we don't rely on functions like `lighten` too much. A dark theme is perfect for this as most issues are visible immediately.

The dark theme has a bit of the Matrix feel to it:
thredded-theme-dark

### DIFF
--- a/app/assets/stylesheets/thredded/_base.scss
+++ b/app/assets/stylesheets/thredded/_base.scss
@@ -3,9 +3,10 @@
 @import "base/variables";
 
 @import "base/alerts";
-@import "base/grid";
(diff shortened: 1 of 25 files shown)


jayroh:
@glebm merged, of course :). Could you update the CHANGELOG with your most recent changes when you have a minute? I keep forgetting to do it too so I'll probably be the voice reminding us all, including myself, to do so :smiley_cat:

PR #198, opened 2016-04-06
Change read state on topic link click

glebm:
A minor improvement, but this way the read state is reflected after
going to the topic and then back (via the back button).


jayroh:
@glebm go to gitter.im for a second

jayroh:
Looking forward to trying this one out tomorrow :)

Issue #183, opened 2016-04-02
Private topic form user names


glebm:
* [x] The form users field should fetch the user names from the server as the user types, because loading all the user names right away is slow for websites with tens of thousands of users.
* [ ] The form page should accept a parameter to pre-fill the user names, so that we can link to the existing conversation or the form from the currently online users list.

PR #199, opened 2016-04-06
Private topic form user names #183

glebm:
new-private-topic

* Autocomplete user names remotely. #183 
* Fix autocomplete styles to use thredded form styles.
* Add `$thredded-form-color/background` variables.
* Avoid creating notification jobs when seeding, bump the number of
  records, and log progress.
* Fix `includes` in topics and private topics controller. There is still
  an issue of loading read state for each topic, but this can't be fixed
  easily due to the decorator architecture.
* Update DbTextSearch to 0.2.0 for indexed prefix matching in user search.
* Fix Messageboard slugs to allow creating messageboards with reserved
  names.
* Update changelog

### DIFF
--- a/app/controllers/thredded/private_topics_controller.rb
+++ b/app/controllers/thredded/private_topics_controller.rb
@@ -3,15 +3,16 @@ module Thredded
     helper_method :private_topic
 
     def index
-      @new_private_topic = PrivateTopicForm.new(user: thredded_current_user)
-      @private_topics = PrivateTopic
-                          .uniq
-                          .for_user(thredded_current_user)
-                          .order('updated_at DESC')
-                          .on_page(params[:page])
-                          .load
+      @new_private_topic        = PrivateTopicForm.new(user: thredded_current_user)
+      @private_topics           = PrivateTopic
+                                    .uniq
+                                    .for_user(thredded_current_user)
+                                    .order('updated_at DESC')
+                                    .includes(:last_user, :user)
+                                    .on_page(params[:page])
+                                    .load
       @decorated_private_topics = Thredded::UserPrivateTopicDecorator
-        .decorate_all(thredded_current_user, @private_topics)
+                                    .decorate_all(thredded_current_user, @private_topics)
     end
 
     def show
@@ -19,9 +20,9 @@ module Thredded
       UserReadsPrivateTopic.new(private_topic, thredded_current_user).run
 
       @posts = private_topic
-        .posts
-        .includes(:user)
-        .order('id ASC')
+                 .posts
+                 .includes(:user)
+                 .order('id ASC')
 
       @post = private_topic.posts.build
     end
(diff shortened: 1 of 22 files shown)


app/models/thredded/messageboard.rb
```diff
@@ -3,11 +3,13 @@ class Messageboard < ActiveRecord::Base
     FILTERS = %w(markdown bbcode)
 
     extend FriendlyId
-    friendly_id :name, use: :slugged
+    friendly_id :slug_candidates,
+                use:            [:slugged, :reserved],
+                # Avoid route conflicts
+                reserved_words: %w(messageboards private-topics autocomplete-users theme-preview)
```
jayroh: I like how you deferred `slug_candidates` to the private method below, returning that array. What about doing the same with this array of possible route conflicts?

```ruby
# ...
reserved_words: possible_route_conflicts

private

def possible_route_conflicts
  %w(messageboards private-topics autocomplete-users theme-preview)
end
```

config/routes.rb
```diff
@@ -1,20 +1,24 @@
...
-  get '/:messageboard_id/:id/edit(.:format)' => 'topics#edit', as: :edit_messageboard_topic
-  get '/:messageboard_id/:id/page-:page(.:format)' => 'topics#show', as: :paged_messageboard_topic_posts, constraints: { page: /\d+/ }
-  get '/:messageboard_id/category/:category_id' => 'topics#category', as: :messageboard_topics_categories
+  scope path: '/:messageboard_id' do
+    get '/preferences/edit' => 'preferences#edit'
+    get '/new(.:format)' => 'topics#new', as: :new_messageboard_topic
+    get '/:id/edit(.:format)' => 'topics#edit', as: :edit_messageboard_topic
+    get '/:id/page-:page(.:format)' => 'topics#show', as: :paged_messageboard_topic_posts, constraints: { page: /\d+/ }
+    get '/category/:category_id' => 'topics#category', as: :messageboard_topics_categories
+  end
```
jayroh: :clap: :clap: :clap: :clap: :clap:

lib/thredded/seed_database.rb
```diff
@@ -21,6 +29,13 @@ def self.run(users: 5, topics: 26, posts: (0..20))
       s.create_topics(count: topics)
       s.create_posts(count: posts)
       s.create_private_posts(count: posts)
+    ensure
+      # Re-enable callbacks
+      SKIP_CALLBACKS.each { |(klass, *args)| klass.set_callback(*args) }
```
jayroh: "Hello. I am here to make it fancy af up in this bish"

app/controllers/thredded/private_topics_controller.rb
```diff
@@ -56,10 +57,13 @@ def private_topic
     def new_private_topic_params
       params
         .require(:private_topic)
-        .permit(:title, :locked, :sticky, :content, user_ids: [], category_ids: [])
+        .permit(:title, :locked, :sticky, :content, :user_ids, user_ids: [], category_ids: [])
         .merge(
           user: thredded_current_user,
-          ip: request.remote_ip)
+          ip: request.remote_ip).tap do |p|
```
jayroh: Could we move the `)` and everything after on that line down to the next line? There's a bit of visual/mental parsing needed here to catch that the merge ends and a tap begins

jayroh:
Just a few comments on this but I'll leave them up to you :)

Otherwise, let's get this merged in when you're ready!

app/controllers/thredded/private_topics_controller.rb
glebm: Done

app/models/thredded/messageboard.rb
glebm: Acknowledged, but can't pass a symbol here, and any method called here would need to be defined beforehand (above). We could add a constant but then it can't be private.

jayroh: Good point, man. Good point. Forgot that that is in the class scope.

If there's a way to accomplish that where it's *not* obnoxious, then I'd lean towards that but if not - don't worry about it.

PR #202, opened 2016-04-06
Fix pagination routes

glebm:
* All paginated routes now use the 'page-%d' format.
* 'page-%d' is now a reserved slug for topics and messageboards.
* Moves messageboard_topic_posts_path to messageboard_topic_path,
  same for private topics. This fixes the page numbers generated
  by kaminari (previously it generated ?page=n).

Also discovered bugs with last page link generation and left TODOs.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,8 +1,16 @@
 Thredded::Engine.routes.draw do
   resource :theme_preview, only: [:show], path: 'theme-preview' if %w(development test).include? Rails.env
 
-  resources :private_topics, path: 'private-topics' do
-    resources :private_posts, path: '', except: [:index], controller: 'posts'
+  page_constraint = { page: /[1-9]\d*/ }
+
+  scope path: 'private-topics' do
+    resource :private_topic, only: [:new], path: ''
+    resources :private_topics, except: [:new, :show], path: '' do
+      member do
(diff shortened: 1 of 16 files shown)


glebm:
I have a couple of improvements to URL handling ready to go in after this:

* Extract URL generation methods into a UrlsHelper, which we include into the controller, and make available in the helpers.
* Add a `post_path` method that generates the topic path with the right page and anchor, for both private and public posts.

config/routes.rb
```diff
@@ -11,20 +19,19 @@
...
+    resource :topic, path: 'topics', only: [:new]
+    resources :topics, path: '', except: [:index, :new, :show] do
+      collection do
+        get '(page-:page)', action: :index, as: '', constraints: page_constraint
+        get '/category/:category_id', action: :category, as: :categories
+      end
+      member do
+        get '(page-:page)', action: :show, as: '', constraints: page_constraint
+      end
+      resources :posts, except: [:index, :show], path: ''
```
jayroh: This is beautiful

jayroh:
>  the right page and the post anchor

Looking forward to this one. I never got a chance to look into why this broke in the first place :/

jayroh:
Thank you @glebm!

PR #206, opened 2016-04-08
Refactor search

glebm:
1. Use active_record_union to avoid manual SQL building.

   This avoids the `select id` subselect, making the query faster.

   This also allows us to return the scope, allowing to remove order and
   limit concerns from the search class. The controller now appends the
   order, so another Search implementation could still use its own
   order.

2. Removes usage of an exception for control flow.


TheDudeWithTheThing:
As the original author of the sql version I fully support this effort!  Would have done it myself eventually but time is not on my side.

app/controllers/thredded/topics_controller.rb
```diff
@@ -31,10 +31,17 @@ def show
     end
 
     def search
-      @topics = Topic.search(params[:q], messageboard)
+      query = params[:q].to_s
+      @topics = Topic.search(query, messageboard).order(updated_at: :desc).limit(50)
```
jayroh: This is excellent.

thredded.gemspec
```diff
@@ -19,6 +19,7 @@ Gem::Specification.new do |s|
   # backend
   s.add_dependency 'bbcoder', '~> 1.0'
   s.add_dependency 'cancancan'
+  s.add_dependency 'active_record_union', '>= 1.1.1'
```
jayroh: Nice gem. I like what I see so far

glebm:
This is good to review, I'll add pagination and move success/no results messages into the view in a subsequent PR.

lib/thredded/topics_search.rb
```diff
@@ -0,0 +1,58 @@
...
+        else
+          []
+        end
+    end
+
+    def text
+      @terms['text']
+    end
+  end
+end
```
jayroh: Love the functionality here, _however_ ... it could use some test coverage -- more-so as a resource for documentation and exhibiting behavior and less so for correctness or "cover your ass code coverage".  Search method has several branches (looks like 2^2?) so it would be helpful to see some scenarios outlined.

THAT BEING SAID ... I'm more than happy leaving it up to you to merge if you want to add a `# TODO` in here for further work. I obviously trust the code :).

glebm: Got you, will add the tests before merging.

jayroh: :two_hearts: :heart: :two_hearts:

jayroh: Feel free to merge when you're good and comfy, Gleb.

Thanks again :)

glebm:
This is good to review. Good call on tests, caught a couple of bugs.
Changes since the previous review are in the commit titles.

spec/models/thredded/topic_spec.rb
```diff
@@ -38,71 +38,62 @@ module Thredded
   end
 
   describe Topic, '.search' do
-    it 'with text' do
+    test_title = 'Xyzzy'
+
+    # On MySQL, a transaction has to complete before the full text search index is updated
```
jayroh: MYSQL!

:tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada:

glebm:
Looks like this fails with Rails 4.1 + PostgreSQL. I've [opened an issue](https://github.com/brianhempel/active_record_union/issues/8) with active_record_union. I'd rather we just drop support for 4.1, though. Is anyone still using it?

jayroh:
I know I'm not! With 5.0 around the corner we should probably move that way anyway.

(Which begs the question - we should probably add 5 to the test matrix and see how it looks but that's outside the scope of this PR.)

If you want to drop 4.1, I'm ok with it.

glebm:
4.1 support dropped! :+1: Good to merge!

jayroh:
@glebm if there's anything else worth bubbling up into the CHANGELOG, could you add it in a separate commit in addition to a note about this now supporting 4.2+?

jayroh:
This is A+ keeping it :100: fantastic by the way :smile:

glebm:
@jayroh Changelog updated.

jayroh:
You're the MAN!

glebm:
@jayroh [Almost green](https://travis-ci.org/thredded/thredded/jobs/121828234) on Rails 5, except active_record_union is [not yet compatible](https://github.com/brianhempel/active_record_union/issues/9) :sob:

PR #214, opened 2016-04-11
Adds delete action to post controller

TheDudeWithTheThing:
* Adds PostsController#destroy
* Adds link to delete to dummy app

Tests to follow after questions answered 

that_is_why_you_fail____dummy_app

### DIFF
--- a/app/controllers/thredded/posts_controller.rb
+++ b/app/controllers/thredded/posts_controller.rb
@@ -2,7 +2,7 @@ module Thredded
   class PostsController < Thredded::ApplicationController
     include ActionView::RecordIdentifier
 
-    load_and_authorize_resource only: [:index, :show]
+    load_and_authorize_resource only: [:index, :show, :destroy]
     helper_method :messageboard, :topic
     before_action :update_user_activity
 
@@ -25,6 +25,12 @@ module Thredded
 
     def topic
       post.postable
+    end
+
+    def destroy
+      post.destroy!
+
+      redirect_to request.referrer
     end
 
     private
--- a/app/controllers/thredded/posts_controller.rb
+++ b/app/controllers/thredded/posts_controller.rb
@@ -30,7 +30,7 @@ module Thredded
     def destroy
       post.destroy!
 
-      redirect_to request.referrer
+      redirect_to :back
     end
 
     private
--- a/app/helpers/thredded/urls_helper.rb
+++ b/app/helpers/thredded/urls_helper.rb
@@ -59,6 +59,16 @@ module Thredded
       end
     end
 
+    # @param post [Post, PrivatePost]
+    # @return [String] path to the DELETE PATCH PUT POST endpoint.
+    def delete_post_path(post)
+      if post.private_topic_post?
+        private_topic_private_post_path(post.postable, post)
+      else
+        messageboard_topic_post_path(messageboard, post.postable, post)
+      end
+    end
+
     # @param messageboard [Thredded::Messageboard, nil]
     # @param params [Hash] additional params
     # @return [String] the URL to the global or messageboard edit preferences page.
--- a/app/views/thredded/posts_common/_post.html.erb
+++ b/app/views/thredded/posts_common/_post.html.erb
@@ -19,6 +19,6 @@
   <% end %>
 
   <% if current_ability.can? :destroy, post.original %>
-    <%= link_to 'Delete Post', messageboard_topic_post_path(post.messageboard.slug, post.postable.slug, post.original), method: :delete, class: 'thredded--post--delete' %>
+    <%= link_to 'Delete Post', delete_post_path(post.original), method: :delete, class: 'thredded--post--delete' %>
   <% end %>
 <% end %>
--- a/app/views/thredded/posts_common/_post.html.erb
+++ b/app/views/thredded/posts_common/_post.html.erb
@@ -17,4 +17,8 @@
   <% if current_ability.can? :edit, post.original %>
     <%= link_to 'Edit Post', edit_post_path(post.original), class: 'thredded--post--edit' %>
   <% end %>
+
+  <% if current_ability.can? :destroy, post.original %>
+    <%= link_to 'Delete Post', messageboard_topic_post_path(post.messageboard.slug, post.postable.slug, post.original), method: :delete, class: 'thredded--post--delete' %>
+  <% end %>
 <% end %>
--- a/app/views/thredded/posts_common/_post.html.erb
+++ b/app/views/thredded/posts_common/_post.html.erb
@@ -19,6 +19,10 @@
   <% end %>
 
   <% if current_ability.can? :destroy, post.original %>
-    <%= link_to 'Delete Post', delete_post_path(post.original), method: :delete, class: 'thredded--post--delete' %>
+    <%= link_to 'Delete Post', delete_post_path(post.original),
+      method: :delete,
+      class: 'thredded--post--delete',
+      data: { confirm: I18n.t('thredded.posts.delete.confirm') }
+    %>
   <% end %>
 <% end %>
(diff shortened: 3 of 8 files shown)


app/views/thredded/posts_common/_post.html.erb
```diff
@@ -17,4 +17,8 @@
   <% if current_ability.can? :edit, post.original %>
     <%= link_to 'Edit Post', edit_post_path(post.original), class: 'thredded--post--edit' %>
   <% end %>
+
+  <% if current_ability.can? :destroy, post.original %>
+    <%= link_to 'Delete Post', messageboard_topic_post_path(post.messageboard.slug, post.postable.slug, post.original), method: :delete, class: 'thredded--post--delete' %>
```
TheDudeWithTheThing: @glebm this helper `messageboard_topic_post_path` seems weird when just above there is `edit_post_path`. When using the regular `post_path` method the path that came out of it was the `topic-slug/page-1#post-1`. Is that intended behavior?

app/controllers/thredded/posts_controller.rb
```diff
@@ -27,6 +27,12 @@ def topic
       post.postable
     end
 
+    def destroy
+      post.destroy!
+
+      redirect_to request.referrer
```
TheDudeWithTheThing: I hate this but can change once I figure out what path helper to call

app/views/thredded/posts_common/_post.html.erb
jayroh: Haven't looked into it but I will guess that `edit_post_path` is defined in the url/path/route helper module @glebm created. I'd have to dig in to verify but if that's the case then another method may need to be defined that looks like `post_path(post.original)` and expands it out to how the path helper looks like now.

app/controllers/thredded/posts_controller.rb
jayroh: `redirect_to :back`

glebm: Rails 5 defines a new redirect_back method that requires a mandatory `fallback_location` argument. We can use it even on Rails 4, because we polyfill it in the ApplicationController.

app/helpers/thredded/urls_helper.rb
```diff
@@ -58,5 +58,15 @@ def edit_post_path(post)
         edit_messageboard_topic_post_path(messageboard, post.postable, post)
       end
     end
+
+    # @param post [Post, PrivatePost]
+    # @return [String] path to the DELETE PATCH PUT POST endpoint.
+    def delete_post_path(post)
```
glebm: Rails convention is to just use `post_path` with the `DELETE` HTTP method

app/views/thredded/posts_common/_post.html.erb
```diff
@@ -17,4 +17,8 @@
   <% if current_ability.can? :edit, post.original %>
     <%= link_to 'Edit Post', edit_post_path(post.original), class: 'thredded--post--edit' %>
   <% end %>
+
+  <% if current_ability.can? :destroy, post.original %>
+    <%= link_to 'Delete Post', delete_post_path(post.original), method: :delete, class: 'thredded--post--delete' %>
```
glebm: Might consider using `data-confirm` because it's to click by accident

glebm:
Reviewed, looks good! :+1:  But we probably shouldn't allow deleting the very first post of a topic.

app/helpers/thredded/urls_helper.rb
jayroh: I think he tried to do that but ran into some issues with the helper we have as `post_path` now. @TheDudeWithTheThing was that the case?

app/views/thredded/posts_common/_post.html.erb
jayroh: Great call.

jayroh:
> we probably shouldn't allow deleting the very first post of a topic.

Also a good point! This could be (sort-of) bandaid'd with some css to hide the delete link for the first post.

glebm:
>  This could be (sort-of) bandaid'd with some css to hide the delete link for the first post.
I think it'd be easier to define a cancan rule for this. Solves both client/server side permissions and link-hiding.

jayroh:
Wondering how you would ask an arbitrary post whether it was post #1 in a topic or not, though. Would that be another round trip to the db? Or a flag/column on the post record?

There are certainly ways to do it, of course - I'm trying to think of the 80/20 solution where we wouldn't have to introduce more overhead.

glebm:
It would be a round trip to the DB, but it's really fast, and if a user runs a gigantic forum they can optimize this to avoid the round trip (since we actually do have all the necessary data)

glebm:
I mean a first_post association on the topic, so that there is only one extra query in total

glebm:
I can implement the first post restrictions after this PR is merged, so we can keep it simple here

app/helpers/thredded/urls_helper.rb
TheDudeWithTheThing: @glebm you clobbered the `post_path` so I think it's behaving not as expected

```ruby
    # @param post [Post, PrivatePost]
    # @return [String] path to the topic page with the post anchor.
    def post_path(post, params = {})
      post_url(post, params.merge(only_path: true))
    end
```

resulting URL is (at least in the dummy app)

> http://localhost:9292/thredded/main-board/that-is-why-you-fail-eb7a11aa-27ca-4ea4-89f9-1c7331f05995/page-1#post_2101

TheDudeWithTheThing:
all comments addressed, ready for re-review

jayroh:
@TheDudeWithTheThing - could you add some test coverage for this? On the feature spec level?

TheDudeWithTheThing:
A Butler always tests his code.

spec/features/thredded/user_deletes_post_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'spec_helper'
+
+feature 'User deleting posts' do
+  scenario 'can delete their own post' do
+    user.log_in
+    topic = users_topic
+    topic.visit_topic
+    last_post = last_post(topic)
```
jayroh: I feel like `last_post` would be handy as part of that topic object ... eg: `last_post = topic.last_post`. If you move that method into the topic object and memoize it I could see it being used in the future.

spec/features/thredded/user_deletes_post_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  scenario "cannot delete someone else's post" do
+    user.log_in
+    topic = someone_elses_topic
+    topic.visit_topic
+    last_post = last_post(topic)
+
+    expect(last_post).to be_listed
+    expect(last_post).to_not be_deletable
```
jayroh: ... which would then remove the need here to assign it to a variable. `expect(topic.last_post)` could be used for both of these here.

spec/features/thredded/user_deletes_post_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    topic = users_topic
+    topic.visit_topic
+    last_post = last_post(topic)
+
+    expect(last_post).to be_listed
+    expect(last_post).to be_deletable
+
+    last_post.delete
+
+    expect(last_post).to_not be_listed
```
jayroh: Total nitpick -- you can owe me a kick in the nuts the next time I see you for this, but -- can you change the `to_not`s to `not_to`s?

Haha I'm sorry! Almost the entire suite is using `not_to` so ... please please allow me that one little anal stupidity.

TheDudeWithTheThing: hey I'm just following your lead here

https://github.com/thredded/thredded/blob/master/spec/features/thredded/user_deletes_topic_spec.rb#L23

jayroh:
Thank you Shaun! ❤️ 

@glebm, take it away!

app/helpers/thredded/urls_helper.rb
glebm: @TheDudeWithTheThing Ugh, I see :+1:

glebm:
:+1:

PR #215, opened 2016-04-11
Global Search

### LINKED ISSUE #211: Global search

glebm:
Search across all messageboards.

For now, uses the same controller#action, but the routes are separate
so the global search can use a different action in the future, e.g. one
that groups the results by Messageboard names.

Resolves #211

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -32,6 +32,11 @@ en:
     private_topics:
       errors:
         user_ids_length: Please specify at least one other user.
+    search:
+      form:
+        btn_submit: :thredded.search.form.label
+        label: Search
+        placeholder: Search Topics and Posts
     topics:
       search:
         no_results_message: There are no results for your search - %{query}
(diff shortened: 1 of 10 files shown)


config/locales/en.yml
```diff
@@ -32,6 +32,11 @@ en:
     private_topics:
       errors:
         user_ids_length: Please specify at least one other user.
+    search:
+      form:
+        btn_submit: :thredded.search.form.label
```
jayroh: Full disclosure - I'm an i18n dummy. I haven't fully i18n'd an app to this day (I regret that, by the way).

That being said - is it common practice to shorten "button" to "btn"? If that's not a convention, I'd lean towards expanding this out. If it's a convention, please ignore me :)

jayroh:
Just one comment here @glebm

config/locales/en.yml
glebm: btn is a common abbreviation of button, popularized by Bootstrap. Example: [Discourse](https://github.com/discourse/discourse/blob/master/config/locales/client.en.yml#L479).

PR #216, opened 2016-04-12
Add youtube and vimeo support by default

jayroh:
It's a messageboard, people will post videos - I'm pretty sure of it. This had
allowed iframes only from youtube but by pulling in the youtube pipeline that
should negate the need to define all this ourselves.

Get rid of the whitelist customization - it's is no longer needed.


glebm:
:+1:

PR #219, opened 2016-04-12
[SMB] Seeds users with preferences for dummy app

### LINKED ISSUE #218: Error when creating new private message
Not sure if it's just how the dummy app is initialized but when creating a new private topic I get the following

undefined method `notify_on_message?' for #<Thredded::UserPreference:0x007fb1a0726740>

TheDudeWithTheThing:
Fixes #218

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -48,7 +48,7 @@ module Thredded
 
     def create_users(count: 5)
       log "Creating #{count} users..."
-      @users = [user] + FactoryGirl.create_list(:user, count)
+      @users = [user] + FactoryGirl.create_list(:user, count, :with_user_preferences)
     end
 
     def create_messageboard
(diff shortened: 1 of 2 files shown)


jayroh:
Beautiful! Thank you Shaun! 🎉

glebm:
This shouldn't be necessary. Thredded only creates a Preference record once the user changes the preferences. By default, there is no preference record, so this doesn't represent the data a real app would have. IMO this should be reverted or only applied to a subset of the users.

TheDudeWithTheThing:
REVERT THIS @jayroh 

I'll confirm this is fixed / broken with the latest version post revert.

jayroh:
Whoops. Ok. We'll get this situated shortly

jayroh:
Just pushed a commit up that effectively reverts this. No big deal

TheDudeWithTheThing:
sorry to all parties involved!

jayroh:
I wonder what Vince has to say about all this ... paging Vince ...

PR #221, opened 2016-04-13
WIP - Allow both bbcode and markdown

jayroh:
Instead of asking messageboard administrators to decide which "filter" to use for posts just let them use a subset of bbcode and markdown simultaneously.

TODO: move migration from default generated location into the rolled up migration, and the 0.2.0 -> 0.3.0 migration.


app/models/concerns/thredded/post_common.rb
```diff
@@ -59,37 +58,22 @@ def context_options
...
       )
-    end
-
-    def html_filter_for_pipeline
-      if filter == 'bbcode'
-        HTML::Pipeline::BbcodeFilter
-      else
-        HTML::Pipeline::MarkdownFilter
-      end
+      HTML::Pipeline::SanitizationFilter::WHITELIST
```
jayroh: Thinking that this:

```ruby
def sanitize_whitelist
  HTML::Pipeline::SanitizationFilter::WHITELIST[:elements] += %w(
    fieldset
    legend
    blockquote
  )
  HTML::Pipeline::SanitizationFilter::WHITELIST
end
```

... could definitely be better. I'm just tired and my brain is fried.

lib/html/pipeline/bbcode_filter.rb
```diff
@@ -5,14 +5,22 @@ class Pipeline
...
-        "<p>#{html}</p>"
+        html
+      end
+
+      def remove_url_link_contents(html)
+        doc = Nokogiri::HTML::DocumentFragment.parse(html)
+        doc.css('a').each do |link|
+          link.content = link.content.gsub(%r{https?://}, '')
+        end
+        doc.to_html
```
jayroh: The reason this is here is because the combination of bbcode and markdown butchers the generated markup for something like:

```
[url]http://yahoo.com[/url]  =>  <a href="http://yahoo.com"><a href="http://yahoo.com">http://yahoo.com</a></a>
```

With this method it'll find the contents of `<a>` tags and remove the `https?://` before it hits the markdown filter.

spec/models/thredded/post_spec.rb
```diff
@@ -69,7 +65,7 @@ module Thredded
       topic = create(:topic)
       future_time = 3.hours.from_now
       create(:post, postable: topic, user: joel, content: 'posting here')
-      Timecop.freeze(future_time) do
+      travel_to future_time do
```
jayroh: Will need to remove Timecop entirely and use the built-in rails helper. This is just a start.

spec/models/thredded/post_spec.rb
```diff
@@ -85,105 +81,142 @@ module Thredded
...
+      expect(post.filtered_content(view_context)).to eq expected_html
     end
 
     def parsed_html(html)
-      Nokogiri::HTML::DocumentFragment.parse(html).to_hash
+      Nokogiri::HTML::DocumentFragment.parse(html) { |config| config.noblanks }
+        .to_html
+        .gsub(/^\s*/, '')
+        .gsub(/\s*$/, '')
+        .gsub(/^$\n/, '')
```
jayroh: I know for sure this can be optimized but ... 💤

app/models/concerns/thredded/post_common.rb
glebm: `blockquote` is already in the whitelist, and it doesn't seem that fieldset/legend should be whitelisted.

spec/models/thredded/post_spec.rb
glebm: :+1: didn't know about `travel_to`

spec/models/thredded/post_spec.rb
TheDudeWithTheThing: `.to_html.delete(' ').delete("\n")`

weird there's no native ruby method to do such a thing

```
[5] pry(main)> to_html = " a string with spaces\n"
=> " a string with spaces\n"
[6] pry(main)> to_html.delete(' ').delete("\n")
=> "astringwithspaces"
```

app/models/concerns/thredded/post_common.rb
jayroh: > blockquote is already in the whitelist, and it doesn't seem that fieldset/legend should be whitelisted.

I think you're right. Markdown will just wrap a blockquote, right? If that's the case then, yeah - we should retain some parity there.  Will inspect this html once I submit :)

jayroh: Yup. `blockquote`!

spec/models/thredded/post_spec.rb
jayroh: Agreed. I wish there was. I think the use case is narrow and approachable enough, though, where someone patient enough could write a regex that does the job. Unfortunately I was/am tired and didn't want to jump through those hoops more than just chaining these gsubs, haha.

jayroh:
I think this is ready for another review.

If/when you guys are ok with it I'll squash and rebase before merging.

spec/models/thredded/post_spec.rb
TheDudeWithTheThing:
```
[8] pry(main)> to_html.gsub /[\s|\n]/, ''
=> "astringwithspaces"
```

jayroh: I want to keep spaces. I DO want to get rid of

1. Empty lines.
2. left padding.
3. right padding.

glebm: `.gsub(/^\s*/, '')` breaks code snippets.

jayroh: This method is internal to only this spec so it wouldn't break anything. This is just to get rid of the noise to make comparing the result with the expected output easier.

glebm:
I've just clicked the merge button without realizing this hasn't been rebased yet :sob: 
No big deal though!

jayroh:
¯\\\_(ツ)\_/¯  Enh, no big deal.

PR #228, opened 2016-04-14
Replace CanCan with Pundit

glebm:
Pundit is the modern alternative to CanCan.

* Plain old ruby objects that actually correspond to our old Permissions objects, so one less level of indirection and half the code.
* Easier to test, because plain ruby objects plus neat rspec helpers.
* Easier to reason about, because all of the permissions are defined in the policy method (no CanCan magic disjunctive `can`, overriding `cannot`, and other horrors).

Also, shoulda-matchers conflict with Pundit matchers, so I just removed them. They are not really testing anything (that Rails doesn't already test internally).


db/migrate/20160329231848_create_thredded.rb
```diff
@@ -50,7 +49,7 @@ def change
 
     create_table :thredded_posts do |t|
       t.integer :user_id, limit: 4
-      t.text :content, limit: 65535
+      t.text :content, limit: 65_535
```
jayroh: nice - I like this :)

glebm: Added db/ to Rubocop, this was one of the warnings :)

jayroh:
Never considered making this move too too much but all of your points are more than salient. First reaction was to wonder if it was worth it but after looking at the diffs, and your reasoning, yup - I'm sold.

Thank you @glebm - gonna merge this right on in :)

Issue #213, opened 2016-04-11
Change copy from "Private Topics" to "Messages"?


glebm:
Private topics are basically messages, and everyone is familiar with how messages work, while I've never seen "Private Topics" before. Should we change the copy?
@jayroh

jayroh:
Naming is hard.

I think this is a perfect opportunity to throw this label title in an i18n translation and let anyone change it to what they want.

* Messages
* Private Topics
* Private Messages
* Direct Messages
* Group Messages

It could be so many things!

glebm:
The nav/label is already in i18n (but not the mailers, yet). I'm talking about the default here. "Private topics" does not seem like a good default, it's not well-recognized.

jayroh:
I can see the argument against "private topics", but I would say that just "messages" is a bit on the vague side. I think "Private Messages" would be a safe and happy compromise.  Looks like that's what discourse uses:

![](https://www.evernote.com/l/AUqFdRkOIORNJYdvEkLX6EzrHZmJ1W-9u_MB/image.png)

I don't *want* to mimic discourse but if they've rolled that functionality out with that then it must be intuitive enough.

glebm:
:+1: Private messages are intuitive enough

PR #229, opened 2016-04-14
Change copy: Private Topics to Private Messages

glebm:
"Private Topics" are now "Private Messages".
A thread of "Private Messages" is referred to as just "Conversation", because the "Private" bit is always prominent in the navigation / title.

A take on #213.


jayroh:
:ship: it, @glebm! :tada: !

jayroh:
@gleb-zilla!

Issue #230, opened 2016-04-14
Markdown images no longer work


glebm:
```
![img](https://cloud.githubusercontent.com/assets/216339/14379806/da716a1e-fd73-11e5-90f1-6dbdb708d3d5.png)
```

Expected:

> an image here

Actual:

> !(https://cloud.githubusercontent.com/assets/216339/14379806/da716a1e-fd73-11e5-90f1-6dbdb708d3d5.png)

jayroh:
Mf'er

PR #232, opened 2016-04-15
[JRO] Allow img tags in posts

jayroh:
`img` tags were busted according to issue #230.

This commit adds regression test & fix bbcode/mkdn interop. Change the config for bbcode img tags so that `img` won't be interpreted as a bbcode image tag ... make it require a closing `[/img]` tag.

### DIFF
--- a/lib/thredded/engine.rb
+++ b/lib/thredded/engine.rb
@@ -31,5 +31,13 @@ module Thredded
         thredded/breadcrumb-chevron.svg
       )
     end
+
+    initializer 'thredded.setup_bbcoder' do
+      BBCoder.configure do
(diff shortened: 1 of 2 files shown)


jayroh:
If/when this branch is OK'd I'll want to squash and merge manually so just give me the heads up when I get the ":+1:"

lib/thredded/engine.rb
```diff
@@ -31,5 +31,13 @@ class Engine < ::Rails::Engine
         thredded/breadcrumb-chevron.svg
       )
     end
+
+    initializer 'thredded.setup_bbcoder' do
+      BBCoder.configure do
```
glebm: It's a bit of a shame that this library requires global configuration, oh well.

glebm:
:+1:

PR #233, opened 2016-04-15
Add rel="nofollow noopener" to external links

jayroh:
If someone links, in a post, to an external link, we should add `rel="nofollow noopener"`. `nofollow` to avoid google-juice'ing the link, and noopener to open a new tab/window on click.

### DIFF
--- a/app/models/concerns/thredded/post_common.rb
+++ b/app/models/concerns/thredded/post_common.rb
@@ -54,7 +54,16 @@ module Thredded
       {
         asset_root: Rails.application.config.action_controller.asset_host || '',
         post:       self,
+        whitelist:  sanitize_whitelist,
       }
+    end
+
+    def sanitize_whitelist
+      HTML::Pipeline::SanitizationFilter::WHITELIST.deep_merge(
+        attributes: {
+          'a' => %w(href rel),
+        }
+      )
     end
 
     def update_parent_last_user_and_timestamp
--- a/app/models/concerns/thredded/post_common.rb
+++ b/app/models/concerns/thredded/post_common.rb
@@ -65,7 +65,10 @@ module Thredded
 
           a_tags = node.css('a')
           a_tags.each do |a_tag|
-            a_tag['rel'] = 'nofollow noopener' if a_tag['href'].starts_with? 'http'
+            if a_tag['href'].starts_with? 'http'
+              a_tag['target'] = '_blank'
+              a_tag['rel'] = 'nofollow noopener'
+            end
           end
         end,
       ]
--- a/app/models/concerns/thredded/post_common.rb
+++ b/app/models/concerns/thredded/post_common.rb
@@ -59,10 +59,20 @@ module Thredded
     end
 
     def sanitize_whitelist
+      HTML::Pipeline::SanitizationFilter::WHITELIST[:transformers] += [
+        lambda do |env|
+          node = env[:node]
+
+          a_tags = node.css('a')
+          a_tags.each do |a_tag|
+            a_tag['rel'] = 'nofollow noopener' if a_tag['href'].starts_with? 'http'
+          end
+        end,
+      ]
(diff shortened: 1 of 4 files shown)


jayroh:
*NOTE:* This is a PR against another branch that will need to be merged BEFORE this is merged. 

## Do not merge this branch (yet)

Once both of these branches are accepted I'll merge them in manually.

lib/thredded.rb
```diff
@@ -24,30 +24,31 @@
...
-  self.avatar_url = ->(user) { Gravatar.src(user.email, 128, 'mm') }
   self.active_user_threshold = 5.minutes
-  self.layout = 'thredded/application'
+  self.admin_column = :admin
+  self.avatar_url = ->(user) { Gravatar.src(user.email, 128, 'mm') }
   self.email_reply_to = -> postable { "#{postable.hash_id}@#{Thredded.email_incoming_host}" }
+  self.layout = 'thredded/application'
   self.moderator_column = :admin
-  self.admin_column = :admin
+  self.user_name_column = :name
```
jayroh: These changes are merely for formatting and alphabetical ordering.

... Because OCD.

app/models/concerns/thredded/post_common.rb
```diff
@@ -54,9 +54,31 @@ def context_options
       {
         asset_root: Rails.application.config.action_controller.asset_host || '',
         post:       self,
+        whitelist:  sanitize_whitelist,
       }
     end
 
+    def sanitize_whitelist
+      HTML::Pipeline::SanitizationFilter::WHITELIST[:transformers] += [
```
glebm: `s/+=/+` because `+=` modifies the constant in another module.

app/models/concerns/thredded/post_common.rb
```diff
@@ -54,9 +54,31 @@ def context_options
       {
         asset_root: Rails.application.config.action_controller.asset_host || '',
         post:       self,
+        whitelist:  sanitize_whitelist,
       }
     end
 
+    def sanitize_whitelist
```
glebm: We probably should define our own constant, because currently:
a) It modifies a constant in another module, dangerous (the user may be using the `SanitizationFilter::WHITELIST` for something else).
b) It evaluates this for every post.

app/models/concerns/thredded/post_common.rb
```diff
@@ -54,9 +54,31 @@ def context_options
...
+      HTML::Pipeline::SanitizationFilter::WHITELIST[:transformers] += [
+        lambda do |env|
+          node = env[:node]
+
+          a_tags = node.css('a')
```
glebm: Instead of adding a transforming and allowing rel, let's use the `add_attributes` features:

```ruby
add_attributes: {
  'a' => {'target' => 'blank', 'rel' => 'nofollow noopener'}
}
```

glebm:
Reviewed :+1:

app/models/concerns/thredded/post_common.rb
jayroh: Great points. I've addressed these here - dd0c02e580fdf763628541a0a48c53ce8c7f6ba5

What do you think?

app/models/concerns/thredded/post_common.rb
jayroh: This would definitely be easier! However, these attributes would be applied to internal links too. 

External links? Ok, I'm totally on board :). Relative path links, though? Eeeennnhhh notsomuch.

That's why I went the route of the transformer thing ... ask the link `href` where it's going, then add the attributes.

Whataya say?!

jayroh:
I can't reopen this because I killed the target branch but we'll keep this here for discussion purposes until things look OK :)

@glebm - back up for review with the changes in dd0c02e580fdf763628541a0a48c53ce8c7f6ba5

app/models/concerns/thredded/post_common.rb
glebm: SGTM

PR #236, opened 2016-04-15
[JRO] Add iframe to element whitelist

jayroh:
This fixes an issue where videos from vimeo and youtube are no longer
being allowed through the sanitizer.

While I'm in here I've also crunched all tests for bbcode and markdown
being processed down into one test looping through sample code and
expected output in a YAML file.

This allows us to reenforce the ModuleLength cop in rubocop back to 175
lines. The Post spec kept forcing us to bump that Max number up as we
added regression tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -53,7 +53,7 @@ Metrics/MethodLength:
   - 'db/**/*'
 
 Metrics/ModuleLength:
-  Max: 200
+  Max: 175
 
 Metrics/PerceivedComplexity:
   Enabled: false
--- a/app/models/concerns/thredded/post_common.rb
+++ b/app/models/concerns/thredded/post_common.rb
@@ -17,10 +17,16 @@ module Thredded
       end
     ].freeze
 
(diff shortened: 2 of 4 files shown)


glebm:
For security, the iframe should be sandboxed, something like:

```
sandbox="allow-same-origin allow-forms allow-scripts"
```

For compatibility with screen readers, the iframe should also have the `seamless` attribute.

jayroh:
Done!

Only interesting thing - I don't see if/where you can add an attribute without a value applied so I made seamless -> `seamless="true"`.

I looked into the source here - https://github.com/rgrove/sanitize/blob/master/lib/sanitize/transformers/clean_element.rb - and it looks like it's going to add a value no matter what, so ... true should be ok for now.

I pulled down the sanitize source, maybe I'll play with adding that as a feature.

jayroh:
> nit: the seamless attribute should either have no value, or the value seamless

Ha - you have impeccable timing.

glebm:
Awesome! :+1:

jayroh:
Was going to squash these but the history here is valuable :)

I'll merge and deploy.

PR #240, opened 2016-04-16
Run demo on heroku

jayroh:
Sort of.

* Assets aren't being served.
* Errors aren't showing up in the logs (bizarre as hell)

Otherwise things look like they're working.

@glebm @TheDudeWithTheThing - you should have emails with heroku credentials and repo information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,10 @@
 db/schema*.rb
 Gemfile.lock
 *.gemfile.lock
+!heroku.gemfile.lock
 .bundle/
-bin/
+bin/*
+!bin/rails
 log/*.log
 pkg/
 spec/dummy/db/*.sqlite3
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: bundle exec puma -C ./spec/dummy/config/puma.heroku.rb
--- a/Rakefile
+++ b/Rakefile
@@ -24,15 +24,6 @@ end
 APP_RAKEFILE = File.expand_path('../spec/dummy/Rakefile', __FILE__)
 load 'rails/tasks/engine.rake'
 
-require 'rubocop/rake_task'
-RuboCop::RakeTask.new
-
-require 'rspec/core/rake_task'
-RSpec::Core::RakeTask.new(:spec)
-
-task(:default).clear
-task default: [:spec, :rubocop]
-
 # Common methods for the test_all_dbs, test_all_gemfiles, and test_all Rake tasks.
 module TestTasks
   module_function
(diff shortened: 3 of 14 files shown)


jayroh:
I'm so annoyed that I can't get assets to serve .... ugh.

💤

glebm:
I'll have a look. We can probably just have Rails serve the assets and cache them with CloudFront.

glebm:
Got the assets and logging working.

1. Cleaned up the config, moved all heroku stuff into heroku-named files (heroku.gemfile, puma.heroku.rb, etc).
2. Switched to a custom buildpack that uses heroku.gemfile instead of Gemfile.
3. Squashed and rebased all the commits.

glebm:
We should put CloudFront in front of it, so that assets are automatically served from a CDN.

glebm:
Also runs Rails 5 now

jayroh:
@glebm! THANK YOU. Honestly this was driving me crazy and I couldn't sleep. Thank you for fixing this up.

For future reference I'm going to put the link to the buildpack here so anyone looking (including myself) can find it - https://github.com/glebm/heroku-buildpack-ruby  👈  that is very handy

jayroh:
```ruby
config.public_file_server.enabled = true
```

Was this the part needed to get the assets served? Because oh my god this was driving me batshit insane

jayroh:
Going to merge this in and spin up cloudflare on top of it.

Since I have a good number of dependencies for thredded.com at it's current nameserver I'll hose this under thredded.org - which makes some sense, as it could house further properties under that (blog, etc)

glebm:
> ```ruby
> config.public_file_server.enabled = true
> ```
>
> Was this the part needed to get the assets served? Because oh my god this was driving me batshit insane

Yeah, Heroku doesn't have a file server, so we need Rails to do that instead. This is a Rails 5 thing, it is the same as `config.serve_static_files` in Rails 4. While serving files from Rails is slow, Cloudflare will be in front of it and it will cache the assets, and serve them from the CloudFlare CDN.

jayroh:
After poking around a little bit, it looks like there are some AR / db pool
issues?

> FATAL -- : ActiveRecord::ConnectionTimeoutError (could not obtain a
connection from the pool within
5.000 seconds (waited 5.002 seconds); all pooled connections were in use):

Seeing this quite a bit after clicking around the app for a few minutes.

And then I provisioned New Relic and for some reason can't get that to
report stats ... even after 20 minutes or so. Hmm.

I was optimistic and thought it would "just work" and committed what new
relic needs to master, so they're in there even though things aren't
working. Vexing, I tell you :/

On Sat, Apr 16, 2016 at 10:28 AM, Gleb Mazovetskiy <notifications@github.com

glebm:
The pool size must be at least workers x threads.

glebm:
I'll fix it

glebm:
OK, this should fix the pool size error https://github.com/thredded/thredded/commit/247fe96309c815757414e862ba00192ebcc036e6

jayroh:
Can I steal/kidnap you and make you work with me everywhere.

ha ha just kidding. sort of. haha ... ok not really.

(thank you Gleb)

PR #246, opened 2016-04-17
[JRO] Rake task to re-seed database

jayroh:
This commit allows us to quickly and easily truncate database tables, then
re-seed with new sample data. This can, and will, be used in the demo
application to blow away data and re-create it on the daily.

Addresses Issue #245

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -133,3 +133,23 @@ else
   task(:default).clear
   task default: [:spec, :rubocop]
 end
+
+namespace :db do
+  desc "Wipe out all tables' data"
+  task truncate: :environment do
+    connection = ActiveRecord::Base.connection
+    tables = connection.tables - %w(schema_migrations)
+
+    tables.each do |table|
+      if connection.adapter_name =~ /sqlite/i
+        connection.execute("DELETE FROM #{table}")
+        connection.execute("DELETE FROM sqlite_sequence where name='#{table}'")
+      else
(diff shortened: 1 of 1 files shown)


glebm:
Why not just run this daily:

```
rake db:drop db:create db:migrate db:seed
```

jayroh:
This would skip drop, create, and migrate completely.

And - I'd probably use this for my dev environment pretty regularly too so the less I have to do, the better 😛

glebm:
It only shaves off 1-3 seconds but sure :+1:

Rakefile
```diff
@@ -133,3 +133,27 @@ else
...
   task default: [:spec, :rubocop]
 end
+
+namespace :db do
+  desc "Wipe out all tables' data"
+  task truncate: :environment do
+    adapter = ActiveRecord::Base.configurations[::Rails.env]['adapter']
+    connection = ActiveRecord::Base.connection
+
+    tables = connection.tables.select do |table|
```
glebm:
```ruby
connection.tables - ['schema_migrations']
```

Rakefile
```diff
@@ -133,3 +133,27 @@ else
...
+      table.exclude?('schema_migrations')
+    end
+
+    tables.each do |table|
+      if adapter.include? 'sqlite'
```
glebm:
```ruby
connection.adapter_name =~ /sqlite/i
```

PR #247, opened 2016-04-17
[JRO] Add google analytics script tag

jayroh:


glebm:
The Javascript has to be a bit different because Turbolinks: http://reed.github.io/turbolinks-compatibility/google_universal_analytics.html

jayroh:
Oh man I'm so glad you posted that. That's super super handy.

glebm:
:+1:

Issue #243, opened 2016-04-16
Fix read state performance


glebm:
1. It's loading the read state for *all* the users on every page load.
2. The read states for private and public topics are handled differently for no reason.

PR #253, opened 2016-04-19
Refactor read state and improve performance

glebm:
Read state is now handled the same way for private and regular topics.

Only loads the read state for the current user, resolving #243.

Adds a new (Private)Topic.unread(user) methods that returns a relation
with all unread topics. We can now show the count of unread topics in a
messageboard if we want to.

Fixes last page tracking.

Refactors decorators a bit and removes MessageboardDecorator (not used).


app/controllers/thredded/topics_controller.rb
```diff
@@ -3,7 +3,7 @@ module Thredded
   class TopicsController < Thredded::ApplicationController
     before_action :thredded_require_login!,
                   only: [:edit, :new, :update, :create, :destroy]
-    before_action :update_user_activity
+    after_action :update_user_activity
```
jayroh: ha. this should have been this way the entire time, now thatI look at it. Wouldn't be surprised if this was a big part of the read-tracking bug

app/jobs/thredded/activity_updater_job.rb
```diff
@@ -4,10 +4,10 @@ class ActivityUpdaterJob < ::ActiveJob::Base
     queue_as :default
 
     def perform(user_id, messageboard_id)
-      now = Time.zone.now
+      now = Time.current
 
-      user_detail = Thredded::UserDetail.for_user_id(user_id)
-      user_detail.update_column(:last_seen_at, now)
+      user_detail = Thredded::UserDetail.where(user_id: user_id).first_or_initialize
+      user_detail.update!(last_seen_at: now)
```
jayroh: `update!` will save the object if it's a new record and update it if it's an existing record?

jayroh: I think I know the answer to this - but just checking. It's an interesting behavior.

jayroh:
This looks great @glebm - nice work.

Looking forward to trying this against the db at thredded.com

app/controllers/thredded/topics_controller.rb
glebm: It just improves performance a bit, because currently the `:async` ActiveJob adapter runs the jobs in the same process (in a different thread, but same process, so it blocks on MRI).

app/jobs/thredded/activity_updater_job.rb
glebm: Yes, `update!` assigns the attributes and calls `save!`.

PR #263, opened 2016-04-27
Add Messageboard Group.

hisea:


config/routes.rb
```diff
@@ -23,6 +23,7 @@
 
   resource :preferences, only: [:edit, :update]
   resource :messageboard, path: 'messageboards', only: [:new]
+  resources :messageboard_groups, only: [:new, :create]
```
glebm: Let's scope all admin routes under `scope path: 'admin' do ... end` and add `admin` to [reserved words](https://github.com/thredded/thredded/blob/master/app/models/thredded/messageboard.rb#L8) to avoid a possible conflict with a messageboard slug. Also the `admin` routes need to be moved above (e.g. line 16) to be matched before the messageboard route.

app/controllers/thredded/messageboard_groups_controller.rb
```diff
@@ -0,0 +1,28 @@
+module Thredded
+  class MessageboardGroupsController < Thredded::ApplicationController
+
+    def new
+      @messageboard_group = MessageboardGroup.new
+      authorize_creating @messageboard_group
```
glebm: `authorize_creating` will try to throw a custom error class that doesn't exist. We don't really need a custom exception class here, so let's just use `authorize @messageboard_group, :create?`

app/views/thredded/messageboards/index.html.erb
```diff
@@ -3,9 +3,22 @@
 <% content_for :thredded_breadcrumbs, render('thredded/shared/breadcrumbs') %>
 <%= thredded_page do %>
   <section class="thredded--main-section thredded--messageboards">
-    <%= render @messageboards %>
+    <h3> Top Level Board </h3>
```
glebm: Probably best to not render a title for top level boards because the categories are optional and it will look weird if there are no other categories.

app/models/thredded/messageboard.rb
```diff
@@ -33,8 +33,13 @@ class Messageboard < ActiveRecord::Base
              through:    :recently_active_user_details,
              source:     :user
 
+    belongs_to :thredded_messageboard_group
```
glebm: Since we are in the `Thredded` module, the association name should just be `messageboard_group`.
Also it's always good to add `inverse_of` so that certain calls don't perform extra queries: `inverse_of: :messageboards`.

app/models/thredded/messageboard_group.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+module Thredded
+  class MessageboardGroup < ActiveRecord::Base
+    has_many :messageboards
```
glebm: If a category is destroyed, the foreign key should be nullified, so `dependent: :nullify`. Also, `inverse_of: :messageboard_group`.

glebm:
Looking good so far! You're on the right track! I've added a few comments.

Additionally, it'd be great if all strings in the templates were i18n'd. We're not 100% i18n yet, but we've are moving all the strings to i18n whenever we touch a template.

Migration should live in `db/upgrade_migrations/######_upgrade_v0_3_to_v0_4.rb`, and also the `db/migrate/#####_create_thredded` migration should be modified with the new tables/keys.

app/controllers/thredded/messageboards_controller.rb
```diff
@@ -3,10 +3,12 @@ module Thredded
   class MessageboardsController < Thredded::ApplicationController
     def index
       @messageboards = thredded_current_user.thredded_can_read_messageboards
```
glebm: We can just have a single instance variable:

```ruby
@messageboards_by_group = thredded_current_user.thredded_can_read_messageboards.preload(:group).group_by(&:group)
```

hisea:
Thanks for your feedback. will update tonight.

db/upgrade_migrations/20160410111522_upgrade_v0_2_to_v0_3.rb
```diff
@@ -26,6 +26,15 @@ def up
     end
     remove_column :thredded_private_users, :read
     remove_index :thredded_private_users, name: :index_thredded_private_users_on_read
+
```
glebm: v0.3 is already out, so this should be in upgrade_v0_3_to_v0_4

db/migrate/20160329231848_create_thredded.rb
```diff
@@ -164,6 +164,15 @@ def change
         t.timestamp :read_at, null: false
       end
       add_index table_name, [:user_id, :postable_id], name: :"#{table_name}_user_postable", unique: true
+
+      create_table :thredded_messageboard_groups do |t|
+        t.string :name
+        t.timestamps
```
glebm: `null: false`

db/migrate/20160329231848_create_thredded.rb
```diff
@@ -164,6 +164,15 @@ def change
         t.timestamp :read_at, null: false
       end
       add_index table_name, [:user_id, :postable_id], name: :"#{table_name}_user_postable", unique: true
+
+      create_table :thredded_messageboard_groups do |t|
```
glebm: Looks like this accidentally ended up inside an `.each` loop above.

hisea:
@glebm Sorry for the back and forth. updated the create_thredded migration. should be out of the each loop now.

glebm:
No worries about the back and forth, this is the way it's supposed to work :) I'll have a look later tonight

glebm:
There are a few failing tests, and the 02->03 migration should be reverted (as per my comment), looking good otherwise.

app/controllers/thredded/messageboards_controller.rb
```diff
@@ -2,11 +2,15 @@
 module Thredded
   class MessageboardsController < Thredded::ApplicationController
     def index
-      @messageboards = thredded_current_user.thredded_can_read_messageboards
+      @groups = thredded_current_user
+               .thredded_can_read_messageboards
+               .preload(:group).group_by(&:group)
+               .collect{ |group| GroupView.new(*group) }
```
glebm: nit: Instead of `|group|`, use `|(group, messageboards)|`.

app/view_models/thredded/group_view.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+module Thredded
+  # A view model for a page of PostViews.
+  class GroupView
```
glebm: nit: a more specific name such as `MessageboardsGroupView` or `GroupMessageboardsView` would be better. Also, the comment.

hisea:
I fixed the failed tests locally and all passing(with rake spec), 
but still fails on CI. do I have to run anything special?

glebm:
Strange that the tests are passing for you locally. Are they passing with `rspec` as well? The failure as reported by `rspec` is:

```
  1) I18n does not have missing keys
     Failure/Error:
       expect(missing_keys).to be_empty,
         "Missing #{missing_keys.leaves.count} i18n keys, run `i18n-tasks missing' to show them"
     
       Missing 3 i18n keys, run `i18n-tasks missing' to show them
```

Looks like we're missing translations for the newly added Portuguese keys. I think @ramirovjr can take care of these once this is merged.

Also, please run `rubocop`, fix, and rebase.

glebm:
It's not failing locally because the CI is running it against the merged version. It will be failing once you rebase.

hisea:
thanks @glebm, I did rebase, fixed all rubocop except one:

```
db/migrate/20160329231848_create_thredded.rb:2:1: C: Class has too many lines. [157/154]
class CreateThredded < ActiveRecord::Migration ...
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

185 files inspected, 1 offense detected
```

I guess we'll have to leave this one as is?

ramirovjr:
Yes, just ping me when it's merged that I'll add the translation and submit the PR.

glebm:
@hisea Made some fixes (referenced above), and merged! Thanks! :tada:

Issue #271, opened 2016-05-01
seems to depend on devise-like logout path -- needs doc or config


timdiggins:
When installing it seems to depend on the logout path destroy_user_session_path -- I'm assuming this is part of an implicit expectation of using Devise, which is actually fine, but think this expectation needs a doc, or if not, needs an explicit config in the thredded.rb initializer

jayroh:
Hey @timdiggins thanks for reporting this. This is most likely a remnant from its previous life as a full rails app and just never bubbled it's way up until now.

I'm not at my computer currently so I can't search - could you let us know where in particular you're encountering this issue?

glebm:
It's in the standalone nav partial. We need to document that this partial needs to be overriden if not using devise

glebm:
Now documented.

zapnap:
As an alternative... I simply added a route to comply with the required devise route names (e.g. destroy_user_session_path) and this initially appears to work.

However, after some time it began triggering an undefined method error for ActionDispatch. I wasn't sure why this was happening at first and was scratching my head for several days but now it appears that it happens *only* after the RailsAdmin engine mount point is hit. Very strange. Looks like this is doc'd here also https://github.com/thredded/thredded/issues/280

timdiggins:
@zapnap think it's simpler just to override the partial -- there are instructions in the README: https://github.com/thredded/thredded#standalone-layout

@glebm maybe the devise view (the inside of the `li. thredded--user-navigation--standalone`) should be in a separate partial, and not be referenced directly from _standalone, instead just come in as via a view_hook in the initializer (maybe by default) maybe?

zapnap:
Same issue with overriding the partial. Something to do with RA clobbering the referenced routes? Strange. Will need to dig in further.

glebm:
A few diagnostic questions:
1. Are you using thredded v0.8.2?
2. Have you considered using Thredded [with your application layout](https://github.com/thredded/thredded#application-layout) instead of standalone?
3. What is the exact error message?

Also, when referencing main_app routes, they must be referenced via <code><b>main_app.</b>my_path</code>.

zapnap:
[...]

PR #273, opened 2016-05-01
[JRO] Add note about devise assumption in view partial

jayroh:
Specifically `app/views/thredded/shared/nav/_standalone.html.erb`

### DIFF
--- a/README.mkdn
+++ b/README.mkdn
@@ -165,6 +165,14 @@ Whenever possible, use the styles and i18n to customize Thredded to your needs.
 Thredded views also provide two content_tags available to yield - `:thredded_page_title` and `:thredded_page_id`.
 The views within Thredded pass those up through to your layout if you would like to use them.
 
+#### View Assumption(s)
+
+If you are using the standalone thredded layout (the layout included with this gem -
+`thredded/application`) the navigation makes the assumption that you are using devise as your auth
+library. If you are using something different you may override the partial
+`app/views/thredded/shared/nav/_standalone.html.erb` and use the appropriate log in / sign out path
+helpers.
+
 ### Emails
 
 Thredded sends several notification emails to the users. You can override in the same way as the views.


jayroh:
@timdiggins I've added some copy in the README regarding issue #271. Would like to know your thoughts as to whether this is sufficient.

I'm wondering at this point 

1. If that partial is particularly necessary. Could things be just as effective without it?
2. Is that partial named appropriately? It contains an `LI` element that houses a log in/out link, but it is called `app/views/thredded/shared/nav/_standalone.html.erb`. Should it, instead, be something like...?

`app/views/thredded/shared/nav/_authentication.html.erb`
`app/views/thredded/shared/nav/_standalone_authentication.html.erb`

timdiggins:
@jayroh thanks for fast response, and this does fix the problem I was experiencing, however because it's called from within the engine, it appears I have to call path helpers with the `mainapp` prefix in that override, so maybe would be worth indicating that -- I'll add a comment to your commit.

With regard to naming, I think `app/views/thredded/shared/nav/_authentication.html.erb` is clearer, however it might help to use the word "standalone" as a more explicit namespace: the docs would be clearer if the thredded builtin layout was named `thredded/standalone` (as then it's easy to refer to in the docs) and maybe standalone-layout partials were in their own directory, so: `app/views/thredded/standalone/nav/_authentication.html.erb`. However this might be too much churn for you.

glebm:
I didn't expect many people to actually use the standalone layout, but I guess it's the fault of the Readme that people do.

In the "Install" steps, we currently don't emphasize setting a layout, but we instruct people to add Thredded CSS + JS to their styles although that's not necessary when using the standalone layout. I think the installation instructions should have two options: App Layout and Thredded Standalone Layout, and the latter one is where the devise bit should go.

With regards to the view partial, yeah, `_standalone_authentication` may be a better name.
Authentication links seem pretty important to leave out.

Tangentially, since Thredded does not have a view hooks system (e.g. [like this one](https://github.com/glebm/rails_email_preview#hooks)) and  the only way to extend the views is by overriding partials, perhaps we should additionally provide empty `shared/nav/before` and `shared/nav/after` partials that would allow adding more links before/after the navigation.

glebm:
@timdiggins The `main_app` prefix should not be necessary if using a regular route. Thredded delegates all the routes to the main_app automatically.

README.mkdn
```diff
@@ -165,6 +165,14 @@ Whenever possible, use the styles and i18n to customize Thredded to your needs.
 Thredded views also provide two content_tags available to yield - `:thredded_page_title` and `:thredded_page_id`.
 The views within Thredded pass those up through to your layout if you would like to use them.
 
+#### View Assumption(s)
+
+If you are using the standalone thredded layout (the layout included with this gem -
+`thredded/application`) the navigation makes the assumption that you are using devise as your auth
+library. If you are using something different you may override the partial
+`app/views/thredded/shared/nav/_standalone.html.erb` and use the appropriate log in / sign out path
+helpers.
```
timdiggins: Think it needs to be slightly more urgent ('need to') and would be good to indicate the need to use mainapp: 
> If you are using something different you will need to override the partial
`app/views/thredded/shared/nav/_standalone.html.erb` and use your appropriate log in / sign out path helpers, e.g:  `<li><%= link_to "Login", mainapp.login_path %></li>`

timdiggins:
@glebm Hmm, it was for the legacy app I was testing with (it uses a non-resourceful path), but I'll have another look now

timdiggins:
No, I seem to need the main_app prefix for any path helper I use (whether route is resourceful or not). 

FYI Additionally, when using the mainapp's own layout, I discovered that while path helpers were fine, antiue-style hash-based paths (`url_for(:controller => 'x', :action => 'y')` and link_to equivalent) didn't work (were only looking in the thredded engine).

glebm:
Looks like we're only including the MainAppRouteDelegator for non-standalone layouts: https://github.com/thredded/thredded/blob/d5282138633a9f4b140c02cb5053f16c3411f8a1/lib/thredded/engine.rb#L22-L25

I'll change that

glebm:
I'll merge this and then "refactor" the layout stuff in the Readme in a follow-up commit

jayroh:
> I'll merge this and then "refactor" the layout stuff in the Readme in a follow-up commit

Nice, Thanks @glebm

PR #283, opened 2016-05-05
[JRO] Better style quotes + better word wrap

jayroh:
Quotes need a little bit more differentiation from the rest of post
content copy and can use a little styling.

In addition - `word-wrap: break-all` is a bit over-eager when it comes to
wrapping so changing it to `break-word` smoothes that out a little bit.

### DIFF
--- a/app/assets/stylesheets/thredded/base/_typography.scss
+++ b/app/assets/stylesheets/thredded/base/_typography.scss
@@ -39,3 +39,12 @@
   fill: currentColor;
 }
 
+%thredded--quote {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/thredded/base/_typography.scss
```diff
@@ -39,3 +39,12 @@
   fill: currentColor;
 }
 
+%thredded--quote {
```
glebm: `thredded--blockquote`, because "quote" usually refers to the inline <q> element.

PR #286, opened 2016-05-06
[JRO] Add spoiler tag(s)

jayroh:
A feature quickly reported as missing from thredded.com was the
spoiler(s) tag. Wherever there may be lots of TV or Movie talk - this
tag will be almost necessary.

### DIFF
--- a/app/assets/stylesheets/thredded/components/_post.scss
+++ b/app/assets/stylesheets/thredded/components/_post.scss
@@ -77,4 +77,15 @@
   blockquote {
     @extend %thredded--blockquote;
   }
+
+  .spoiler {
+    color: $thredded-spoiler-hidden-color;
+    background-color: $thredded-spoiler-hidden-color;
+    cursor: pointer;
+
+    &:hover,
+    &:focus {
+      color: $thredded-spoiler-shown-color;
+    }
+  }
 }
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/thredded/components/_post.scss
```diff
@@ -77,4 +77,15 @@
   blockquote {
     @extend %thredded--blockquote;
   }
+
+  .spoiler {
```
glebm: This should be namespaced, e.g. `.thredded--post--content--spoiler`.

glebm:
Perhaps this should use the HTML5 details/summary tags to be screen-reader friendly?

Interactive example:

<details>
	<summary>Click to show spoiler</summary>
	<p>Spoiler content here</p>
</details>

Source:

```html
<details>
	<summary>Click to show spoiler</summary>
	<p>Spoiler content here</p>
</details>
```

Documentation: https://developer.mozilla.org/en/docs/Web/HTML/Element/details

app/assets/stylesheets/thredded/components/_post.scss
jayroh: Good point Gleb. I'll make that change.

jayroh:
I'm on the fence on this one. On the one hand - I know this is the right thing to do. On the other hand, the current implementation is a nice clean one.

I think the right thing, with the right solution, is probably the best way to handle it though. I'll look around for a nice and screen-reader friendly solution that can sit atop the example you give above.

app/assets/stylesheets/thredded/components/_post.scss
jayroh: PR #287 open with this change.

PR #292, opened 2016-05-11
[JRO] Auto-correct rubocop errors

jayroh:
Travis is using rubocop 0.40.0 and there's a new cop there that doesn't like where some of our end-parens are. Things here in this PR have been fixed with `rubocop -a`.


TheDudeWithTheThing:
nice, I do this anyway

👍

PR #293, opened 2016-05-11
[JRO] Add empty partials before/after post textarea

jayroh:
These two partials are merely for theming/overriding/extensibility for
anyone who would like to add some functionality around the common posts
textarea form element. Use cases for this include:

Adding some sort of wysiwyg js boilerplate, adding buttons, links, or
instructions for what markdown/bbcode is supported.

NOTE: This is an initial step before implementing a more robust view
customization API for thredded.

### DIFF
--- a/app/views/thredded/posts/_content_field.html.erb
+++ b/app/views/thredded/posts/_content_field.html.erb
@@ -1,4 +1,6 @@
 <li>
   <%= form.label :content, content_label %>
+  <%= render 'thredded/posts/before_content_field', form: form %>
   <%= form.text_area :content, { rows: 5, required: true }  %>
+  <%= render 'thredded/posts/after_content_field', form: form %>
 </li>
--- a/app/views/thredded/posts/_content_field.html.erb
+++ b/app/views/thredded/posts/_content_field.html.erb
@@ -1,6 +1,0 @@
-<li>
-  <%= form.label :content, content_label %>
-  <%= render 'thredded/posts/before_content_field', form: form %>
-  <%= form.text_area :content, { rows: 5, required: true }  %>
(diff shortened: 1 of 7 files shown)


app/views/thredded/posts/_content_field.html.erb
```diff
@@ -1,4 +1,6 @@
 <li>
   <%= form.label :content, content_label %>
+  <%= render 'thredded/posts/before_content_field', form: form %>
   <%= form.text_area :content, { rows: 5, required: true }  %>
+  <%= render 'thredded/posts/after_content_field', form: form %>
```
glebm: I would say this should be just "after_content" (not "after_content_field"), since it's not a given that a field will be rendered there, and that this should be documented, but since this is an interim solution, LGTM anyway.

jayroh: I like your reasoning for the name - great point. And thanks for the
reminder on the documentation part. I got lost bouncing around branches
fixing rubocop stuff and forgot to take care of that part :)

jayroh: @glebm thinking about this a little more. What do you think about `[before|after]_textarea` for these partials instead? Because `[before|after]_content` could be misconstrued as being, literally, before and after a post's contents.

glebm: Maybe `posts/form/before_content`?

jayroh: I think that's sufficient - surely.

jayroh:
@glebm renamed the two new partials and moved the content field partial to the same path so that things are reliably grouped. What do you think?

glebm:
:+1:

PR #301, opened 2016-05-16
Permalink URLs for posts and private posts

glebm:
Currently we send emails with the URL to the page of the topic where the
post is at. However, when posts are deleted the page may change.

This introduces permalinks for posts and private posts that always
redirect to the correct page.

### DIFF
--- a/app/models/thredded/messageboard.rb
+++ b/app/models/thredded/messageboard.rb
@@ -6,7 +6,16 @@ module Thredded
                 use:            [:slugged, :reserved],
                 # Avoid route conflicts
                 reserved_words: ::Thredded::FriendlyIdReservedWordsAndPagination.new(
-                  %w(messageboards preferences private-topics autocomplete-users theme-preview admin)
+                  %w(
+                    admin
+                    autocomplete-users
+                    messageboards
+                    posts
+                    preferences
+                    private-posts
+                    private-topics
+                    theme-preview
+                  )
                 )
 
     validates :name, uniqueness: true, length: { maximum: 60 }, presence: true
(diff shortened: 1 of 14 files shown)


config/routes.rb
```diff
@@ -2,7 +2,8 @@
 Thredded::Engine.routes.draw do
   resource :theme_preview, only: [:show], path: 'theme-preview' if %w(development test).include? Rails.env
 
-  page_constraint = { page: /[1-9]\d*/ }
+  positive_int = /[1-9]\d*/
+  page_constraint = { page: positive_int }
```
jayroh: love this

app/models/thredded/messageboard.rb
```diff
@@ -6,7 +6,8 @@ class Messageboard < ActiveRecord::Base
                 use:            [:slugged, :reserved],
                 # Avoid route conflicts
                 reserved_words: ::Thredded::FriendlyIdReservedWordsAndPagination.new(
-                  %w(messageboards preferences private-topics autocomplete-users theme-preview admin)
+                  %w(messageboards preferences private-topics private-posts posts autocomplete-users
+                     theme-preview admin)
```
jayroh: This array has gotten big enough where it's a little bit of mental overhead to parse it on a quick look. Could you break these out into their own lines? It'd make diffs cleaner too.

jayroh:
Looks great @glebm - just a couple comments in there.

app/models/thredded/messageboard.rb
glebm: Good call, done.

Issue #45, opened 2014-11-23
Moderation and Spam control


glebm:
Perhaps I am missing something, but I couldn't find a way to pre-moderate (a la forem) or even delete a topic (I have set `user_details.superadmin` to true).

PR #304, opened 2016-05-22
A basic moderation system

glebm:
Adds a basic moderation system to Thredded! :boom: 

Readme description:

> Thredded comes with two options for the moderation system:
> 
> 1. Pre-emptive moderation, where posts from first-time users are not published until they have been approved.
> 2. Reactive moderation, where posts from first-time users are published immediately but enter the moderation queue.
> 
> This is controlled by the `Thredded.content_visible_while_pending_moderation` setting.
> 
> Users, topics, and posts can be in one of three moderation states: `pending_moderation`, `approved`, and `blocked`. By default, new users are `pending_moderation`, and new posts and topics
inherit their default moderation_state from the user's.
>
> This means that when you approve a new user's post, all of their later posts will be approved automatically.
> 
> Additionally, users always see their own posts regardless of the moderation state. For blocked users, this means they might not realize they have been blocked right away.
> 
> Blocked users cannot send private messages.

TODO: Counters should count only approved content. This is rather difficult to do, so let's postpone until we add counter_culture.

Refs #45

---

Moderation queue:
moderation-1

---

Moderation queue after approving a post (on mobile):
moderation-mobile

---

Moderation history:
moderation-history

---

Pre-emptive moderation user notice:

moderation-post

/cc @timdiggins @jayroh


glebm:
Please ignore the ❌, it's just CodeClimate over-reacting.

timdiggins:
@glebm This looks very cool. Some thoughts:

1) If I want to not do any moderation in an installation of the engine*, can I just ignore it safely, or should there be a setting to turn moderation off. Or alternatively a way to indicate that some users are "pre-moderated" (a user method, which could default to using the user extender)

2) I ran this with the seed data and it showed up a couple of problems: need some pagination on the moderation page I think, and some preloading in the controller. And maybe need to adjust seed-data? 

*OT: I have a possibly atypical use case / set of use cases of Thredded, maybe need to discuss this in a different strand: https://www.thredded.com/meta/a-use-case-of-thredded

glebm:
@timdiggins 

1) The default for the moderation system is "reactive", meaning all the posts pending moderation will be visible immediately. If you want you can also hide the "Moderation" link from the menu it will be exactly the same as before.

The "pre-approved" indicator is a matter of the default value of the "Thredded::UserDetail" column. If you `change_column_default` to "approved", all users will be approved by default, and the moderation queue will always be empty. Alternatively, you can change the default to `nil` and decide it on the per-user basis with a `before_save` on the `Thredded::UserDetail` model.

2) :+1: on pagination, and adjusting the seed data. Will get back when done.

glebm:
I also plan to improve the UI and add a users tab to the moderation in follow-up PRs.

jayroh:
Wow this is looking really freaking good so far!

@glebm give a shout when you think this is pending a final review. Sounds like you have some items still pending implementation/tweaking.

glebm:
All the pending items are now done. Also the readable entities are now resolved via Policy::Scope classes, making it easier to extend. @jayroh @timdiggins This is good to review! :ship:

app/models/concerns/thredded/content_moderation_state.rb
```diff
@@ -30,16 +30,16 @@ module ContentModerationState
...
         !blocked?
       else
         approved?
       end
     end
 
-    def visible_to_user?(user)
-      visible_to_all? || (!user.thredded_anonymous? && user_id == user.id)
+    def moderation_state_visible_to_user?(user)
+      moderation_state_visible_to_all? || (!user.thredded_anonymous? && user_id == user.id)
```
timdiggins: @glebm can't you just do `moderation_state_visible_to_all? || user_id == user.id` because `NullUsers`'s `user_id` is nil, and `user.id` will never be nil

app/models/concerns/thredded/content_moderation_state.rb
```diff
@@ -30,16 +30,16 @@ module ContentModerationState
       end)
     end
 
-    def visible_to_all?
+    def moderation_state_visible_to_all?
```
timdiggins: @glebm I found this a bit confusing at first (I read it as "is Moderation State (whether moderated or not)visible to all / user" 

Of course it really means "is it (due to moderation state) visible to user". It would be nice but I don't think it's possible to crisply pack that meaning into the method names. Maybe comment(s) such as following might help a future me:

```ruby
  #determines if this should be visible to anyone based on the moderation state
  def moderation_state_visible_to_all?
...
  #determines if this should be visible to a specific user based on the moderation state`
  def moderation_state_visible_to_user?(user)
```

README.mkdn
```diff
@@ -381,22 +381,31 @@ For an example of how to do this, see the initializer.
...
 
 Blocked users cannot send private messages.
 
+### Disabling moderation
+
+To disable moderation, e.g. if you run internal forums that do not need moderation, run the following migration:
+
+```ruby
+change_column_default :thredded_user_details, :moderation_state, 1 # approved
+```
```
timdiggins: Perfect thanks!

timdiggins:
@glebm I gave this branch a workout in the browser (both standalone and within my main_app)when you released the first commit. I haven't with the extra commits, but I have reviewed them and it seems great to me.

app/models/concerns/thredded/content_moderation_state.rb
glebm: If a user that posted a post is deleted, the post's `user_id` will be `nil`.

app/models/concerns/thredded/content_moderation_state.rb
glebm: Done

glebm:
@timdiggins I've addressed your comments, thanks!

glebm:
@jayroh Rebased and should be good to merge!

glebm:
Actually, just going to go ahead and merge this (I'm on a merging spree today :smile: ). This PR is getting too big, and any concurrent PRs will necessarily have merge conflicts. So it's better to merge this now and address comments if any in subsequent PRs.

jayroh:
Good with me man. There was nothing I saw in there that was cause for alarm
so - good move

PR #308, opened 2016-05-24
Docs and script for setting up a dummy server

timdiggins:
There may be a better way to do this, but this was helpful to me when trying out the PR.


glebm:
There is already documentation at https://github.com/thredded/thredded#development.

PostgreSQL is not necessary as the dummy app uses sqlite by default. However, if you want to use PostgreSQL or MySQL, there is the `script/create-db-users` (also documented above).

The dev server can be started with `rake dev:server`, also documented above.

timdiggins:
Ok not sure why I didn't spot this / use this

Issue #284, opened 2016-05-05
Thinking about @mention notification and highlighting


timdiggins:
If I have read the codebase correctly it seems like @mention notification is set up to just queue an email when an @ mention is found in a created post.

However I would like to especially highlight unread posts which at mention the viewer. (so the unread posts are highlighted in one way and the mentioning unread posts are highlighted in another).

Additionally, it might be that the mechanism of notification is different from a single email -- e.g. post a notification via slack or similar channel.

Is some kind of abstraction layer  for notification possible/envisage?

jayroh:
@timdiggins regarding the first part. I think it's doable, sure, but envision that it could/should probably be done asynchronously on the front-end. I think we have several involved queries going on in the current topics#index controller/action that would have to have another layer added atop if we were to bubble up whether or not you (the rhetorical you) were mentioned in that topic.

Or do you mean in topics#show - where the posts are listed?

* * *

As for the methods by which people are notified - I've started looking into abstracting that out to use [ActiveSupport:Notifications](http://api.rubyonrails.org/classes/ActiveSupport/Notifications.html) to make it easier to subscribe to those types of events and react appropriately. The @ mention and Private Topic notifications would probably end up being refactored to use that mechanism instead of being called from controllers.

timdiggins:
@jayroh I guess I was thinking more generally than a particular interface that the most important thing is "where am I mentioned" -- this is as important as "private messages to me" and could be (depending on the style/kind of forum) more important than new topics.

So maybe this is more like a cross cutting concern. It's very like private messages... something like "unread mentions of me" -- could be implemented by a model which holds "mentions of a person in a post" and can then marked as read when seen by that person.

[Of course the fact that posts are  editable makes this a bit more complicated, haven't thought that through].

----

as to the second part,  I think that having some kind of notifications abstraction is great. I haven't used ActiveSupport:Notifications explicitly so I'll look into that. 
I'm thinking in general about notifications not necessarily in real time/near real time, but more being able to generate emails or other status notifications based on user preferences (e.g. An email with a list of topics mentioning you that you haven't read (in the last hour/day or similar)).

----

[...]

PR #310, opened 2016-05-28
Follow and notify

timdiggins:
Addresses the first plan in https://github.com/thredded/thredded/issues/284#issuecomment-221548470

You can follow a topic manually, or automatically when you start, reply, or are mentioned in a post. 

Email notification now happens for all followed topics and private messages you are a member of (no longer dependent on being mentioned).

examples

----

Comments invited @jayroh @glebm 

Features missing from the original plan:

* show follow status in the list of topics within a messageboard -- however I'm not 100% sure this is needed as part of **this** PR ( I would like to change the display of topics to be more 
* have a global or per-messageboard preference for sending emails for followed topics (on by default) -- not -- however i'm not 100% sure if this is needed

Things still needing cleaning up:

* need to sort out the upgrade_migrations (but need to have a target version number in order to sort this)
* squash everything into one commit (leaving the commits, so you can see the intentions for now).


jayroh:
Hi @timdiggins  :). Hi @EllaNancyFay :)

If you don't mind I'm going to make some small readability and style suggestions before digging into the code. I'm a stickler for vertical rather than horizontal readability in code so ... please bear with me 😄

app/commands/thredded/autofollow_mentioned_users.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def autofollows
+      user_names = Thredded::AtNotificationExtractor.new(post.content).run
+      members = post.readers_from_user_names(user_names).to_a
+
+      members.delete post.user
+      members = exclude_those_opting_out_of_at_notifications(members)
+
+      members.each do |user|
+        Thredded::UserTopicFollow.create_unique(user.id, @post.postable_id, Thredded::UserTopicFollow::REASON_MENTIONED)
```
jayroh:
```ruby
Thredded::UserTopicFollow.create_unique(
  user.id,
  post.postable_id,
  Thredded::UserTopicFollow::REASON_MENTIONED
)
```

That `@post` can turn into `post` since it's an instance attribute per the `attr_reader`

app/commands/thredded/autofollow_mentioned_users.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    private
+
+    attr_reader :post
+
+    def exclude_those_opting_out_of_at_notifications(members)
+      members.select do |member|
+        member.thredded_user_preference.notify_on_mention? && member.thredded_user_messageboard_preferences.in(post
+            .messageboard).notify_on_mention?
```
jayroh:
```ruby
member.thredded_user_preference.notify_on_mention? &&
  member.thredded_user_messageboard_preferences.in(post.messageboard).notify_on_mention?
```

What do you think of this sort of approach? I comment here becuase the current rhythm breaks at something of an odd spot for me. What do you think?

app/commands/thredded/autofollow_mentioned_users.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+module Thredded
+  class AutofollowMentionedUsers
+    def initialize(post)
+      @post = post
+    end
+
+    def run
+      members = autofollows
+      return unless members.present?
```
jayroh: I'm unsure whether this return needs to exist here because the work has already been done.

If I'm reading this right, we'd want the guard clause at the beginning to ensure that we're not doing any subsequent work *if* there are no members present - is that right? If so, what about:

```ruby
def run
  return if members.empty?
  autofollow_members
end

private

def members
  user_names = Thredded::AtNotificationExtractor.new(post.content).run
  members = post.readers_from_user_names(user_names).to_a
  members.delete(post.user)
  members = exclude_those_opting_out_of_at_notifications(members)
  members
end

def autofollow_members
  Thredded::UserTopicFollow.create_unique(
    user.id,
    post.postable_id,
    Thredded::UserTopicFollow::REASON_MENTIONED
  )
end

File: app/commands/thredded/notify_following_users.rb [hunk 4]
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+module Thredded
+  class NotifyFollowingUsers
+    def initialize(post)
+      @post = post
+    end
+
+    def run
+      return unless targetted_users.present?
```
jayroh: I've gotten into the habit of erring on the side of using `if` instead of `unless` where the possibility is there. In this case we can represent this line similarly with

```ruby
return if targeted_users.empty?
```

Also - ~~targetted~~ => targeted. See [here](http://www.future-perfect.co.uk/grammar-tip/is-it-targetted-or-targeted/)

app/jobs/thredded/auto_follow_and_notify_job.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+module Thredded
+  class AutoFollowAndNotifyJob < ::ActiveJob::Base
+    queue_as :default
+
+    def perform(post_id)
+      post = Post.find(post_id)
+
+      AutofollowMentionedUsers.new(post).run
+      NotifyFollowingUsers.new(post).run
```
jayroh: This looks sooooooo niiiiiice!

jayroh: Ruby I <3 you. What a beautiful language

app/models/thredded/post.rb
```diff
@@ -30,5 +32,12 @@ def readers_from_user_names(user_names)
         .new(Thredded.user_class.thredded_messageboards_readers([messageboard]), Thredded.user_name_column)
         .in(user_names)
     end
+
+    def auto_follow_and_notify
+      # need to do this in-process so that it appears to them immediately
+      UserTopicFollow.create_unique(user.id, postable_id, UserTopicFollow::REASON_POSTED)
+      # everything else can happen later
+      AutoFollowAndNotifyJob.perform_later(id)
+    end
```
jayroh: 👍

spec/dummy/config/environments/test.rb
```diff
@@ -10,6 +10,6 @@
   config.action_controller.allow_forgery_protection = false
   config.action_mailer.default_url_options = { host: 'localhost:3000' }
   config.action_mailer.delivery_method = :test
-  config.action_mailer.perform_deliveries = false
+  config.action_mailer.perform_deliveries = true
```
jayroh: Was this causing emails not to populate the ActionMailer emails array during tests? Eesh that's a rough one to have been missing all this time

spec/features/thredded/user_follows_topic_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  let(:user) { PageObject::User.new(create(:user)) }
+  let(:an_unfollowed_topic) do
+    topic = create(:topic, with_posts: 1, messageboard: create(:messageboard))
+    PageObject::Topic.new(topic)
+  end
+  let(:a_followed_topic) do
+    topic = create(:topic, with_posts: 1, messageboard: create(:messageboard))
+    Thredded::UserTopicFollow.create_unique(user.user.id, topic.id)
+    PageObject::Topic.new(topic)
+  end
```
jayroh: Would you be ok with putting these `let` at the top of the feature? I'm a fan of the [xunit 4-phase style](http://xunitpatterns.com/Four%20Phase%20Test.html) where it's read 

1. setup
2. exercise
3. verify
4. tear-down

If this test grows much larger and these let's get pushed down to the bottom of the file/buffer it *might* end up with a bit of a [mystery guest](https://robots.thoughtbot.com/mystery-guest) smell.

jayroh:
So excited to get this rolled out!

app/commands/thredded/autofollow_mentioned_users.rb
timdiggins: Agreed @jayroh not clear -- need to split out the work from the extraction of who to perform the work on. Will adjust (along with other tweaks) in a commit

spec/dummy/config/environments/test.rb
timdiggins: @jayroh actually, there weren't any tests of deliveries per se, only mocking tests (or tests of delivery side-effects). I found these too much work, easier to let it happen and test the deliveries.

app/commands/thredded/notify_following_users.rb
```diff
@@ -6,13 +6,13 @@ def initialize(post)
     end
 
     def run
-      return unless targetted_users.present?
-      PostMailer.post_notification(@post.id, targetted_users.map(&:email)).deliver_now
-      MembersMarkedNotified.new(@post, targetted_users).run
+      return if targeted_users.empty?
+      PostMailer.post_notification(@post.id, targeted_users.map(&:email)).deliver_now
+      MembersMarkedNotified.new(@post, targeted_users).run
```
timdiggins: @jayroh I'm wondering if we should actually be doing this. Should we perhaps send separate notifications to each person and mark them as notified. Otherwise we're divulging email addresses to all followers I think (ok in my use case, not in all). Could be wrapped up in one job/method, but per user

jayroh: How does this in particular open email addresses to others?  could we change this to BCC: instead of TO:

timdiggins:
So I think we're about ready to go on this one, other than squashing and fixing the migrations. 

There's two things which are unresolved:

* Possibly changing to bcc (this is not a new issue introduced in this PR though -- could solve that in a separate PR)
* Have a global (or per-messageboard) preference for sending emails for followed topics (on by default) - -- do you think we need this?

app/models/concerns/thredded/topic_common.rb
```diff
@@ -47,22 +47,30 @@ def unread(user)
           .merge(reads_class.where(reads[:id].eq(nil).or(reads[:read_at].lt(topics[:updated_at]))))
       end
 
+      def read_states_by_topics_lookup(user)
+        read_states_by_topic_id =
+          reflect_on_association(:user_read_states).klass
+            .where(user_id: user.id, postable_id: current_scope.map(&:id))
+            .group_by(&:postable_id)
+
+        def read_states_by_topic_id.get(topic, null_value = nil)
```
glebm: This flushes MRI method cache on every call to the parent method, which is very slow. Seems that you can simply use a Hash with a default block here. http://ruby-doc.org/core-2.2.0/Hash.html#method-c-new

app/models/concerns/thredded/topic_common.rb
```diff
@@ -47,22 +47,30 @@ def unread(user)
           .merge(reads_class.where(reads[:id].eq(nil).or(reads[:read_at].lt(topics[:updated_at]))))
       end
 
+      def read_states_by_topics_lookup(user)
```
glebm: Complex methods should have documented return type and arguments. Also, looks like this one should be private?

spec/i18n_spec.rb
```diff
@@ -4,7 +4,7 @@
 RSpec.describe 'I18n' do
   let(:i18n) { I18n::Tasks::BaseTask.new }
   let(:missing_keys) { i18n.missing_keys }
-  let(:unused_keys) { i18n.unused_keys }
+  let(:unused_keys) { i18n.unused_keys(strict: false) }
```
glebm: The non-strict mode is really error-prone. It's best to always use strict, specifying magic comments where necessary.

app/views/thredded/topics/_header.html.erb
```diff
@@ -9,4 +9,18 @@
     <%= link_to t('thredded.topics.edit'), topic.edit_path,
                 class: 'thredded--topic-header--edit-topic' %>
   <% end %>
+  <% if current_user %>
+    <% if follow = current_user.following?(topic)%>
+      <div class="thredded--topic-follow-info thredded--following">
+        <p>
+          <%= t("thredded.topics.following.#{follow.reason}") %></p>
```
glebm:
```erb
<% # i18n-tasks-use t('thredded.topics.following.reason1') t('thredded.topics.following.reason2') %>
```

db/migrate/20160525143440_thredded_user_topic_follows.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+class ThreddedUserTopicFollows < ActiveRecord::Migration
```
glebm: As this will be in the v0.5 release, let's move to upgrade_migrations/...v04_to_05

spec/dummy/config/environments/test.rb
glebm: In theory, the preview tests ensure that templates compile, and other tests ensure that the delivery tasks are scheduled. So perhaps there is no need for this.

app/models/concerns/thredded/topic_common.rb
timdiggins: @glebm Hash with default value won't work as we need an accessor that passes in the topic (to prevent the O(n) SQL calls on every row). Maybe the answer is to have a specific class to handle the lookup. I'll try that and you see what you think

app/models/concerns/thredded/topic_common.rb
```diff
@@ -47,24 +47,45 @@ def unread(user)
...
       end
 
+      private
+
+      # @param user [Thredded.user_class]
+      # @return [ByPostableLookup]
+      def read_states_by_topics_lookup(user, null_value)
+        read_states = reflect_on_association(:user_read_states).klass
+          .where(user_id: user.id, postable_id: current_scope.map(&:id))
+        Thredded::TopicCommon::Lookup.new(read_states, null_value)
```
glebm: Seems like this could simply return a Hash, e.g.:

```ruby
read_states.inject(Hash.new(null_value)) { |h, rs| h[rs.postable_id] = rs; h }
```

The `.postable` assignment can happen within inject or in `with_read_states` (note that multiple calls to `current_scope.map` do not result in multiple queries).

timdiggins: @glebm I see what you using the default value for the hash and the Hash.new{} version and preferable to creating the extra abstraction of the class if we don't need it.

However, I don't get what you mean by "The .postable assignment can happen within the inject". The multiple queries we are trying to avoid are the reverse lookup of topic from read_state (used I assume within the views). I haven't checked this, just assuming that's true as I saw them in the views. And perhaps right now not needed for the Follows, but easier to have them in parallel (and will probably end up using this lookup eventually).

glebm: Sorry, I meant this:

```ruby
postable_by_id = current_scope.inject({}) { |h, p| h.update(p.id => p) }
read_states.each { |rs| rs.postable = postable_by_id[rs.postable_id] }
read_states.inject(Hash.new(null_value)) { |h, rs| h.update(rs.postable_id => rs) }
```

timdiggins: @glebm do you really prefer that to the solution that's in the PR at present? If so, I'll go with it, but I  don't see it's much better (it's less clear what's going on and why to my mind).

glebm: @timdiggins I think the hash solution can be made clearer with better variable/method names and maybe a comment. I prefer it to the class one because a simple pre-populated data structure is easier to use than a custom class, because Hash methods are well-known. Either way is fine by me though.

jayroh: I'm coming in here with fresh eyes so maybe I can help with some outside perspective.

While the hash inject/reduce is certainly down closer to the metal (Ha! Ruby! Close to the metal!) it's a bit more difficult to grok. I agree that with some better naming that could be more readable but, for what it's worth, I tried to take that code snippet and expand it out with some different names and comments and I couldn't quite nail it. To be quite honest - I don't use map/reduce or collect/inject often. I'm not sure I'd be able to quite pull my thoughts together with this approach easily. I could deal if I had to, of course - but a lesser experienced rubyist *might* look at this and ["derp" themselves out the door](https://media4.giphy.com/media/1WQLKmc1Gfhny/200.gif).

I should probably admit, actually, my eyes are less than fresh after chasing my kid around for the evening. So - I'd probably vote for clarity and a well named, well factored class. The less I have to think, the better.

That being said. 

@timdiggins would you be averse to extracting the protected class into a name-spaced file and adding some tests? Considering the conversation here I'm coming mostly from the angle of a test providing both good documentation and exhibiting example behavior.

app/models/thredded/user_topic_follow.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+module Thredded
+  class UserTopicFollow < ActiveRecord::Base
+    REASON_MANUAL = 'manual'
```
glebm: How about making this an enum? http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html

app/models/concerns/thredded/topic_common.rb
timdiggins: @jayroh @glebm On reflection the utility class is doing two things -- managing the inverse lookup performance gain and also dealing with null values. If it was just dealing with the inverse lookup and was basically a hash, then it would make it a lot easier to understand and we wouldn't need so much code or documentation or specs. Am pushing a commit to show this

app/models/thredded/user_topic_follow.rb
timdiggins: @glebm agree maybe should have done it that way in the first place, but it works right now and other than saving a bit of space in the db doesn't seem to have a compelling need?

glebm: Well, it's the standard way of doing enums in Rails. Out of the box it gives you validations, scopes, and a method that enumerates over all the values.

timdiggins: @glebm yes, yes I agree. Will see if I can find time to do (but trying to squeeze in other unrelated fixes and improvements)

app/models/thredded/user_topic_follow.rb
```diff
@@ -0,0 +1,23 @@
...
+    REASON_POSTED = 'posted'
+    REASON_MENTIONED = 'mentioned'
```
glebm: You can use symbols when assigning. When checking, you can use e.g. `user_topic_follow.mentioned?`.

app/models/thredded/user_topic_follow.rb
```diff
@@ -0,0 +1,22 @@
...
+    belongs_to :topic, inverse_of: :user_follows
+
+    validates :user_id, presence: true
+    validates :topic_id, presence: true
+
+    # shim to behave like postable-related (though actually only ever related to topic)
+    alias_attribute :postable_id, :topic_id
+    alias_attribute :postable, :topic
+
+    def self.create_unique(user_id, topic_id, reason = REASON_MANUAL)
```
glebm: What is the expected behaviour of this method when a user_id/topic_id record already exists?

timdiggins: @glebm The aim was no-op (ie. create unless already exists). Maybe should change name to `create_unless_exists` to make this more obvious.

glebm: yeah, that would be a better name

glebm:
Looks good to merge to me (pending renamind `create_unique` to `create_unless_exists`). :+1: 
@jayroh ?

glebm:
Actually, just going to merge it, so that I can rebase the moderation PR. Let's address any further comments in a follow-up PR before releasing 0.5.

PR #315, opened 2016-06-02
Config whitelist & filters

lawso017:
In order to get Thredded working with the basecamp/trix editor, I had to make the following two changes:

1. Since I'm now storing html content, I needed to remove the BBContent pipeline filter -- it was clobbering the img tags inside links. So I moved the list of content_pipeline_filters to the initializer.

2. Since Trix stores uploaded images as img elements inside a figure tag nested within a link, I needed to broaden the whitelist elements to allow figure tags (and figcaption). So I moved the whitelist elements to the initializer as well.

Not sure if this is a direction you'd want to go, but thought I'd pass it on.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ Gemfile.lock
 *.gemfile.lock
 !heroku.gemfile.lock
 .bundle/
+.ruby-version
 bin/*
 !bin/rails
 log/*.log
--- a/app/models/concerns/thredded/post_common.rb
+++ b/app/models/concerns/thredded/post_common.rb
@@ -30,7 +30,7 @@ module Thredded
       add_attributes: {
         'iframe' => {
           'seamless' => 'seamless',
-          'sandbox' => 'allow-forms allow-scripts allow-same-origin',
+          'sandbox' => 'allow-forms allow-scripts',
         }
       }
     ).freeze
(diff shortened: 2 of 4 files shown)


glebm:
Moving these to the initializer doesn't make it much different from having them as the constants on the PostCommon class. Neither are very modular.

What I'd like to have in the future is a `ContentFormatter` class instead of having this functionality in `PostCommon`. Then the user can monkey-patch/configure the aspects of the default `ContentFormatter` or even use a different one.

Meanwhile:
* `figure` and `figcaption` should be whitelisted by default. 
*  You can remove the BBContent filter by reassigning the constant in the initializer.

jayroh:
Would agree on `figure` and `figcaption`.  Also +1 on the ContentFormatter class. The responsibility should belong to another class. They're a little too tightly coupled now.

lawso017:
Thanks, that makes sense. ContentFormatter sounds great.

This may be a dumb question, but how do you recommend reassigning the constant in the initializer? I'm not familiar with that technique.

glebm:
@lawso017 Sorry, it's not as simple as I thought it was, but the good news is that I've implemented the `ContentFormatter` in #321 so once that's merged it will be very easy to configure: https://github.com/thredded/thredded/pull/321/files#diff-d24aa57ddf38369b21abef6c31f43cae.

PR #319, opened 2016-06-03
fix problem with <a> without href, update related post factories to check that

tomczakdominik:


glebm:
Since `<a>` without href is invalid HTML, we should rather make it valid (like it does currently), or remove the tag completely keeping only the contents. Closing for now.

spec/factories.rb
```diff
@@ -145,3 +146,9 @@
     page 1
   end
 end
+
+def post_content_with_tags
+  HTML::Pipeline::SanitizationFilter::WHITELIST[:elements].sample(rand(1..3)).map do |tag|
+    "#{content_tag(tag, Faker::Hacker.say_something_smart)}<a #{Faker::Hacker.say_something_smart}>"
+  end.join
+end
```
jayroh: Since @glebm fixed this in another PR we probably won't be pulling in this PR in particular but there are some ideas here that may be valuable. Namely edge cases with tags used in unorthodox ways. For those cases, the approach I would take is to add sample markdown or bbcode to [the yml file here] and test against its expected output. 

If I'm reading this right it looks like your default content for a post would always contain between 1 and 3 links? For situations like this I wouldn't leave a default factory up to being random. Instead I would suggest using [factory girl traits] that describe, with a good name, what we'd see from an object when its created. In this case maybe that would call for a trait called `:with_empty_a_tag`?

Just food for thought 😄. Wanted to trade (hopefully) constructive feedback for the work you put in. Thank you for that, @tomczakdominik.

[factory girl traits]: https://robots.thoughtbot.com/remove-duplication-with-factorygirls-traits
[the yml file here]: https://github.com/thredded/thredded/blob/master/spec/lib/thredded/post_contents.yml

PR #321, opened 2016-06-04
Extract content formatting into a separate class

glebm:
Refs #315

The configuration options have been implemented not under the top-level `Thredded` namespace to make it easier to swap out the entire formatter.


glebm:
I'm going to go ahead and merge this (so that I can incorporate these changes into the Moderation PR, which currently uses a hack to render previous post contents even when the post is gone), and will address the comments if any before release in a follow-up PR.

Issue #322, opened 2016-06-05
When user creates a topic, it should be marked as read, not remain unread


timdiggins:
When a user creates a topic, it displays the list of topics in which that topic is seen as unread.

Should be marked as read.
(Doesn't happen for private topics, as after creating a private topic, the private topic itself is displayed which marks it as read).

PR #323, opened 2016-06-05
Ensure that the creator of a topic sees it as 'read' rather than 'unread'

timdiggins:
Addresses #322

### DIFF
--- a/app/controllers/thredded/topics_controller.rb
+++ b/app/controllers/thredded/topics_controller.rb
@@ -74,7 +74,6 @@ module Thredded
       @new_topic = TopicForm.new(new_topic_params)
       authorize_creating @new_topic.topic
       if @new_topic.save
-        UserTopicReadState.read_on_first_post!(thredded_current_user, @new_topic.topic) if signed_in?
 
         redirect_to messageboard_topics_path(messageboard)
       else
--- a/app/controllers/thredded/topics_controller.rb
+++ b/app/controllers/thredded/topics_controller.rb
@@ -74,6 +74,8 @@ module Thredded
       @new_topic = TopicForm.new(new_topic_params)
       authorize_creating @new_topic.topic
       if @new_topic.save
+        UserTopicReadState.create!(user: thredded_current_user, postable: @new_topic.topic, read_at: Time.zone.now, page: 1) if signed_in?
+
         redirect_to messageboard_topics_path(messageboard)
       else
         render :new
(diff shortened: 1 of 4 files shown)


timdiggins:
@glebm @jayroh I submitted this with multiple commits to show my thought process as I'm not 100% sure what you think of doing this in the Form object (it strikes me as best place -- I think it would be good to consider moving other functionality from after_commit callbacks into Form objects or similar).

glebm:
I think the split between model callbacks and form objects/commands is:
1. If something should always happen, use a model callback.
2. If something should only happen in response to a user action, do not use a callback.

So, in this case, having this in the form is better because this should happen only when a user creates a topic, and not necessarily when a topic is created (e.g. through some other means, such as a script that imports topics).

app/forms/thredded/topic_form.rb
```diff
@@ -36,6 +36,7 @@ def save
       ActiveRecord::Base.transaction do
         topic.save!
         post.save!
+        UserTopicReadState.read_on_first_post!(user, topic) if topic.previous_changes.include?(:id)
```
glebm: Let's have the same happen for private topics?

app/controllers/thredded/topics_controller.rb
```diff
@@ -74,6 +74,7 @@ def create
       @new_topic = TopicForm.new(new_topic_params)
       authorize_creating @new_topic.topic
       if @new_topic.save
+
```
glebm: nit: extra empty line

app/forms/thredded/topic_form.rb
glebm: I see this is not strictly necessary for private topics because the topic is displayed after creating it, but I think it'd better to not have the code rely on this.

timdiggins: @glebm I think we ought to have a feature spec that checks that the private topic is read for the creator. I don't think it matters how it happens, but the feature spec should then catch the change. I started creating the test, but it was quite a high bar as there is no existing similar feature spec, and I find the style of these feature specs quite heavyweight and hard-work to create. I could create a quick and light one which is just a capybara spec, if you like?

timdiggins:
@glebm and following on from https://github.com/thredded/thredded/pull/323#issuecomment-223800267 I think the follow, notifications should probably be in forms rather than as callbacks (I didn't write them that way as they were replacing exisitng callbacks) -- I think that could be a future improvement though rather than rejecting that PR altogether...

glebm:
@timdiggins Yeah, can definitely be a future improvement.

app/forms/thredded/topic_form.rb
glebm: @timdiggins Nah, it's fine, doesn't really matter that much, thanks

PR #326, opened 2016-06-10
Translating new keys from the pt-BR locale file

ramirovjr:
I've seen that some keys were added and missing translation.
If anything if wrong or needs correction, please contact me.

Thanks!


glebm:
Thank you! 👍

Issue #328, opened 2016-06-11
Provide a partial for rendering the user's recent posts


glebm:
To provide the parent app with an easy way to render these, e.g. on the user's profile page.
This would require the parent app to include thredded helpers where the partial is rendered.

Something like [Cells](https://github.com/apotonick/cells) would be a cleaner solution, but we're not currently using anything like it.

PR #329, opened 2016-06-11
Provide a partial to render user's recent posts

glebm:
Refs #328 

thredded-user

@jayroh @timdiggins

### DIFF
--- a/README.mkdn
+++ b/README.mkdn
@@ -148,6 +148,22 @@ Include thredded JavaScripts in your `application.js`:
 Thredded views also provide two `content_tag`s available to yield - `:thredded_page_title` and `:thredded_page_id`.
 The views within Thredded pass those up through to your layout if you would like to use them.
 
+### User profile page
+
+Thredded does not provide a user's profile page, but it provides a helper for rendering the user's recent posts
+in your app's user profile page.
+
+To use it:
+1. Include `Thredded::ApplicationHelper` in the app's helpers module.
+2. Render the partial like this:
+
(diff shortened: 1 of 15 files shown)


app/views/thredded/users/_recent_posts.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  # The thredded_current_user method is not available here when rendering from main_app.
+  current_user = send(Thredded.current_user_method) || Thredded::NullUser.new
+  posts = Thredded::PostsPageView.new(
+      current_user,
+      Pundit.policy_scope!(current_user, user.thredded_posts)
+          .where(messageboard_id: Pundit.policy_scope!(current_user, Thredded::Messageboard.all).pluck(:id))
+          .order_newest_first
+          .includes(:postable)
+          .limit(local_assigns[:limit] || 10))
+%>
```
jayroh: Was putting all of the above code in the view as a means to provide the person/team implementing this in their parent app with the code they could use in a controller? If that's the intent I would call that out in the comment above. At the risk of making poor assumptions and stating the obvious - all this work in the view is an anti-pattern 😜

jayroh:
Just one comment in there but otherwise this looks great @glebm :)

app/views/thredded/users/_recent_posts.html.erb
glebm: I've a changed the approach a little bit, now it's a bit less bad. Please take another look!

PR #333, opened 2016-06-14
Private topic searching

lawso017:
Here's a take on searching within Private Topics. The urls_helper conditional seems a bit inelegant, but it works (along with moving the query constrained routes up above the private-topics block).

There's some duplication that could be refactored out between the two TopicsSearch classes but wasn't sure how far to take that. Also didn't see much of a need for User or Category(?) searching in the PrivateTopicSearch.

### DIFF
--- a/app/helpers/thredded/urls_helper.rb
+++ b/app/helpers/thredded/urls_helper.rb
@@ -97,7 +97,9 @@ module Thredded
     # @param messageboard [Thredded::Messageboard, nil]
     # @return [String] the path to the global or messageboard search.
     def search_path(messageboard = nil)
-      if messageboard.try(:persisted?)
+      if controller_name == 'private_topics'
+        private_topics_search_path
+      elsif messageboard.try(:persisted?)
         messageboard_search_path(messageboard)
       else
         messageboards_search_path
--- a/lib/thredded/private_topics_search.rb
+++ b/lib/thredded/private_topics_search.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+require_dependency 'thredded/search_parser'
+module Thredded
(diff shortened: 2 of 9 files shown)


glebm:
We also need to add indexes so that full-text search works on MySQL and is fast on Postgres:

```ruby
DbTextSearch::FullText.add_index connection, :thredded_private_topics, :title,
                       name: :thredded_private_topics_title_fts

DbTextSearch::FullText.add_index connection, :thredded_private_posts, :content,
                       name: :thredded_private_posts_content_fts
```

lib/thredded/private_topics_search.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+require_dependency 'thredded/search_parser'
+module Thredded
+  class PrivateTopicsSearch
```
glebm: This looks almost identical to TopicsSearch, perhaps they can both inherit from TopicsCommonSearch and share most of the implementation?

app/helpers/thredded/urls_helper.rb
```diff
@@ -97,7 +97,9 @@ def edit_preferences_path(messageboard = nil, params = {})
     # @param messageboard [Thredded::Messageboard, nil]
     # @return [String] the path to the global or messageboard search.
     def search_path(messageboard = nil)
```
glebm: I think it would be better to render the search form from every view that needs it explicitly, and have separate partials for topics and private topics searches. We'll also need different text for the private topics search form, something like  "Search Messages", so it makes sense to separate them.

glebm:
Closing, please re-open once the concerns voiced above have been addressed.

PR #335, opened 2016-06-18
Modify readme

chucobo124:
I think we can have more in the readme file to guide dev to have more clear direction.

### DIFF
--- a/README.mkdn
+++ b/README.mkdn
@@ -47,6 +47,14 @@ Add the gem to your Gemfile:
 gem 'thredded', '~> 0.5.1'
 ```
 
+Add `user.rb` file into modle for initialize.
+
+```console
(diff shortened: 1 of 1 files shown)


glebm:
I think we'd be better off providing a skeleton for a standalone guide, and keeping the Thredded Readme focused on Thredded as an engine.

README.mkdn
```diff
@@ -47,6 +47,15 @@ Add the gem to your Gemfile:
 gem 'thredded', '~> 0.5.1'
 ```
 
+Add `user.rb` file into modle for initialize. Please be sure you have current_user method in your user.rb.
```
glebm: Note that the `current_user` method should be defined in the ApplicationController, not in the model

PR #337, opened 2016-06-26
 updated select2 to 4.0

waclock:


glebm:
What about the code in these files?
https://github.com/thredded/thredded/blob/master/app/assets/javascripts/thredded/components/users_select.es6
https://github.com/thredded/thredded/blob/master/app/assets/stylesheets/thredded/components/_select2.scss

BTW, last time I looked at v4 a few months ago it seemed to still lack many of the features of v3 and the documentation was lacking. Not sure if this is still the case.

glebm:
Closing, please re-open once the concerns have been addressed.

Issue #260, opened 2016-04-25
Mark all private topics as read


jayroh:
If there are a lot of old private topics you want to mark as read it would be helpful to have a quick way to mark all of them as read.

TheDudeWithTheThing:
yes plz, maybe I'll grab this one, if I assign myself to this issue then I'm working on it, otherwise consider it still up for grabs

PR #343, opened 2016-07-20
[SMB] Mark privates as read

TheDudeWithTheThing:
#260 

I just threw the link in, not sure where it would best fit

private_messages___thredded_demo

### DIFF
--- a/app/commands/thredded/mark_all_read.rb
+++ b/app/commands/thredded/mark_all_read.rb
@@ -1,27 +1,21 @@
 # frozen_string_literal: true
 module Thredded
   class MarkAllRead
-    def initialize(topic_type, user)
-      @topic_type = topic_type
-      @user = user
-    end
+    def self.run(user)
+      unread_topics = Thredded::PrivateTopic.unread(user)
+      return if unread_topics.empty?
 
-    def run
-      @topic_type.unread(@user).each do |topic|
+      unread_topics.each do |topic|
         last_post = topic.posts.last
         total_pages = topic.posts.page(1).total_pages
 
         UserPrivateTopicReadState.touch!(
-          @user.id,
+          user.id,
           topic.id,
           last_post,
           total_pages
         )
       end
-    end
-
-    def self.run(topic_type, user)
-      new(topic_type, user).run
     end
   end
 end
--- a/app/commands/thredded/mark_all_read.rb
+++ b/app/commands/thredded/mark_all_read.rb
@@ -10,6 +10,7 @@ module Thredded
       @topic_type.unread(@user).each do |topic|
         last_post = topic.posts.last
         total_pages = topic.posts.page(1).total_pages
+
         UserPrivateTopicReadState.touch!(
           @user.id,
           topic.id,
--- a/app/commands/thredded/mark_all_read.rb
+++ b/app/commands/thredded/mark_all_read.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+module Thredded
+  class MarkAllRead
+    def initialize(topic_type, user)
+      @topic_type = topic_type
+      @user = user
+    end
+
+    def run
+      @topic_type.unread(@user).each do |topic|
+        last_post = topic.posts.last
+        total_pages = topic.posts.page(1).total_pages
+        UserPrivateTopicReadState.touch!(
+          @user.id,
+          topic.id,
+          last_post,
+          total_pages
+        )
+      end
+    end
+
+    def self.run(topic_type, user)
+      new(topic_type, user).run
+    end
+  end
+end
--- a/app/controllers/thredded/private_topics_controller.rb
+++ b/app/controllers/thredded/private_topics_controller.rb
@@ -66,12 +66,6 @@ module Thredded
       end
     end
 
-    def mark_all_read
-      MarkAllRead.run(Thredded::PrivateTopic, thredded_current_user) if signed_in?
-
-      redirect_to request.referer
-    end
-
     private
 
     def current_page
--- a/app/controllers/thredded/private_topics_controller.rb
+++ b/app/controllers/thredded/private_topics_controller.rb
@@ -66,6 +66,11 @@ module Thredded
       end
     end
 
+    def mark_all_read
+      MarkAllRead.run(Thredded::PrivateTopic, thredded_current_user) if signed_in?
+      redirect_to request.referer
+    end
+
     private
 
     def current_page
--- a/app/controllers/thredded/private_topics_controller.rb
+++ b/app/controllers/thredded/private_topics_controller.rb
@@ -68,6 +68,7 @@ module Thredded
 
     def mark_all_read
       MarkAllRead.run(Thredded::PrivateTopic, thredded_current_user) if signed_in?
+
       redirect_to request.referer
     end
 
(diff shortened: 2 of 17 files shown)


app/models/concerns/thredded/topic_common.rb
```diff
@@ -10,7 +10,7 @@ module TopicCommon
                  foreign_key: 'last_user_id'
 
       scope :order_recently_updated_first, -> { order(updated_at: :desc, id: :desc) }
-      scope :on_page, -> (page_num) { page(page_num).per(30) }
+      scope :on_page, -> (page_num) { page(page_num) }
```
TheDudeWithTheThing: having `per(30)` here overwrote the `paginates_per 50` on line 6

config/routes.rb
```diff
@@ -11,6 +11,9 @@
       member do
         get '(page-:page)', action: :show, as: '', constraints: page_constraint
       end
+      collection do
+        put :mark_all_read
```
TheDudeWithTheThing: I'm thinking a put here since it's somewhat idempotent, up for debate!

TheDudeWithTheThing:
2 things

1. Should `MarkAllRead.run` lock the user in case there are millions of topics to mark as read and it would fight against race conditions, double clicks, etc....?

2. Same vein, I'm thinking we should grab current time and update with that in case new posts are posted while this is running.

Thoughts @glebm @jayroh

config/locales/pt-BR.yml
```diff
@@ -47,6 +47,7 @@ pt-BR:
       edit_post: Editar Post
       edit_private_topic: :thredded.nav.edit_topic
       edit_topic: Editar
+      mark_all_read: Marcar Como Lido
```
TheDudeWithTheThing: @jayroh translated via google translate, this might say something horrible

obrigado!

app/models/concerns/thredded/topic_common.rb
jayroh: the `paginates_per 50` is tied to a conditional. In most/all cases - does that conditional resolve to true?

1. I'm wondering if that conditional can be removed if such is the case.
2. If it wavers between true/false I'm guessing the `per(30)` is to lock it to a default?

In either case, you're right. It's a little confusing to have two approaches 7 lines apart with anything stating why

app/views/thredded/private_topics/_breadcrumbs.html.erb
```diff
@@ -2,3 +2,8 @@
   <li><%= link_to t('thredded.nav.all_messageboards'), messageboards_path -%></li>
   <li><%= link_to t('thredded.nav.private_topics'), private_topics_path -%></li>
 </ul>
+
+<ul class="thredded--navigation-breadcrumbs">
+  <li><%= link_to t('thredded.nav.mark_all_read'), mark_all_read_private_topics_path, { method: :put } -%></li>
+</ul>
+
```
jayroh: As far as where to put this - what about somewhere that's directly related to, and within, the private topics section where it's not too too far up in peoples' faces. Maybe at the bottom of the page 1 listing?

jayroh: After we merge this, Shaun, I can work to spiff it up a little bit.

config/locales/pt-BR.yml
jayroh: Haaaaaa! Good effort!  🎉 

@ramirovjr what do you think about this one?

config/routes.rb
jayroh: If we're going to be shallow and pedantic I try to not use the collection/member router stuff and stick with resources as much as possible.

```ruby
resources :private_topics, except: [:new, :show], path: '' do
  resources :read_state, only: [:destroy]
end
```

app/controllers/thredded/private_topics_controller.rb
```diff
@@ -66,6 +66,12 @@ def update
       end
     end
 
+    def mark_all_read
+      MarkAllRead.run(Thredded::PrivateTopic, thredded_current_user) if signed_in?
```
jayroh: I don't think we'd need to inject the `Thredded::PrivateTopic` class here because there aren't plans to run this against regular topics (I don't think?). What do you think? Worth keeping around for normal topics? Since we don't keep a long running count shown to users for regular topics it's not as desirable as having it for the private topics.

app/commands/thredded/mark_all_read.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+module Thredded
+  class MarkAllRead
+    def initialize(topic_type, user)
+      @topic_type = topic_type
+      @user = user
+    end
+
+    def run
+      @topic_type.unread(@user).each do |topic|
```
jayroh: What do you think about something like ... 

```ruby
unread_topics = @topic_type.unread(@user)
return if unread_topics.empty?

unread_topics.each do |topic|
  # ... 
```

Using a guard clause could bail out from the rest of everything nice and early. It's a negligible benefit since the enumerator will for all intents and purposes never loop through an empty collection but it still reads pretty well.

config/routes.rb
TheDudeWithTheThing: so just `resource :mark_all_read, only: [:update]`

app/controllers/thredded/private_topics_controller.rb
TheDudeWithTheThing: I wasn't sure if we'd do this on public topics so got all "generic". I'm fine with privates only.

app/models/concerns/thredded/topic_common.rb
TheDudeWithTheThing: I figured `if respond_to?(:paginates_per)` was making sure the gem was loaded or something, either way having both buys us nothing.

The real question is, do we want a page size of 50 or 30?

app/views/thredded/private_topics/_breadcrumbs.html.erb
TheDudeWithTheThing: Yeah I'd prefer to not touch css else it will become a MESS

config/locales/pt-BR.yml
TheDudeWithTheThing: Sepultura #1!

TheDudeWithTheThing:
@jayroh feedback addressed

TheDudeWithTheThing:
@jayroh @glebm anything else here?

jayroh:
@TheDudeWithTheThing it just needs some work on the UX/UI side and then it'll ship. I'll take that part on myself :)

jayroh:
I did a little work on getting this button styled up and a little more while I was in there.



@glebm @TheDudeWithTheThing - I invite you to take a look at your leisure :sunglasses:

TheDudeWithTheThing:
look at that thing!  

👍  👏

glebm:
Looks good!


> Same vein, I'm thinking we should grab current time and update with that in case new posts are posted while this is running.

Sounds good. Might use the ID instead of a primary key, as these are guaranteed to be unique and monotonically increasing.

> Should MarkAllRead.run lock the user in case there are millions of topics to mark as read and it would fight against race conditions, double clicks, etc....?

Since marking something read twice is the same as marking something read once (in other words, it's idempotent), this doesn't seem necessary.

Duleja:
Hi! I was looking for this "mark all as read" option for messageboard/s.
Is there an option to mark certain messageboard or maybe even all messageboards as read from the notification settings maybe?

Sorry if this was answered before

PR #345, opened 2016-07-22
[SMB] Adds unique constraint to MessageboardGroup.name

TheDudeWithTheThing:
Still to come

* Add migration script to append a # to any existing non unique names. 
* Add DB level constraint.


TheDudeWithTheThing:
Not sure what version this would go in so just assumed the next, which is 0.63. This has a migration which will need to be run. It appends a number to any name that is not valid. It goes backwards by created date so that the first one created should not get a number appended.  We then add a unique index to the column.

`20160723012349_upgrade_v062_to_v063.rb`

TheDudeWithTheThing:
@glebm can this go in?

db/upgrade_migrations/20160723012349_upgrade_v062_to_v063.rb
```diff
@@ -0,0 +1,22 @@
...
+  def up
+    i = 1
+    Thredded::MessageboardGroup.order(created_at: :desc).each do |group|
+      Rails.logger.info group.id
+      next if group.valid?
+
+      group.name = "#{group.name}-#{i}"
+      group.save!
+      i += 1
+    end
```
jayroh: Is there a way to do this with plain SQL? General rule of thumb within migrations is to avoid using AR models in the off chance they go away, or their structure changes.

spec/features/thredded/user_creates_messageboard_group_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    user = an_admin
+    user.log_in
+
+    group = set_up_a_messageboard_group
+    group.create_messageboard_group
+    group.visit_new_messageboard_group_form
+    group.submit_form
+
+    expect(group).to be_duplicate
+  end
```
jayroh: What's the scenario where we would *want* a duplicate group? Is there a reason we would want one?

config/locales/pt-BR.yml
```diff
@@ -23,6 +23,7 @@ pt-BR:
       updated_notice: Fórum de mensagem foi atualizado
     messageboard_group:
       create: Criar um novo grupo de mensagens
+      saved: Groupo de mensagens %{name} salvou
```
jayroh: @ramirovjr is this an accurate translation?

ramirovjr: Needs a fix, I'll do it today later.
Sorry for being away for awhile but was overloaded with some work.

jayroh: Man, no apologies necessary. We're lucky you even know this exists so any and all help is greatly appreciated

Thank you, Ramiro :)

spec/features/thredded/user_creates_messageboard_group_spec.rb
TheDudeWithTheThing: No, the test is looking for an error message that it didn't get created.

I'd say you never want a dupe, that's what the PR is preventing

db/upgrade_migrations/20160723012349_upgrade_v062_to_v063.rb
TheDudeWithTheThing: The sql could break too? Not sure if there is a way, we could also just make people fix their own dupes since there may need to be human decisions involved.

app/models/thredded/messageboard_group.rb
```diff
@@ -7,5 +7,6 @@ class MessageboardGroup < ActiveRecord::Base
              dependent: :nullify
 
     validates :name, presence: true
+    validates :name, uniqueness: true
```
glebm: nit: can just add `, uniqueness: true` to the line above.

db/upgrade_migrations/20160723012349_upgrade_v062_to_v063.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+class UpgradeV062ToV063 < ActiveRecord::Migration[5.0]
```
glebm: The `[5.0]` part needs to be removed so that it runs on Rails 4.2

db/upgrade_migrations/20160723012349_upgrade_v062_to_v063.rb
```diff
@@ -0,0 +1,22 @@
...
+  def up
+    i = 1
+    Thredded::MessageboardGroup.order(created_at: :desc).each do |group|
```
glebm: Perhaps we in a way that will work even if a user is running this migration on a previous version of Thredded (i.e. valid? returns true):

```ruby
Thredded::MessageboardGroup.where(
  name: Thredded::MessageboardGroup.group(:name).having('count(id) > 1').select(:name)
).group_by(&:name).each_value do |messageboard_groups|
   messageboard_groups.each_with_index do |messageboard_group, i|
     messageboard_group.update!(name: "#{messageboard_group.name}-#{i + 2}")
  end
end
```

db/upgrade_migrations/20160723012349_upgrade_v062_to_v063.rb
glebm: Generally the migration should only be run with the version of the gem it is migrating to. However, many will run it on the version of the gem it's migrating from, so it'd be nice if it worked there as well.

Using only plain activerecord features I think is fine, because these depend only on the schema. However, using `valid?` is dangerous because we might change the validation in a minor/patch version and then this migration will no longer do the expected thing if the record is invalid due to some other reason.

db/upgrade_migrations/20160723012349_upgrade_v062_to_v063.rb
glebm: Also, should be at least MINOR level update, so this should be UpgradeV06ToV07.

jayroh:
I just rebased this branch, so heads up @TheDudeWithTheThing 

Going to see if I can get this situated and ready for merging

glebm:
@jayroh Let's v0.6.3 before merging this, as this has a database change so the next version will be v0.7.0. Also see my other comments.

jayroh:
@glebm :+1:

On Sun, Aug 14, 2016 at 11:41 AM, Gleb Mazovetskiy <notifications@github.com

TheDudeWithTheThing:
@jayroh want me to take the fixes?

jayroh:
Sure!

TheDudeWithTheThing:
The results are not what I'd expect, are we ok going this route?

TheDudeWithTheThing:
@jayroh @glebm feedback incorporated

app/controllers/thredded/messageboard_groups_controller.rb
```diff
@@ -12,8 +12,10 @@ def create
       authorize @messageboard_group, :create?
 
       if @messageboard_group.save
-        redirect_to root_path
+        redirect_to root_path, notice: I18n.t('thredded.messageboard_group.saved', name: @messageboard_group.name)
       else
+        flash[:notice] = @messageboard_group.errors.full_messages.to_sentence
```
glebm: Flash will only appear after a redirect. We can use `flash.now` instead.

spec/features/thredded/user_creates_messageboard_group_spec.rb
jayroh: Ok I gotcha. If that's the case the exercise should be more descriptive of what the page object is doing wrong in order to test the assertion that there's an error.

1. Change `group.submit_form` to something like `group.submit_form_with_duplicate_group_ information`
2. Change the assertion to communicate more as far as what you expect. Something like `expect(group).to have_duplicate_messageboard_group_error`

jayroh:
> The results are not what I'd expect, are we ok going this route?
> 
> 

What's wrong here?

TheDudeWithTheThing:
Instead of leaving 1 unchanged and appending -# to the rest it just changes all of them, probably not a big deal as someone would have to go through and do some manual cleanup anyway

db/upgrade_migrations/20160723012349_upgrade_v0_6_to_v0_7.rb
```diff
@@ -0,0 +1,23 @@
...
+  def up
+    Thredded::MessageboardGroup.transaction do
+      Thredded::MessageboardGroup.where(
+        name: Thredded::MessageboardGroup.group(:name).having('count(id) > 1').select(:name)
+      ).group_by(&:name).each_value do |messageboard_groups|
+        messageboard_groups.each_with_index do |messageboard_group, i|
+          messageboard_group.update!(name: "#{messageboard_group.name}-#{i + 2}")
+        end
+      end
+    end
```
jayroh: If the suffix to the group name doesn't end up looking like

* group-name
* group-name-1
* group-name-2

Then could we assign a random int suffix to the duplicate name columns with pure sql?

random int between 1 and 10

```
postgres=# SELECT floor(random()*(10-1+1))+1;
 ?column?
----------
     8
(1 row)
```

append to column value

```
update the_table
  set the_column = concat(the_column, 'this will be appended')
where
```

had to look these up but this should be possible with pure sql and might be a little easier to follow then the pure ruby code above.

1. http://www.techonthenet.com/postgresql/functions/random.php
2 http://stackoverflow.com/questions/30162486/append-text-to-column-data-in-postgresql-if-columns-initial-value-is-null

TheDudeWithTheThing: random has a chance to collide though

glebm: If you want the suffix to look like this, simply replace this:

```ruby
messageboard_groups.each_with_index do |messageboard_group, i|
```

with

```ruby
messageboard_groups.from(1).each_with_index do |messageboard_group, i|
```

and `i + 2` -> `i + 1`

in the migration

jayroh:
lgtm

glebm:
LGTM, just need resolving conflicts and let's merge

TheDudeWithTheThing:
@glebm conflicts resolved!

PR #354, opened 2016-08-04
Scopes for finding topics followed by / unread by current user

timdiggins:
I need this for my main_app and for the 'plugin' I'm developing:

`Topic.unread_followed_by(current_user).count`

but also provides:
`Topic.followed_by(current_user)`

### DIFF
--- a/app/models/thredded/topic.rb
+++ b/app/models/thredded/topic.rb
@@ -16,21 +16,10 @@ module Thredded
     scope :order_sticky_first, -> { order(sticky: :desc) }
 
     scope :followed_by, lambda { |user|
-      joins('INNER JOIN thredded_user_topic_follows ON thredded_user_topic_follows.topic_id = thredded_topics.id')
-        .where('thredded_user_topic_follows.user_id = ?', user.id)
+      joins(:user_follows)
+        .where(thredded_user_topic_follows: { user_id: user.id })
     }
-    scope :unread_by, lambda{ |user|
-      joins(<<-SQL)
-        LEFT OUTER JOIN thredded_user_topic_read_states
-          ON thredded_user_topic_read_states.postable_id = thredded_topics.id
-             AND thredded_user_topic_read_states.user_id = #{user.id}
-      SQL
-        .where(<<-SQL)
-          thredded_user_topic_read_states.read_at is NULL OR
-          thredded_user_topic_read_states.read_at < thredded_topics.updated_at
-        SQL
-    }
-    scope :unread_followed_by, ->(user) { followed_by(user).unread_by(user) }
+    scope :unread_followed_by, ->(user) { followed_by(user).unread(user) }
 
     extend FriendlyId
     friendly_id :slug_candidates,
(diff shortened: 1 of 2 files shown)


app/models/thredded/topic.rb
```diff
@@ -15,6 +15,23 @@ class Topic < ActiveRecord::Base
 
     scope :order_sticky_first, -> { order(sticky: :desc) }
 
+    scope :followed_by, lambda { |user|
+      joins('INNER JOIN thredded_user_topic_follows ON thredded_user_topic_follows.topic_id = thredded_topics.id')
```
glebm: This is just `joins(:user_follows)`

app/models/thredded/topic.rb
```diff
@@ -15,6 +15,23 @@ class Topic < ActiveRecord::Base
...
+        .where('thredded_user_topic_follows.user_id = ?', user.id)
```
glebm: There are several ways to do this without using sql strings:

1. `where(thredded_user_topic_follows: {user_id: user.id})`
2. `merge(Thredded::UserTopicFollow.where(user_id: user.id))`
3. `merge(user.thredded_topic_follows)`

app/models/thredded/topic.rb
```diff
@@ -15,6 +15,23 @@ class Topic < ActiveRecord::Base
...
+    }
+    scope :unread_by, lambda{ |user|
+      joins(<<-SQL)
```
glebm: It's best to do this with AREL instead of custom SQL fragments, because arel queries are easier to refactor because they are composable, there is less scope for errors, and the generated SQL will be more compact for the specific database. See here for an example of how `LEFT JOIN` can be done with Arel: https://github.com/thredded/thredded/blob/8772b099863c441f5716285a44fb8648517eb083/app/models/concerns/thredded/topic_common.rb#L44

app/models/thredded/topic.rb
```diff
@@ -15,6 +15,23 @@ class Topic < ActiveRecord::Base
 
     scope :order_sticky_first, -> { order(sticky: :desc) }
 
+    scope :followed_by, lambda { |user|
+      joins('INNER JOIN thredded_user_topic_follows ON thredded_user_topic_follows.topic_id = thredded_topics.id')
+        .where('thredded_user_topic_follows.user_id = ?', user.id)
+    }
+    scope :unread_by, lambda{ |user|
```
glebm: This scope already exists, it's called `unread` and is defined in https://github.com/thredded/thredded/blob/8772b099863c441f5716285a44fb8648517eb083/app/models/concerns/thredded/topic_common.rb#L40

timdiggins: @glebm ah cool -- so in other words this PR was unnecessary (but useful for the notes -- thanks! I've never really fully got arel, but I guess I need to)

timdiggins: no I'm being stupid, only unread_by is a duplicate. Have fixed this in the attached PR. If it's good I'll rebase and squash

glebm:
:+1:

app/models/thredded/topic.rb
glebm: Hadn't seen this comment before I went ahead and merged it using the Github merge&squash thing.

PR #367, opened 2016-08-09
allow the class TopicsView uses to wrap topics to be customized

timdiggins:


glebm:
Looks like we should switch to Cells sooner rather than later :smile:

PR #368, opened 2016-08-09
Extract seed database as partially tested library

timdiggins:
Can be used in console to generate specific kinds of data when testing


lib/thredded/database_seeder.rb
```diff
@@ -0,0 +1,269 @@
...
+
+      def create(*_args)
+        fail 'override in your class!'
+      end
+
+      def find
+        model_class.all.to_a
+      end
+
+      def find_or_create(*args)
```
glebm: nit: Inconsistent return type. This method return a single instance from the `if` branch, but an `Array` from the `else` branch (for the ones that accept `count` in create).

glebm: Looks like this is the only public method. Perhaps moving it right after `initialize` and marking the rest of the methods `protected` would help readability here.

lib/thredded/database_seeder.rb
```diff
@@ -0,0 +1,269 @@
...
+        self.class::MODEL_CLASS
+      end
+
+      delegate :exists?, to: :model_class
+
+      def create(*_args)
+        fail 'override in your class!'
+      end
+
+      def find
```
glebm: I think make the find and create methods abstract and definining them in two base classes would help clarify intent here:

```ruby
# @abstract
class FirstSeedData < SeedData
end

# @abstract
class CollectionSeedData < SeedData
end
```

lib/thredded/database_seeder.rb
glebm: Also, looks like `exists?` is not necessary, as `find` can simply return `nil`.

```ruby
# @abstract
class SeedData

  def initialize(seeder) ...

  def find_or_create ...

  protected

  # @abstract
  def find
    fail 'Unimplemented'
  end
  
  def create
    fail 'Unimplemented'
  end
  ...
end
```

lib/thredded/database_seeder.rb
```diff
@@ -0,0 +1,269 @@
...
+        model_class.all.to_a
+      end
+
+      def find_or_create(*args)
+        return @stored if @stored
+        @stored = if exists?
```
glebm: So if `exists?` is removed this method would simply be `@stored ||= find.presence || create(*args)`.

For this to work, the collection seed data `find` return value should be changed from `.all.to_a` to simply `.all`.

lib/thredded/database_seeder.rb
```diff
@@ -0,0 +1,269 @@
...
+
+    class Posts < BaseSeedData
+      MODEL_CLASS = Post
+
+      def create(count: (1..1))
+        log "Creating #{count} additional posts in each topic..."
+        seeder.topics.flat_map do |topic|
+          posts = []
+          written = (1 + rand(5)).days.ago
+          (count.min + rand(count.max + 1)).times do
```
glebm: nit: Can be simplified to `.times.map` to avoid manually accumulating posts in the `posts` variable.

lib/thredded/database_seeder.rb
timdiggins: great, but don't need to use `presence` at all if find returns `nil`. Not 100% sure if I need to do the to_a to ensure array is cached rather than re-returned, but it works.

lib/thredded/database_seeder.rb
timdiggins: ok, rubocop then demanded I use `Array.new` with a block rather than `times.map` (seems less clear to me but ...)

Issue #352, opened 2016-08-03
Sporadic spec failure in travis ci


EllaNancyFay:
**Failure**
```
1) Thredded::TopicView#states prepends a read state
     Failure/Error: expect(topic_view.states[0]).to eq :read
     
       expected: :read
            got: :unread
```

**Builds**
https://travis-ci.org/thredded/thredded/jobs/149263730

Also happened here, but was fixed by rebuild

https://travis-ci.org/red56/thredded/builds/149438017

jayroh:
I've seen the test suite fail locally as well. I'm confident that there is
some state leaking across tests depending on the order they're run at that
time. I don't have it in front of me now but I'm pretty sure we're running
the suite in random order, so ... to fix this would be someone getting a
failure locally, re-running it against the rspec seed #, and determining
where the state leakage is

timdiggins:
@jayroh @EllaNancyFay this one doesn't look like an ordering problem because if it was being run in a random order we'd be able to see (and then reproduce the error locally):

```
Randomized with seed 63715
```

however that's not output in the travis build https://travis-ci.org/thredded/thredded/jobs/149263730

jayroh:
Is the order of the tests running held all in that one seed number?

I thought it saved the test order to disk and referred to it via the seed number ¯\\\_(ツ)\_/¯

glebm:
@jayroh The seed number is the random number generator seed. The rest of the program is deterministic, so in a sense yes the order of the test is held in that one number (for a fixed codebase).

I've made them run at random as they weren't previously, but yeah this issue is something else (don't know what yet).

glebm:
Got the database log of a failure here, if anyone wants to take a closer look!
https://travis-ci.org/thredded/thredded/jobs/151749083

timdiggins:
@glebm @jayroh @EllaNancyFay 
I think that by creating and using `Topic#last_post_at` we've fixed this spec (and real) bug. 

If you want the full reasoning, read on:

if you image a test like the following:

```ruby
      topic = create(:topic, with_posts: 3)
      last_post = topic.posts.order_oldest_first.last

      expect(topic.last_post_at - last_post.created_at).to eq(0) # fine
      expect(topic.updated_at - last_post.created_at).to eq(0) # fails, see below
```

The failure is:

```
     Failure/Error: expect(topic.updated_at - last_post.created_at).to eq(0)
     
       expected: 0
            got: 0.11082
```

[...]

PR #369, opened 2016-08-09
Order posts by created_at rather than id

### LINKED ISSUE #360: Posts and PrivatePosts should be ordered by created_at not by id
Currently posts & private posts when `order_oldest_first` / `order_newest_first` are actually (under the hood) ordered by id.

This seems appealing in some ways, but it is a bit of a nightmare. In particular if the creation order is screwed up (e.g. by importing) then you have problems. It also just feels wrong to me -- and it was deeply surprising when I found out.

I have started changing this behaviour to be ordered by `created_at`

Of course changing this is quite a bit of work, as much test data currently relies on this, and some untested behaviours (e.g. `Post` / `PrivatePost` `#page()`) 

However, I've also discovered that there are one or two places where some desired behaviours rely on a natural id-ordering (rather than explicit call to `order_oldest_first`) which is not dependable (at least in postgres).

- [x] change ordering
- [x] deal with spec failures (including the read-state spec mentioned in #352)
- [x] add in default `order_oldest_first` for the associations in `Topic` and `PrivateTopic`
- [x] write spec for and change `#page` for both `Post` and `PrivatePost`
- [x] dry up `#page` by extracting to either PostCommon or PageCalculable

timdiggins:
closes #360

### DIFF
--- a/app/models/concerns/thredded/post_common.rb
+++ b/app/models/concerns/thredded/post_common.rb
@@ -13,8 +13,8 @@ module Thredded
 
       validates :content, presence: true
 
-      scope :order_oldest_first, -> { order(created_at: :asc) }
-      scope :order_newest_first, -> { order(created_at: :desc) }
+      scope :order_oldest_first, -> { order(created_at: :asc, id: :asc) }
+      scope :order_newest_first, -> { order(created_at: :desc, id: :desc) }
     end
 
     def avatar_url
--- a/app/models/concerns/thredded/post_common.rb
+++ b/app/models/concerns/thredded/post_common.rb
@@ -13,8 +13,8 @@ module Thredded
 
       validates :content, presence: true
 
-      scope :order_oldest_first, -> { order(id: :asc) }
-      scope :order_newest_first, -> { order(id: :desc) }
+      scope :order_oldest_first, -> { order(created_at: :asc) }
+      scope :order_newest_first, -> { order(created_at: :desc) }
     end
 
     def avatar_url
--- a/app/models/concerns/thredded/post_common.rb
+++ b/app/models/concerns/thredded/post_common.rb
@@ -21,6 +21,10 @@ module Thredded
       Thredded.avatar_url.call(user)
     end
 
+    def calculate_page(postable_posts, per_page)
+      1 + postable_posts.where(postable_posts.arel_table[:created_at].lt(created_at)).count / per_page
+    end
+
     # @param view_context [Object] the context of the rendering view.
     # @return [String] formatted and sanitized html-safe post content.
     def filtered_content(view_context, users_provider: -> (names) { readers_from_user_names(names) })
--- a/app/models/thredded/topic.rb
+++ b/app/models/thredded/topic.rb
@@ -44,7 +44,7 @@ module Thredded
                inverse_of:    :topics,
                counter_cache: :topics_count
 
-    has_many :posts, -> { order_oldest_first },
+    has_many :posts,
              class_name:  'Thredded::Post',
              foreign_key: :postable_id,
              inverse_of:  :postable,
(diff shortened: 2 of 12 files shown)


app/models/concerns/thredded/post_common.rb
```diff
@@ -13,14 +13,18 @@ module PostCommon
 
       validates :content, presence: true
 
-      scope :order_oldest_first, -> { order(id: :asc) }
-      scope :order_newest_first, -> { order(id: :desc) }
+      scope :order_oldest_first, -> { order(created_at: :asc) }
```
glebm: This should also sort by ID in case of conflicts to establish a stable ordering. `order(created_at: :asc, id: :asc)`

app/models/thredded/topic.rb
```diff
@@ -44,7 +44,7 @@ class Topic < ActiveRecord::Base
                inverse_of:    :topics,
                counter_cache: :topics_count
 
-    has_many :posts,
+    has_many :posts, -> { order_oldest_first },
```
glebm: Is this actually used anywhere? I thought all the current code establishes the order explicitly. Default scopes lead to some very unexpected behaviour. E.g. topic.posts.order_newest_first will no longer work (because it uses `order` which adds order conditions to the end).

> but it seemed a bit repititive to add the order to every single place we are using .posts...
There isn't a whole lot of these. We also use different ones (e.g. on the user's recent posts it's newest first, on the other views it's oldest first, elsewhere in the codebase outside of the view it's either one or something else entirely).

app/models/concerns/thredded/post_common.rb
timdiggins: Ugh, really? -- millisecond accuracy in postgres/sqlite. and (ok) second accuracy in mysql. How many times are these *actually* going to clash? And would it be surprising to anyone if the ordering was indeterminate at this point (and how many times are they going to clash at a page boundary. I'm going to go out on a limb and say 'never').

app/models/thredded/topic.rb
timdiggins: @glebm Ugh, you're right -- `order_newest_first` becomes very complicated. if we redefine it as reorder() then it starts a pattern of unexpected behaviour.

Ok, I guess so. I notice about 5 places where I'd like to see an order and it's missing. Only one of them is critical (two are in specs, and the other two are only used when there is only one post, so I've just added a comment)

https://github.com/thredded/thredded/blob/master/app/commands/thredded/mark_all_read.rb#L9. That this is in very recently added code does suggest to me that order_oldest_first is a kind of natural order that ought to be default, but I agree the problems around defaults in scopes and associations are too big. Just have to watch out for this in PRs.

Have also spotted a marginal error in a spec/features/page_object. Really don't like these. Just saying. They make the feature code harder to reason about and harder to write.

app/models/concerns/thredded/post_common.rb
glebm: There will be clashes in integration tests, leading to hard to debug flakes. Actually, thinking about it, this might be the reason for #352!

And I've seen clashes like this in production on MySQL, it's not that unlikely because [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem).

timdiggins: I'm not sure this is really covered by the birthday problem, but I'm not going to argue that one! Don't think #352 is really the same, but in general think the answer for test problems is to improve the test data! 

however if you've seen this in production, and for the sake of a more complex index and a few extra characters of code... I give in! 😃

glebm: If 10 posts are made within 5 minutes, the probability of there being at least two posts with exactly the same timestamp is 14%! Calculator: http://algoviz.org/OpenDSA/dev/OpenDSA/AV/Hashing/Birthday.html

PR #375, opened 2016-08-11
current_user reference updated

chrislewis60:
The current_user reference didn't work in our app because the helper was current_member (we're using Thredded with RefineryCMS). Having looked at what the current_user reference was doing, I think it's ok to change that to be @user, which is already defined in the same method and is based on the Thredded.user_class config option. I've checked that it still does what I expect in our app and the tests all pass so let me know if that's an appropriate change to make?

### DIFF
--- a/app/controllers/thredded/moderation_controller.rb
+++ b/app/controllers/thredded/moderation_controller.rb
@@ -68,7 +68,7 @@ module Thredded
         .order_newest_first
         .includes(:postable)
         .page(current_page)
-      @posts = PostsPageView.new(current_user, posts_scope)
+      @posts = PostsPageView.new(@user, posts_scope)
     end
 
     def moderate_user


glebm:
You can simply set `Thredded.current_user_method = :current_member` in the [initializer](https://github.com/thredded/thredded/blob/master/lib/generators/thredded/install/templates/initializer.rb#L23).

chrislewis60:
I have set it to be :current_member in the initializer. The problem is that 'current_user' is hardcoded in the controller without referencing the initializer.

glebm:
@chrislewis60 Oh, right. Fixed. It should've been `thredded_current_user`. `@user` refers to a different user there.

chrislewis60:
Great - `thredded_current_user` works for me. I hadn't spotted that the PostsPageView needs the user that's viewing the posts, not the user who's posts you're viewing. Thanks.

PR #383, opened 2016-08-23
Have a topic's last user default to it's creator

zimkies:


glebm:
@zimkies last_user [is assigned](https://github.com/thredded/thredded/blob/8b14b9ba2f334b028f4dcbc3949fdfa41d496ed7/app/models/thredded/post.rb#L64) whenever a post is created or destroyed, so this should not be necessary.

zimkies:
@glebm whoops sorry I was just trying to customize thredded for our own application, didn't mean to PR here. 

However - this still is an issue you might want to address. When a topic is created, sometimes it doesn't have a post created with it. In that state, the UI displays that 'Deleted User' next to the topic on the messageboard page, which is not what we want.

glebm:
@zimkies Thredded doesn't allow topics without posts in them. There are other parts of Thredded that also rely on there always being a post in every topic.

zimkies:
Got it! This was a result of the customization then. Sorry to bother you!

glebm:
@zimkies Looking at the code, I did find a scenario (a bug) where last_user_id can be nil in stock Thredded: when a new topic is created by a user whose moderation state is not "approved". Will fix.

zimkies:
Cool - yes, I think that was actually the problem I ran into

-Ben

PR #390, opened 2016-08-26
Use thredded_current_user instead of current_user

Nerian:
It should be using thredded_current_user instead of current_user. Otherwise this fails if current_user is not the method configured in the initializer.


jayroh:
Looks like this exposes another bug altogether. According to the travis run(s).

```
Failure/Error: <% if follow = thredded_current_user.following?(topic)%>
     
     ActionView::Template::Error:
       undefined method `following?' for #<Thredded::NullUser:0x000000079635c8>
```

The `Thredded::NullUser` class needs a `following?` instance method added (probably returning false) in order for this to pass.

Nerian:
Moved to https://github.com/thredded/thredded/pull/391

PR #392, opened 2016-08-31
Optionally show all following users (on topic page)

### LINKED ISSUE #348: List of following users in topic page
We'd like to add a list of users that are following the topic underneath its header in Topic#show. While this not be to everyone's requirement, and in some use cases would be deeply problematic, would like to add a configuration option (by default off).

```Thredded.show_followers_in_topic: true```

Any thoughts?

### PR DESCRIPTION
**Pull request by timdiggins**
There's an initialization configuration option (default false):

```
# self.show_following_users_in_topic = false
```

When this is true, it adds the list of following people to the topic page: 

it_s_not_impossible__i_used_to_bullseye_womp_rats_in_my_t-16_back_home__they_re_not_much_bigger_than_two_meters____thredded_demo

This is not appropriate for every Thredded installation, but it is appropriate for some (particularly with a smaller set of followers).

resolves #348

### DIFF
--- a/app/assets/stylesheets/thredded/components/_topic-header.scss
+++ b/app/assets/stylesheets/thredded/components/_topic-header.scss
@@ -55,3 +55,8 @@
     }
   }
 }
+
+&--topic-following-users{
+  font-size: $thredded-font-size-small;
+  color: $thredded-secondary-text-color;
+}
--- a/app/assets/stylesheets/thredded/components/_topic-header.scss
+++ b/app/assets/stylesheets/thredded/components/_topic-header.scss
@@ -56,7 +56,7 @@
   }
 }
 
-&--topic-following-users{
+&--topic-followers{
   font-size: $thredded-font-size-small;
   color: $thredded-secondary-text-color;
 }
--- a/app/helpers/thredded/application_helper.rb
+++ b/app/helpers/thredded/application_helper.rb
@@ -17,6 +17,16 @@ module Thredded
       render partial: 'thredded/users/link', locals: { user: user }
     end
 
+    # @param user [Thredded.user_class]
+    # @return [String] wrapped @mention string
+    def user_mention(user)
+      if user.to_s.include?(' ')
+        %(@"#{user}")
+      else
+        "@#{user}"
+      end
+    end
+
     # @param datetime [DateTime]
     # @param default [String] a string to return if time is nil.
     # @return [String] html_safe datetime presentation
--- a/app/helpers/thredded/application_helper.rb
+++ b/app/helpers/thredded/application_helper.rb
@@ -20,10 +20,11 @@ module Thredded
     # @param user [Thredded.user_class]
     # @return [String] wrapped @mention string
     def user_mention(user)
-      if user.to_s.include?(' ')
-        %(@"#{user}")
+      username = user.send(Thredded.user_name_column)
+      if username.include?(' ')
+        %(@"#{username}")
       else
-        "@#{user}"
+        "@#{username}"
       end
     end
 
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: app/assets/stylesheets/thredded/components/_topic-header.scss [hunk 1]
```diff
@@ -55,3 +55,8 @@
     }
   }
 }
+
+&--topic-following-users{
```
glebm: topic-followers?

app/view_models/thredded/topic_view.rb
```diff
@@ -2,7 +2,7 @@
 module Thredded
   # A view model for Topic.
   class TopicView < BaseTopicView
-    delegate :categories, :id, :blocked?, :last_moderation_record,
+    delegate :categories, :id, :blocked?, :last_moderation_record, :following_users,
```
glebm: followers?

app/views/thredded/topics/_following_users.html.erb
```diff
@@ -0,0 +1,12 @@
+<% if Thredded.show_following_users_in_topic %>
+  <div class="thredded--topic-following-users">
+    <% if topic.following_users.present? %>
+      Following:
```
glebm: "Followed by:"?
Please i18n

app/views/thredded/topics/_following_users.html.erb
```diff
@@ -0,0 +1,12 @@
...
+      <% topic.following_users.each do |user| %>
+        <%= "@#{user}" %>
```
glebm: perhaps these should be user links?

app/views/thredded/topics/_header.html.erb
```diff
@@ -24,4 +24,5 @@
       </div>
     <% end %>
   <% end %>
+  <%= render partial: 'thredded/topics/following_users', locals: {topic: topic} %>
```
glebm: followers?

lib/generators/thredded/install/templates/initializer.rb
```diff
@@ -37,6 +37,9 @@
 # Whether posts and topics pending moderation are visible to regular users.
 Thredded.content_visible_while_pending_moderation = true
 
+# Whether users that are following a topic are listed on topic page.
+Thredded.show_following_users_in_topic = false
```
glebm: show_topic_followers?

app/views/thredded/topics/_following_users.html.erb
timdiggins: yes. much better @glebm

app/views/thredded/topics/_following_users.html.erb
timdiggins: @glebm My thinking is that it helps new users to understand how the mentioning system works. And yes, gotcha on the quotes.

timdiggins:
@glebm that's all of your suggestions done, apart from the user_link -- have made sure they quote correctly (I couldn't find an existing helper so made one).

app/helpers/thredded/application_helper.rb
```diff
@@ -17,6 +17,16 @@ def user_link(user)
       render partial: 'thredded/users/link', locals: { user: user }
     end
 
+    # @param user [Thredded.user_class]
+    # @return [String] wrapped @mention string
+    def user_mention(user)
+      if user.to_s.include?(' ')
```
glebm: Instead of `user.to_s`, this should actually be `user.send(Thredded.user_name_column)`. There is no distinction in the demo, but on a site like GitHub `to_s` would return the full name, but mentions would happen with @-username.

glebm:
LGTM :+1:

PR #393, opened 2016-09-03
rename dummy's current_user to make specs trap misuse of it.

timdiggins:
It would be nice to prevent mistaken uses of current_user, when
we should be using thredded_current_user (see #391) - so maybe
we can have the spec smoke it out by having the dummy use a differently
named (and thus configured) current_user.


jayroh:
This is a great idea. Simple and effective.

glebm:
Awesome! :star2:

Issue #317, opened 2016-06-03
UI tweak to include follow status and modified unread status


timdiggins:
Following on from https://github.com/thredded/thredded/pull/310, @EllaNancyFay and I have tweaked the styles in our main_app as follows -- this could be kept in our main_app, or pulled into thredded as a default if you like it:

**How the topics#index looks, with unread/read and followed/unfollowed:**
In this the bells indicate following, and the bold represents unread. It seems fairly intuitive when you use it (could make a PR so you can try out). I think it could do with adding the total number of posts as just a textual annotation (would want to change thredded DOM for this)
![local](https://cloud.githubusercontent.com/assets/18395/15775097/f0296084-2977-11e6-839b-2dc278b0acda.png)

**The header for a topic which you are following:**
![local](https://cloud.githubusercontent.com/assets/18395/15775060/c0b8df00-2977-11e6-89c8-d38cbab9c600.png)

**The header for a topic which you are not following:**
![local](https://cloud.githubusercontent.com/assets/18395/15775068/cf58d042-2977-11e6-8d53-70eb705b5579.png)

What do you think @glebm @jayroh for inclusion in Thredded (once the main PR is merged)

glebm:
Nice!

jayroh:
I LOVE IT

I have only a couple of comments!

The treatment on topics#index - what do you think about applying an opacity of, say, 70-80% to the bell by default and 100% on :hover/:focus? On a page where there are many follows it might get a little noisy. If possible, take a look and see how it looks with ... 40% of the topics on any given topics#index marked as "following" to see how it looks? Just as an exercise in max edge-cases.

What would it look like if the bell was abutted to the right of the topics header(s)? Both on index and show? I do like seeing the # of posts in a topic on #index and feel like the bell being set to the right might keep the vertical visual weight a little more balanced. With a bell to the left on both cases you keep some additional proximity for the messaging. For example:

![](https://www.evernote.com/l/AUqDZeh2blxPbYeiFMgmuesU0HntTYS4ieAB/image.png)

These two things are closely (if not wholly) related but the proximity doesn't represent this fact.

timdiggins:
Although we've done the change to add Follow status icon in #394 in the lists and headers, there is still the question (in this issue) about whether to change the way that total counts of posts in a topic are represented, but separating out into a different issue: #402

PR #394, opened 2016-09-04
Ui Tweak for follow status

timdiggins:
UI tweak to include follow status and modified unread status following #317.

topic#index:

main_board___thredded_demo

topic#show (header)

you_do_have_your_moments__not_many__but_you_have_them____thredded_demo

Tried a rollover with increased opacity -- but it didn't seem to add much.


timdiggins:
Would be useful to get feedback on proposed tweak to looks @glebm @jayroh but am planning to change dom (so can have something to attach a follow/unfollow js in main_app/plugin)

jayroh:
I think it looks great man. I don't see anything off-hand that should keep you from squashing the commits and merging when you feel comfortable.

app/assets/stylesheets/thredded/components/_topic-header.scss
```diff
@@ -60,3 +60,67 @@
   font-size: $thredded-font-size-small;
   color: $thredded-secondary-text-color;
 }
+
+
+@mixin following {
```
glebm: Sass mixins all live in one global namespace, so ours should be prefixed with `thredded-`.
However, looks like these mixins do not accept any arguments, so they probably should just be placeholder rules.

How about `%thredded-following-icon` and `%thredded-not-following-icon`?

app/assets/stylesheets/thredded/components/_topic-header.scss
```diff
@@ -60,3 +60,67 @@
...
+  &::after {
+    display: inline-block;
+    font-size: 1em;
+    width: 1.4rem;
+    height: 1.4rem;
+    position: absolute;
+    top: 0;
+    right: -1.6rem;
+    background-image: image-url("thredded/follow.svg");
```
glebm: With non-inline SVGs there is no way to change the color from CSS. Does this look OK on all themes?

app/assets/images/thredded/follow.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
glebm: These SVGs should probably be optimized with [svgo](https://github.com/svg/svgo)

app/assets/stylesheets/thredded/components/_topic-header.scss
glebm: You could render a single inline SVG with the icon, and then an svg with `<use>` that references the icon, and style the colors with CSS. http://tympanus.net/codrops/2015/07/16/styling-svg-use-content-css/

timdiggins: @glebm Yes, this won't look good with other themes with other bg colors.

Didn't really understand inline svg before -- this is amazingly useful.

Just to summarize: I can define an svg inline in the dom (but not visible in this position -- will research this more carefully), and then add the `<use>` within the dom each place we want it to go, and then style that as needed for different themes. This will work well for my js-attachment needs too.

glebm: Yes, that's exactly it. You can define an svg inline (you can use the `inline_svg` helper for this, https://github.com/jamesmartin/inline_svg), make it invisible with `display: none`, and `<use>` instances of it in the page with whatever CSS styles needed.

timdiggins:
Ok, we have incorporated all the feedback on svg, AFAICT. Only tweak is making a very faint not-following icon.

main_board___thredded_demo

will merge later today unless I get negative feedback

Issue #398, opened 2016-09-14
Use User#thredded_display_name instead of User#to_s


glebm:
From https://github.com/thredded/thredded/issues/397#issuecomment-246917371

timdiggins:
I'm on this.

timdiggins:
Configurable `thredded_display_name` for User, with a default implementation (proc) of `send(Thredded.user_name_column) || ''`

jayroh:
Would it be prudent to raise a custom Thredded exception if the method is not defined instead of returning an empty string? I have a feeling it could be a bit mysterious if someone sees `""` show up instead of a user's name, email, nickname, etc., and not know why or where this should be set and doesn't think to look through the config or the code.

timdiggins:
@jayroh yes good shout -- but doesn't need to be a custom exception, could just be a custom message like `raise "User.#{Thredded.user_name_column} must not be nil: please set make sure non nil or configure Thredded.user_display_name_method"`

glebm:
@timdiggins Sounds good, so something like this?

```ruby
def thredded_display_name
  send(Thredded.user_display_name_method).presence or fail "..."
end
```

```ruby
module Thredded
  mattr_accessor :user_display_name_method
  
  def self.user_display_name_method
    super || user_name_column
  end

  ...
```

timdiggins:
@glebm @jayroh I started coding this already, thinking it was going to be hard work because we'd have to find all instances of to_s (including implicit uses), so best case was to change `User#to_s` in the dummy app to raise an exception to drive the change (and prevent future ones).

But then while doing it I realized that (AFAICT, still testing it) we can just provide a default `to_s` in `UserExtender` and this will be used if no `to_s` is defined on `User`, but if one is defined, it will be used.

If that is 100% the case (see below ) then we have two options:

1) just implement a default `UserExtender#to_s` and document the fact that thredded uses this behaviour -- this means we can continue to use the idiomatic and very rails-conventional uses of (implicit) `User#to_s`  in the codebase

2) go ahead and replace uses of "User#to_s" with "thredded_display_name" (using the raise trick above)

There is one case where I think adding `UserExtender#to_s` would actually override main_app implementation: where User includes a *different* concern or extends a parent base/abstract class and either one includes a to_s they want to use. Seems possible, if rare.

[...]

PR #401, opened 2016-09-15
Provide default thredded user display

timdiggins:
Address #398 by providing configurable `Thredded.display_name_method` and default `thredded_display_name` implementation

Need to:

* [x] add more tests for configuration (not 100% sure it's right yet)
* [x] add commented out stanza to inititalizer generator template


timdiggins:
Ok ready to merge. Would be good to get a pair of eyes on this @jayroh @glebm

spec/dummy/app/views/layouts/application.html.erb
```diff
@@ -30,7 +30,7 @@
       <header class="app-container-header">
         <span class="app-nav-auth">
           <% if signed_in? %>
-            You are <%= link_to the_current_user, the_current_user %> |
+            You are <%= link_to the_current_user.name, the_current_user %> |
```
glebm: `.thredded_display_name`?

app/models/thredded/user_extender.rb
```diff
@@ -65,5 +65,11 @@ def thredded_anonymous?
     def following?(topic)
       UserTopicFollow.find_by(topic_id: topic.id, user: self)
     end
+
+    def thredded_display_name
+      send(Thredded.user_display_name_method).presence || fail(<<-ERROR)
```
glebm: nit: Usually, `or` is used to indicate control flow.

spec/dummy/app/views/layouts/application.html.erb
timdiggins: @glebm I don't think so, this is within the dummy app, so shouldn't really use thredded methods (but can't use to_s because it will blow up...), if you imagine this was a main_app.

app/models/thredded/user_extender.rb
timdiggins: funnily enough I wrote `or`, then looked at it and thought maybe it was a perlism .

timdiggins: @glebm :-( I think rubocop thinks it's a perlism too:

```
 app/models/thredded/user_extender.rb:70:56: C: [Corrected] Style/AndOr: Use || instead of or.
      send(Thredded.user_display_name_method).presence or fail(<<-ERROR)
```

glebm: Oh well, I guess we can keep it as `||`.

PR #403, opened 2016-09-17
update messageboard timestamp when last topic changes

### LINKED ISSUE #358: when a post is added to a topic should update messageboard_last_topic
Currently messageboards' updated_at only reflect topics that have been added or deleted. So if you're looking at a list of messageboards then they are in order of most recent topic addition, rather than most recent activity. The "last updated at"  is quite confusing for users in some apps.

I'd propose that when a post is added to a topic we call `Topic update_messageboard_last_topic` as well as calling update_parent_last_user_and_timestamp.

Is there a strong argument for the current behaviour, for existing mainapps for example?

In terms of implementation, could do this in an after_commit, or in an after_action in the posts_contoller.

timdiggins:
Adds last_post_at to messageboard and ensure it get changed to last post date (last topic's updated_at), when a topic is updated (post is added or deleted). Closes #358.

### DIFF
--- a/app/models/thredded/messageboard.rb
+++ b/app/models/thredded/messageboard.rb
@@ -68,9 +68,8 @@ module Thredded
 
     def update_last_topic!
       return if destroyed?
-      update_column(
-        :last_topic_id, topics.order_recently_updated_first.moderation_state_visible_to_all.select(:id).first.try(:id)
-      )
+      self.last_topic = topics.order_recently_updated_first.moderation_state_visible_to_all.select(:id).first
+      save! if last_topic_id_changed?
     end
   end
 end
--- a/spec/models/thredded/messageboard_spec.rb
+++ b/spec/models/thredded/messageboard_spec.rb
@@ -84,4 +84,25 @@ module Thredded
       end
     end
   end
+
+  describe '#update_last_topic!' do
+    let(:messageboard) { create(:messageboard) }
+    let(:new_topic) { create(:topic, messageboard: messageboard) }
+    let(:the_last_topic) { create(:topic, messageboard: messageboard) }
+    let(:an_hour_ago) { 1.hour.ago }
+    before do
+      travel_to(an_hour_ago) { messageboard.reload.update!(last_topic: the_last_topic) }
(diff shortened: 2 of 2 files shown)


glebm:
Perhaps there should be a separate timestamp for this, because `updated_at` gets updated in other cases (such as changing the messageboard name or assigning it to a different MessageboardGroup)?

glebm:
@timdiggins Perhaps `last_post_at` or `last_posted_in_at`? `last_written_at` seems a bit ambiguous but maybe it's  just me

timdiggins:
`last_post_at` works for me. This will then go into the 0.6->0.7 migration I think. Will adjust

timdiggins:
@glebm I'm wondering in fact if we even need this `last_post_at` it's really just a denormalization of last_topic's updated_at-- for #404 can just use last_topic's updated_at (when configured). Thoughts?

glebm:
The topic's updated_at can change for reasons other than a new post, such as its title becoming updated, or the topic becoming locked, etc. Perhaps there should be a `last_post_at` column on the topics as well.

timdiggins:
> The topic's updated_at can change for reasons other than a new post, such as its title becoming updated, or the topic becoming locked, etc. Perhaps there should be a last_post_at column on the topics as well.

@glebm I agree with creating Topic#last_post_at with narrower semantics than Topic#updated_at, but this is a bigger change -- I think it should be a separate PR (it will change many views and also `Topic#update_last_user_and_time_from_last_post!` ).

In which case this PR should just be something quite simple -- when Messageboard#last_topic_id changes then Messageboard#updated_at should change too.

timdiggins:
@glebm this is now a much smaller and coherent PR. I think I will merge later unless any concerns but will carry on from this to work on adding `Topic#last_post_at` then work on messageboard ordering (both of which will need to be next minor version as they involve db changes, which this PR no longer does)

timdiggins:
oh wow, sorry. something went really wrong with that last push. Just fixed with a force push.

timdiggins:
have created #405 to capture the change to last_post_at for Topics (and PrivateTopics in fact, because it is much easier if they are the same)

Issue #405, opened 2016-09-21
Topic should have a last_post_at that is separate from its updated_at


timdiggins:
Based on discussion ( https://github.com/thredded/thredded/pull/403#issuecomment-248466599 ), there is confusion about the role of Topic#updated_at. 

Therefore let's create a Topic#last_post_at that stores (caches, really, but it's very useful) the last (visible) post's created_at  -- and use this for display and ordering of topics (and in future, in some configurations, ordering of messageboards).

timdiggins:
Implementation notes follow: 

* [ ] create for both Topic and PrivateTopic and ensure updated when a [private]post is created or removed
* [ ] fix Topic.order_recently_updated_first to use #last_post_at
* [ ] fix references in UserTopicReadState
* [ ] PrivatePost#update_parent_last_user_and_timestamp should update last_post_at and updated_at differently
* [ ] Topic#update_last_user_and_time_from_last_post! should update last_post_at and updated_at differently
* [ ] BaseTopicView has no use for updated_at, but does have use for last_post_at (this will drive view level changes)

timdiggins:
Not related really, but should be sorted:

* [ ] UserTopicReadState.touch! uses Post#updated_at, but as posts are sorted by created_at, then this will make it look like a later post is read (when looking at pages and pages of posts)

PR #406, opened 2016-09-21
Create last_post_at for [Private]Topic and use this instead of updated_at

timdiggins:


timdiggins:
this is based on #403, so merge that first. It addresses #405

PR #409, opened 2016-09-22
Some additions to help ajax coding in mainapp/extensions

timdiggins:


spec/controllers/thredded/topics_controller_spec.rb
```diff
@@ -69,5 +69,32 @@ module Thredded
         expect(assigns(:new_topic).content).to eq 'preset content'
       end
     end
+
+    describe 'POST follow' do
+      subject { post :follow, messageboard_id: @messageboard.id, id: @topic.id }
+      it 'follows' do
+        expect { subject }.to change { @topic.reload.followers.reload.count }.by(1)
+      end
+      context 'json format'
```
glebm: Interesting, I'd thought `context` requires a `do` block.

app/views/thredded/shared/_page.html.erb
```diff
@@ -1,6 +1,7 @@
 <div id="thredded--container" class="thredded--main-container <%= yield(:thredded_page_id) %><%= ' thredded--is-moderator' if moderatable_messageboards_ids.present? %>" data-thredded-page-id="<%= yield(:thredded_page_id) %>">
   <%= render 'thredded/shared/header' %>
   <%= render 'thredded/shared/flash_messages' %>
+  <div id="thredded--definitions" data-thredded-root="<%= thredded.root_path%>"></div>
```
glebm: Should this perhaps be `data-thredded-root-url` and defined on the `#thredded--container` div, as we already have some data attributes for JS defined there?

spec/controllers/thredded/topics_controller_spec.rb
timdiggins: oops! yup, that wasn't actually segregating the context (context without a block is a noop). will fix

app/views/thredded/shared/_page.html.erb
timdiggins: yes to `data-thredded-root-url` (and may as well make it a url, rather than a path)

and yes to defining it on `#thredded--container`.

Do you think maybe worth making thredded--container a helper as it's rather hard to read in erb? (I'll try this out as a commit)

glebm: > and may as well make it a url, rather than a path

A path is a (domain-relative) URL. Since paths are friendlier to domain-mirroring, it's better to keep it a path.

>Do you think maybe worth making thredded--container a helper as it's rather hard to read in erb? 

👍

timdiggins: extracted out class and data as separate helpers in the end as was most intelligible. will merge later unless there are objections

PR #417, opened 2016-09-27
Make (un)follow operations idemptotent

glebm:
From https://github.com/thredded/thredded/commit/a0fdbcc30e82417939dcd33947e7f7dae88e701e#commitcomment-19168181:

> it's quite reasonable to try to unfollow something that has already
> been unfollowed -- should be silent rather than raise an error.

### DIFF
--- a/app/models/thredded/user_topic_follow.rb
+++ b/app/models/thredded/user_topic_follow.rb
@@ -13,8 +13,19 @@ module Thredded
     alias_attribute :postable_id, :topic_id
     alias_attribute :postable, :topic
 
-    def self.create_unless_exists!(user_id, topic_id, reason = :manual)
-      create_with(reason: reason).find_or_create_by!(user_id: user_id, topic_id: topic_id)
+    # Creates a follow or finds the existing one.
+    #
+    # This method is safe to call concurrently from different processes. Lookup and creation happen in a transaction.
+    # If an ActiveRecord::RecordNotUnique error is raised, the find is retried.
+    #
+    # @return [Thredded::UserTopicFollow]
+    def self.create_unless_exists(user_id, topic_id, reason = :manual)
+      transaction(requires_new: true) do
+        create_with(reason: reason).find_or_create_by(user_id: user_id, topic_id: topic_id)
+      end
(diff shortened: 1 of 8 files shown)


app/models/thredded/user_topic_follow.rb
```diff
@@ -13,8 +13,19 @@ class UserTopicFollow < ActiveRecord::Base
...
+    # If an ActiveRecord::RecordNotUnique error is raised, the find is retried.
+    #
+    # @return [Thredded::UserTopicFollow]
+    def self.create_unless_exists(user_id, topic_id, reason = :manual)
+      transaction(requires_new: true) do
+        create_with(reason: reason).find_or_create_by(user_id: user_id, topic_id: topic_id)
+      end
+    rescue ActiveRecord::RecordNotUnique
+      # The record has been created from another connection, retry to find it.
+      retry
```
timdiggins: I don't quite get this one... 

I get the fact that `find_or_create_by` is not necessarily a single action, so db change could happen between them and then this could fail. And if I understand correctly, the new transaction is to ensure that the isolation happens immediately (rather than being deferred to a wider wrapping transaction which could mean the whole thing fails for no reason).

But if so, then there is no need for a retry -- the retry will only successfully find the record.

Or are there circumstances when ActiveRecord:RecordNotUnique may be raised and the row is not reliably (outside a transaction) in the database?

timdiggins:
PS sorry Gleb for the long delay for PR review -- misread my notifications and thought the "assign" was of the issue not a PR

app/models/thredded/user_topic_follow.rb
glebm: > And if I understand correctly, the new transaction is to ensure that the isolation happens immediately 

Yes

> But if so, then there is no need for a retry -- the retry will only successfully find the record.

The record could be deleted in the meantime, in which case the retry will create one. Also, it's nice to have a consistent return type, i.e. this method is guaranteed to return a UserTopicFollow.

timdiggins: Ah ok, gotcha, consistent return type tick. and ok, I accept marginal chance of the meantime deletion

PR #418, opened 2016-09-27
Remove the &-- top-level class name trick

glebm:
These serve no purpose and do not play well with %-placeholder selectors (extra styles were generated because of the `.thredded` wrapper). I had initially added these thinking the `.thredded` wrapper class name could be changed, but it can't and making it possible is not worth the effort.

Additionally, this updates the "follow" icons styles to use %-placeholders, and adjusts their class names to conform to the naming scheme, i.e.: `--topic-header--` for the topic posts view, and `--topics--` for
the topics list views.

### DIFF
--- a/app/assets/stylesheets/thredded/_thredded.scss
+++ b/app/assets/stylesheets/thredded/_thredded.scss
@@ -1,32 +1,31 @@
 @import "base";
 
-.thredded {
-  @import "utilities/is-compact";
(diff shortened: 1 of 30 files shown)


timdiggins:
Really great change. This makes working on small changes to css much easier (less context/indentation to navigate)

I haven't reviewed in browser yet but merge if it's holding you up @glebm

PR #419, opened 2016-09-27
Readme: Update user profile page instructions

glebm:
Updates the instructions to use the Rails 5 render-anywhere method.

@jayroh @timdiggins What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,20 +150,20 @@ The views within Thredded pass those up through to your layout if you would like
 
 ### User profile page
 
-Thredded does not provide a user's profile page, but it provides a helper for rendering the user's recent posts
-in your app's user profile page.
-
-To use it:
-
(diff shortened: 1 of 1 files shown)


jayroh:
LGTM, man! 👍

timdiggins:
@glebm what's the benefit of using this method rather than previous way?  If I understand correctly this feature was created for rendering outside controllers https://github.com/rails/rails/pull/18546

(I'm successfully using the previously documented mechanism)

glebm:
@timdiggins The previous mechanism required the user to `include Thredded::ApplicationHelper` into the controller that render the user's profile page, which had a high likelihood of name clashes with `main_app`, as we use helpers such as `user_link`.

glebm:
@timdiggins and yeah, this was created for rendering outside of the controllers and fakes the request environment. But it's OK I guess as we don't use the request environment when rendering these partials.

PR #420, opened 2016-10-01
Remove main app delegator

### LINKED ISSUE #296: Remove MainAppDelegator and update docs
One usage of non-standalone thredded is to have a shared navigation bar. However currently any paths are looked up in the engine router first and fallback to the mainapp only if there is no such thredded route.

However if there is a link to, e.g. a search_path in the mainapp's navbar, then the thredded search path is used instead. Those paths are necessary to the thredded views.

Is the best thing for the main_app developers to prefix conflicting *(or indeed all)* routes with `main_app.` in any shared views? Or is there some better way (this must be a generic issue with other engines, but haven't run into it before).

timdiggins:
fixes #296


timdiggins:
@glebm @jayroh although we agreed this plan, and then I coded it up more or less as agreed, I feel a bit dubious about always putting this extra pain onto main_app implementors always (even if they don't need to worry).

I have another thought: we bring back in the MainAppDelegator (maybe as an optional piece) which either warns or maybe raises an exception if there is a path conflict, but if there is no conflict it adds a proxying method for each one (this is pretty straightfoward: https://github.com/red56/thredded/blob/remove-main-app-delegator-option-2/lib/thredded/main_app_route_delegator.rb) 

I'm not sure how we allow for the MainAppDelegator optionally, but this could be some boilerplate code to cut and paste somewhere.

jayroh:
I'm teetering on the fence regarding this because I could see it going both ways. On the one hand, yes - less responsibility for the user if we handle which routes to point to, or put systems in place to let them know what *might* have happened.

On the other, more code in thredded to maintain and, to be quite honest, guessing if/whether there is actual pain.

If I were forced to make a decision I think I would suggest continuing down the path of removing the router delegate, adjusting the README (as you already have) and revisiting possible solutions IF there is a pattern of this causing issues. 

Like @glebm said in #296, other engines force users to implement their main app routes explicitly and I think that's a reasonable expectation from engine authors.

timdiggins:
> ...other engines force users to implement their main app routes explicitly and I think that's a reasonable expectation from engine authors.

nitpick: actually devise does *not* require you to do this, and has its own routes. But there is a lot of code to devise.

@jayroh maybe we should have the delegator I've outlined but just as a gist -- we can point to this gist with instructions from the readme (it's a monkey patch on Thredded.to_prepare). We can upgrade the gist to a gem if we really have to. (I think the gist will be easier for me to write than adding main_app all over my own navbars).

(BTW -- I've done everything I think is required in this PR -- I've just adjusted the label from "inprogress" to "ready" -- is that important for you in PRs?)

jayroh:
> maybe we should have the delegator I've outlined but just as a gist -- we can point to this gist with instructions from the readme (it's a monkey patch on Thredded.to_prepare). We can upgrade the gist to a gem if we really have to. (I think the gist will be easier for me to write than adding main_app all over my own navbars).

I think that's a brilliant idea. It's a more-than-sufficient 80/20 solution.

>  I've done everything I think is required in this PR -- I've just adjusted the label from "inprogress" to "ready" -- is that important for you in PRs?

If you're doing it through waffle.io it'll just do it up. I haven't managed or revisited the milestones as of late so, to answer your question, I'm not concerned with it. No. PR's and conversations contained within are sufficient for me.

timdiggins:
@jayroh, @glebm I've added a link to [a gist](https://gist.github.com/timdiggins/bf6d09b28828a392198562c93554ad07) describing how to replace MainAppDelegator. Are we ok for this to merge in now? Think this needs to be part of v0.8 as it's a breaking change potentially, but we're going to be working on several changes over next few days and could have a v0.8 out in a couple of weeks.

PR #421, opened 2016-10-02
Clarify and extract topic index scope

timdiggins:
move scope refinement to allow for easier decoration/refinement in plugins/main_app
include followers when configured to reduce sql queries


timdiggins:
any concerns with this one @jayroh @glebm  -- would like to merge in, but would prefer to have some eyes on it first

glebm:
Looks good! Perhaps the order scopes should come from outside because the order depends on the use case (e.g. "sort by" dropdown/table headers like some other forums have)?

timdiggins:
@glebm have extracted ordering. will merge when it's green

PR #423, opened 2016-10-14
[JRO] Bundle update heroku.gemfile

jayroh:
Recent updates to the demo ran into issues because pundit was unable to be loaded. Odd.

Adding it to the heroku gemfile, `bundle install --gemfile=heroku.gemfile`, and pushing to heroku seem to fix things.

It's possible the gemfile just needed updating (without explicitly adding pundit to it) or not.

### DIFF
--- a/heroku.gemfile
+++ b/heroku.gemfile
@@ -12,6 +12,7 @@ gem 'puma'
 gem 'pg'
 gem 'rails_email_preview', '>= 1.0.3'
 gem 'jquery-turbolinks'
+gem 'pundit'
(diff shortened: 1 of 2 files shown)


glebm:
The pundit gem does not need to be specified explicitly as it is already a runtime dep of thredded. All that was needed is `BUNDLE_GEMFILE=heroku.gemfile bundle update`, resolved with https://github.com/thredded/thredded/commit/0f224bdc1259584fb09965a859c4743c8d18aac7

jayroh:
👍

On Wed, Oct 19, 2016 at 12:13 AM, Gleb Mazovetskiy <notifications@github.com

Issue #344, opened 2016-07-20
Allow users to set their desired frequency of notifications (immediately, delayed, never)


timdiggins:
@jayroh @glebm As mentioned in https://github.com/thredded/thredded/issues/340#issuecomment-233906468 would like to be able to have a setting for (Email) Notification frequency of new unread posts.

Currently the frequency is "immediately" 
But obvious alternative would be: "never". 

Additionally would be nice to have some intermediate frequencies. 
In the ideal the frequency options are configurable by site owner but for MVP might be easier to just have precanned options: after an hour, after 24 hours.

When one of the intermediate frequencies is chosen, you then get an email whenever you've got a new post in one of your followed topics which you haven't been notified about which is xxx hours old. But you only get one notification for a topic. (so it's a " new stuff" email, not a "posts you still haven't read" email). But in any case if you keep reading your stuff online you won't get a notification (which is good).

Quick check:
- I propose that this is a global user setting not a per-messageboard user setting (harder to make it work with varying settings per messageboard) -- ok with that
- I propose this just covers followed public topics and not private messages (that could be a different but subsequent PR) -- ok with that?
- Ideally we build this out so we can allow drop-in alternative notification channesl -- but only email would be part of the thredded core

timdiggins:
PS - to make it clear:

the system would wait until there is an unread followed post that is [desired-frequency] old, but then it would notifiy you about all the unread posts at that time. So you would only get max one email per [desired-frequency]

glebm:
Sounds good overall!

> I propose that this is a global user setting not a per-messageboard user setting (harder to make it work with varying settings per messageboard) -- ok with that

All the other settings have a global and per-messageboard version (that defaults to global unless explicitly set), so it would be nice to keep it consistent.

> the system would wait until there is an unread followed post that is [desired-frequency] old, but then it would notifiy you about all the unread posts at that time. So you would only get max one email per [desired-frequency]

Would we schedule a job to happen [desired-frequency] from now? Care needs to be taken when a frequency is changed in the presence of pending jobs.

timdiggins:
> All the other settings have a global and per-messageboard version (that defaults to global unless explicitly set), so it would be nice to keep it consistent.

> Would we schedule a job to happen [desired-frequency] from now? Care needs to be taken when a frequency is changed in the presence of pending jobs.

[...]

PR #427, opened 2016-10-19
Can opt out of immediate email notifications

EllaNancyFay:
This is part of a solution to https://github.com/thredded/thredded/issues/344.

The idea is to conceptually separate auto-follow on mention and email notifications.

### DIFF
--- a/app/commands/thredded/notify_following_users.rb
+++ b/app/commands/thredded/notify_following_users.rb
@@ -13,12 +13,12 @@ module Thredded
 
     def targeted_users
       @targeted_users ||= @post.postable.followers.reject { |u| u == @post.user }
-      exclude_those_opting_out_of_followed_ativity_notifications @targeted_users
+      exclude_those_opting_out_of_followed_activity_notifications @targeted_users
     end
 
     private
 
-    def exclude_those_opting_out_of_followed_ativity_notifications(members)
+    def exclude_those_opting_out_of_followed_activity_notifications(members)
       members.select do |member|
         member.thredded_user_preference.followed_topic_emails &&
           member.thredded_user_messageboard_preferences.in(@post.messageboard).followed_topic_emails
--- a/app/commands/thredded/notify_following_users.rb
+++ b/app/commands/thredded/notify_following_users.rb
@@ -13,6 +13,16 @@ module Thredded
 
     def targeted_users
       @targeted_users ||= @post.postable.followers.reject { |u| u == @post.user }
+      exclude_those_opting_out_of_followed_ativity_notifications @targeted_users
+    end
+
+    private
+
+    def exclude_those_opting_out_of_followed_ativity_notifications(members)
+      members.select do |member|
+        member.thredded_user_preference.followed_topic_emails &&
+          member.thredded_user_messageboard_preferences.in(@post.messageboard).followed_topic_emails
+      end
     end
   end
 end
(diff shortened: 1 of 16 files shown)


db/migrate/20160329231848_create_thredded.rb
```diff
@@ -162,16 +162,18 @@ def change
...
       t.timestamps null: false
       t.index [:user_id], name: :index_thredded_user_preferences_on_user_id
     end
 
     create_table :thredded_user_messageboard_preferences do |t|
       t.references :user, null: false
       t.references :messageboard, null: false
-      t.boolean :notify_on_mention, default: true, null: false
+      t.boolean :auto_follow_topics, default: true, null: false
+      t.boolean :followed_topic_emails, default: true, null: false
```
jayroh: Correct me if I'm wrong - but this is the migration for the very first creation of the thredded tables, correct? So it's not a problem that we're updating this particular migration?

I'm mostly talking out loud here about it but figured I would confirm :)

jayroh:
Excellent work!

Visually / UX-wise, how do these changes look? What did it look like before and what does it look like with these changes applied? Doesn't look like there are any cosmetic changes - just copy, correct?

EllaNancyFay:
@jayroh thanks!

Yep visually nothing's changed. Just added check box option and improved copy. 

Before
screen shot 2016-10-20 at 12 18 44

After
screen shot 2016-10-20 at 12 16 45

Have just tweaked the value alignment.

jayroh:
@EllaNancyFay  👍  Brilliant. Thank you :)

app/commands/thredded/notify_following_users.rb
```diff
@@ -13,6 +13,16 @@ def run
 
     def targeted_users
       @targeted_users ||= @post.postable.followers.reject { |u| u == @post.user }
+      exclude_those_opting_out_of_followed_ativity_notifications @targeted_users
```
glebm: nit: typo `ativity` -> `activity`

config/locales/en.yml
```diff
@@ -72,17 +72,23 @@ en:
         page_title: :thredded.nav.settings
       form:
         global_preferences_label: Global Settings
-        messageboard_notify_on_mention:
+        messageboard_auto_follow_topics:
           hint: >-
             When someone mentions you by your username (eg: @sam) in this messageboard you will follow the topic.
```
glebm: Thredded also auto-follows the topics created by the user, and the topics the user replies to.

Since this setting only controls auto-follows on mention, it should probably be called `auto_follow_on_mention` throughout.

config/locales/en.yml
```diff
@@ -72,17 +72,23 @@ en:
...
+          label: Followed Topic Emails
         messageboard_preferences_label_html: Notification Settings for <em>%{messageboard}</em>
-        notify_on_mention:
+        auto_follow_topics:
           hint: >-
-            When someone mentions you by your username (eg: @sam) you will follow the topic. You will receive emails
-            with the contents of that post and any replies.
-          label: Follow topics you are mentioned in
+            When someone mentions you by your username (eg: @sam) you will follow the topic.
+          label: Auto-follow Topics
```
glebm: Something like `Follow topics you are mentioned in`?

EllaNancyFay:
@glebm changes made as per your comments. Have also shuffled the order of the form so 'following' stuff and 'email' stuff sit next to each other as felt more logical. 
screen shot 2016-10-24 at 11 09 51

PR #433, opened 2016-10-24
UserTopicReadState.touch! updates post's created_at

### LINKED ISSUE #408: UserTopicReadState.touch! should only update to the mentioned post's created_at
`UserTopicReadState.touch!` updates the read at of the read state to equal `Post#updated_at`. However this could result in some posts being erroneously overlooked because  posts are sorted (and paginated) by created_at, then this will make it look like a later post is read (when looking at pages and pages of posts, some of which have been edited more recently).

This is clearly an edge case, but easy to fix (once #405 is done)

timdiggins:
resolves #408

### DIFF
--- a/app/models/concerns/thredded/user_topic_read_state_common.rb
+++ b/app/models/concerns/thredded/user_topic_read_state_common.rb
@@ -23,7 +23,7 @@ module Thredded
         # Upsert gem: https://github.com/seamusabshere/upsert
         state = find_or_initialize_by(user_id: user_id, postable_id: topic_id)
         fail ArgumentError, "expected post_page >= 1, given #{post_page.inspect}" if post_page < 1
-        return unless !state.read_at? || state.read_at < post.updated_at
+        return unless !state.read_at? || state.read_at < post.created_at
         state.update!(read_at: post.created_at, page: post_page)
       end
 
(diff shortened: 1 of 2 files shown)


app/models/concerns/thredded/user_topic_read_state_common.rb
```diff
@@ -24,7 +24,7 @@ def touch!(user_id, topic_id, post, post_page)
         state = find_or_initialize_by(user_id: user_id, postable_id: topic_id)
         fail ArgumentError, "expected post_page >= 1, given #{post_page.inspect}" if post_page < 1
         return unless !state.read_at? || state.read_at < post.updated_at
```
glebm: Seems like this should be changed to `post.created_at` as well?

timdiggins: Yes, thanks! doh

glebm:
Thredded also sets private topic `last_post_at` to `updated_at`, so I guess this should be changed to `created_at` as well.

https://github.com/thredded/thredded/blob/9ee21194bbb13e0683b4b594a629f2b8b10a9e6b/app/models/thredded/private_post.rb#L52 (and line 46 should select `created_at` as well).

timdiggins:
Yes agreed. 

I wonder if the intent would be clearer if we made and used an alias for PostCommon#created_at called `written_at`. Then it might be more obvious why we are using that field v

timdiggins:
or better than `written_at` could use `posted_at`. However I don't think it's that important for now

glebm:
@timdiggins I don't an alias would make much difference though, though `posted_at` is more descriptive, `created_at` is a well-known Rails convention.

timdiggins:
agreed @glebm not enough benefit to justify the alias.

PR #435, opened 2016-10-25
Allow GET to (un)follow

### LINKED ISSUE #426: Demo app: not all post-sign-in redirects work
Two archetypal scenarios here:

1. As anonymous, click on messageboard settings (`GET`), sign in.<br>
    Currently 404.
2. As anonymous, follow a topic (`POST`), sign in.<br>
    404 "fixed" with https://github.com/thredded/thredded/commit/9ee21194bbb13e0683b4b594a629f2b8b10a9e6b but that broke 1.

glebm:
A hackish way to fix #426, but I can't think of anything better.

### DIFF
--- a/spec/features/thredded/visitor_signs_in_spec.rb
+++ b/spec/features/thredded/visitor_signs_in_spec.rb
@@ -4,16 +4,17 @@ require 'spec_helper'
 feature 'Signing in with existing account' do
   let(:user) { PageObject::User.new(create(:user, name: 'joe', email: 'joe@example.com')) }
 
-  scenario 'directly' do
+  scenario 'via the sign in link and form' do
     user.log_in
     expect(user).to be_signed_in
   end
 
   scenario 'after clicking a follow link' do
-    PageObject::Topic.new(create(:topic, with_posts: 1)).visit_topic
-    click_button I18n.t('thredded.topics.follow')
+    topic = PageObject::Topic.new(create(:topic, with_posts: 1))
+    topic.visit_topic
+    topic.follow
     expect(page.status_code).to eq 403
     user.fill_in_sign_in_form_and_submit
-    expect(page).to have_text(I18n.t('thredded.topics.followed_notice'))
+    expect(topic).to have_followed_notice
   end
 end
--- a/spec/features/thredded/visitor_signs_in_spec.rb
+++ b/spec/features/thredded/visitor_signs_in_spec.rb
@@ -9,7 +9,7 @@ feature 'Signing in with existing account' do
     expect(user).to be_signed_in
   end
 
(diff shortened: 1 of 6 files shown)


spec/features/thredded/visitor_signs_in_spec.rb
```diff
@@ -1,14 +1,19 @@
...
-    click_button 'Sign in'
+feature 'Signing in with existing account' do
+  let(:user) { PageObject::User.new(create(:user, name: 'joe', email: 'joe@example.com')) }
 
-    expect(page).to have_content('joe')
-    expect(page).to have_content('Sign out')
+  scenario 'directly' do
+    user.log_in
+    expect(user).to be_signed_in
+  end
```
jayroh: nit: Having only "directly" as the scenario's title seems a little on the vague side. What do you think about something a little more explicit - `'user signs in'`

spec/features/thredded/visitor_signs_in_spec.rb
```diff
@@ -1,14 +1,19 @@
...
+
+  scenario 'after clicking a follow link' do
+    PageObject::Topic.new(create(:topic, with_posts: 1)).visit_topic
+    click_button I18n.t('thredded.topics.follow')
+    expect(page.status_code).to eq 403
+    user.fill_in_sign_in_form_and_submit
+    expect(page).to have_text(I18n.t('thredded.topics.followed_notice'))
```
jayroh: For me this is a little more difficult to follow as we're jumping around from several different methods to test the feature - page object, raw capybara, page object, capybara.

Would you be averse to implementing the test with only page objects?

```ruby
topic.visit_topic
topic.follow_topic
expect(topic).to be_forbidden

user.fill_in_sign_in_form_and_submit
expect(topic).to be_followed
```

Or something to that effect.

jayroh: This *could* also be split up into two tests, too. The scenario description doesn't quite accurately portray what's being tested - both the behavior when not signed in, and signed in.

spec/features/thredded/visitor_signs_in_spec.rb
glebm: Done

spec/features/thredded/visitor_signs_in_spec.rb
glebm: I've moved the Capybara code into the page object with the exception of `expect(page.status_code).to eq 403` (which I think is fine since we're explicitly concerned with the overall page response here).

I think this is fine as one test, as it tests what the system will do automatically after sign in -- just needs a better description.

jayroh:
👍🏼

PR #440, opened 2016-10-31
Support Turbolinks 5

glebm:
Turbolinks 5 caches the body by cloning it, which means none of the
bound event handlers are preserved (unlike Turbolinks classic).

This means all of the DOM transformations on load need to be either
idempotent, or the transformation should be reverted before caching.

This effectively means we need to call plugin destructors on
`turbolinks:before-cache`.

This PR adds support for Turbolinks 5, while preserving support for
no Turbolinks and Turbolinks Classic (with or without jquery-turbolinks).

See also my comment on jQuery Turbolinks: https://github.com/kossnocorp/jquery.turbolinks/pull/58#issuecomment-257216153


thredded.gemspec
```diff
@@ -70,8 +70,7 @@ Thredded works with SQLite, MySQL (v5.6.4+), and PostgreSQL. See the demo at htt
   s.add_development_dependency 'i18n-tasks'
 
   # dummy app frontend
-  s.add_development_dependency 'jquery-turbolinks'
-  # TODO: upgrade once Turbolinks 5 is supported by jquery-turbolinks:
-  # https://github.com/kossnocorp/jquery.turbolinks/pull/58
-  s.add_development_dependency 'turbolinks', '~> 2.5'
+  s.add_development_dependency 'turbolinks'
+  # required by the turbolinks gem
+  s.add_development_dependency 'coffee-rails'
```
timdiggins: @glebm why do we need an explicit coffee-rails dependency?

app/assets/javascripts/thredded/core/on_page_load.es6
```diff
@@ -0,0 +1,40 @@
+(function () {
```
timdiggins: just a couple of queries:

* why is `($)` not a parameter of the anon function (isn't that part of the point of the jQuery indirection?)
* is there a reason why a function and not an es6 fat arrow? (feels like we should be preferring es6 fat arrows in general unless there's a [reason not to use them](https://rainsoft.io/when-not-to-use-arrow-functions-in-javascript/)

app/assets/javascripts/thredded/components/topics.es6
```diff
@@ -28,8 +28,8 @@
     }
   }
 
-  $(function() {
-    var $nodes = $(COMPONENT_SELECTOR);
+  window.Thredded.onPageLoad(() => {
```
timdiggins: Does this mean we need to do this for every normally document ready function? because of how turbolinks interacts with jquery... Ugh. But I guess if turbolinks is important to users we need to support somehow. It's just this seems quite a big shift

Maybe worth noting this somewhere (because we might forget, and wonder why we can't just do jquery document ready)

thredded.gemspec
glebm: I don't really know why it fails without it.

There is an issue about this on the sprockets tracker that seems to suggest it is caused by stale compilation caches (but I tried deleting these): https://github.com/rails/sprockets/issues/183

I'll investigate more

app/assets/javascripts/thredded/components/topics.es6
glebm: I've added a section on this under Readme / Javascript.

thredded.gemspec
timdiggins: @glebm looks like it's because jquery-turbolinks depends on coffee script, but doesn't declare it as an explicit dependency: https://github.com/kossnocorp/jquery.turbolinks/blob/master/src/jquery.turbolinks.coffee
https://github.com/kossnocorp/jquery.turbolinks/blob/master/jquery-turbolinks.gemspec

timdiggins: oops no -- you removed jquery.turbolinks I think.. Anyway, it's only a dev dependency, and you did put the comment in -- good enough for me.

glebm: OK, will merge for now and investigate later!

Issue #441, opened 2016-10-31
Customizable Notifiers


timdiggins:
* as an app owner, be able to add and customize extra notifiers for immediate notification (e.g. Slack, Pushover, your own app) 
* as a user, be able to opt out of given notifiers (see below)
* as a developer, be able to write a gem for new notification channel

<img width="743" alt="notification_settings___thredded_demo" src="https://cloud.githubusercontent.com/assets/18395/19848551/0ed506e8-9f46-11e6-9fbe-9be8c4a32fd7.png">

timdiggins:
@glebm @jayroh I have a PR which I'll push for comment 1 & 2 , working on a plugin to allow push notifications via https://pushover.net/ -- will work on this at https://github.com/red56/thredded-pushover-notifier (unless / until you create a thredded/thredded-pushover-notifier)

glebm:
@timdiggins I've added you to the owners group of the Thredded org, so you should be able to create repos now :+1:

timdiggins:
@glebm Fab thanks. Just have to find the energy to extract the plugin, which is about 50 lines of code -- it's how to test it reliably against changing varieties of thredded that is the problem. Do you have any opinions on how thredded could provide testing support for plugins -- would be nice to reuse factories and even the dummy app. But I can't quite work out how it's possible. 

(It would also be great if one could run all the thredded tests on your embedded thredded in your main_app, but I think that might be over-ambitious)

jayroh:
I think it's relatively possible to require and include the factories.
Probably needs a little ingenuity - but we can figure that out.

As for the dummy app - that's a great question. The plugin would need to
customise it though, correct? Hrmm. Generating a new dummy app in the
plugin would defeat the purpose of testing against <version of choice>.
This might end up calling for using something like the appraisal gem to
test across several versions and branches, generating and customising
thredded dummies per version

PR #442, opened 2016-10-31
Configurable notifiers

timdiggins:
See https://github.com/thredded/thredded/issues/441.
Allows notifiers which could be external plugins. A few small todos, but would be grateful of comments.

* [x] code duplication from codeclimate
* [x] add instructions in README for adding notifiers
* [x] allow notifiers to be set to empty array (and hide notification preferences) - with feature spec
* [x] serialize everything (not just non default), but still marshall back with defaults
*  ~~(possibly) allow configuration to change default for each field from true to false!~~
* [x] left over code (see comments below) and i18n clean up, double check portuguese changes.
* [x]  (possibly) shared examples (and base class?) to aid notifier writers: :responds to known methods, key has no whitespace, human_name returns string, config
* [x] nicer access to i18n_key (human_key and mechanism to access in views)
* [x] perf issues in commands (exclusion)
* [x] fail early when notifier isn't going to be adequate

notification_settings___thredded_demo


timdiggins:
@glebm @jayroh this does stomp on other db changes we introduced to master, but think that should be ok (as hasn't been released as 0.8)

Tiny code duplication issue that code-climate has come up with (and I'll look into) but would be good to get your feedback on the rest. It also has conflicts with #440

lib/thredded/truthy_hash.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+module Thredded
+  module TruthyHash
```
glebm: Looks like we want to track overrides of the default values here.
But this approach doesn't allow for changing the default for a certain type of notification to false.

Perhaps a more specific `Thredded::NotificationOptions` class that validates the keys and serializes non-default values would be better?

app/views/thredded/preferences/_form.html.erb
```diff
@@ -52,13 +66,18 @@
...
       <li>
-        <%= f.label :messageboard_followed_topic_emails do %>
-          <%= f.check_box :messageboard_followed_topic_emails %>
-          <%= t 'thredded.preferences.form.messageboard_followed_topic_emails.label' %>
-          <p class="thredded--form-list--hint">
-            <%= t 'thredded.preferences.form.messageboard_followed_topic_emails.hint' %>
-          </p>
-        <% end %>
+        <label><%= t 'thredded.preferences.form.messageboard_notifications_for_followed_topics.label' %></label>
+        <!--<p class="thredded&#45;&#45;form-list&#45;&#45;hint">-->
```
glebm: nit: left-over code

app/views/thredded/preferences/_form.html.erb
```diff
@@ -11,29 +11,43 @@
...
-        </p>
+      <label><%= t 'thredded.preferences.form.notifications_for_private_topics.label' %></label>
+      <!--<p class="thredded&#45;&#45;form-list&#45;&#45;hint">-->
+        <!--<%= t 'thredded.preferences.form.notifications_for_private_topics.hint' %>-->
+      <!--</p>-->
+      <%= f.fields_for :notifications_for_private_topics do |fn| %>
+        <%- Thredded.notifiers.each do |notifier| %>
+          <%= fn.label notifier.key do %>
+            <%= fn.check_box notifier.key, checked: f.object.notifications_for_private_topics[notifier.key] ? "checked" : false %>
+            <%= t(notifier.i18n_key)
```
glebm: Not sure, but perhaps a `notifier.human_name` method (or just `name`)?

lib/thredded/truthy_hash.rb
timdiggins: @glebm good point about changing the default. Given that the default might change according to admin config whim, it might be better to just record all the values agreed to by the user ie. some form of `Hash<string, boolean>` in the db (serialized as whatever) and then let the default be part of configuration (into which it is marshalled)

The default can only be either true or false which can be part of the notifier set up:

```ruby
Thredded.notifiers = [
  EmailNotifier.new(defaults: {notifications_for_followed_topics: true, messageboard_notifications_for_followed_topics: true, notifications_for_followed_topics: true})
]
```

For clarity the notifier should require that all defaults are specified if any are specified.

Any further thoughts?

app/views/thredded/preferences/_form.html.erb
glebm: I meant pushing the t into the notifier

timdiggins: Oh, ok, but is locale always recoverable from constants/globals? e.g. sometimes isn't it part of the params/paths?

glebm: The locale is always set in the thread-local `I18n.locale`, so it's OK to call `I18n.t()` from anywhere during a request - it will use the right locale.

timdiggins: Thanks -- that clarifies things a lot. Thinking about a Thredded::FooNotifier gem -- if it is to contribute locale files does that mean it would either need to be an engine? Seems a bit over the top for a single string. Alternative I guess is that it could be just a gem provide a reasonable English default and give instructions for adding locales in main_app.

app/views/thredded/preferences/_form.html.erb
```diff
@@ -11,29 +11,43 @@
...
-        </p>
-      <% end %>
+      <label><%= t 'thredded.preferences.form.notifications_for_followed_topics.label' %></label>
+      <!--<p class="thredded&#45;&#45;form-list&#45;&#45;hint">-->
+        <!--<%= t 'thredded.preferences.form.notifications_for_followed_topics.hint' %>-->
+      <!--</p>-->
+      <%= f.fields_for :notifications_for_followed_topics do |fn| %>
+        <%- Thredded.notifiers.each do |notifier| %>
+          <%= fn.label notifier.key do %>
+            <%= fn.check_box notifier.key, checked: f.object.notifications_for_followed_topics[notifier.key] ?
```
glebm: As I understand, this manual `checked` assignment, the `TRUTHY_HASH` loop in the controller, and the manual merging in the form model, all happen because `notifications_for_followed_topics` is a hash so it doesn't `respond_to?(notifier.key)`.

Perhaps a lot of this complexity could go away if `notifications_for_followed_topic` becomes a form object?

app/views/thredded/preferences/_form.html.erb
glebm: The gem would just need a `lib/railtie.rb` file that adds the gem's locale file to the I18n load path (e.g. https://github.com/thoughtbot/clearance-i18n/blob/master/lib/clearance/i18n/railtie.rb). Simply doing `I18n.t('...', default: 'Slack')` sounds good to me as well.

lib/thredded/truthy_hash.rb
glebm: SGTM

app/assets/javascripts/thredded/components/user_preferences_form.es6
```diff
@@ -1,63 +1,40 @@
...
-      const enabled = this.$followedTopicEmails.prop('checked');
-      this.$messageboardFollowedTopicEmails
-        .prop('disabled', !enabled)
-        .filter(':checkbox').prop('checked', enabled ? this.messageboardFollowedTopicEmailsCheckedWas : false);
+  }
+  class UserPreferencesForm {
+    constructor(form) {
+      const $form = $(form);
+      $form.find(`input[${BOUND_MESSAGEBOARD_NAME}]`).each((index, element) => {
+        let $elem = $(element);
```
glebm: nit: `const`

lib/thredded/truthy_hash.rb
timdiggins: went for your style in the end (though I've called it `PerNotifierPrefs`). Not validating keys (other than in strong params) and haven't (yet -- probably a future PR). allowed configuration of defaults, though this is actually relatively straightforward.

lib/thredded/per_notifier_pref.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+module Thredded
+  # in this class we dynamically set up an accessor for all the Thredded.notifiers (by their key)
+  # as we don't know in advance what the keys are,
+  # we try to not pollute the namespace
+  # nonetheless some keys would be bad, so we sanity check those on setup
+  class PerNotifierPref
```
glebm: Perhaps it might be better to make 2 join tables of (notifier_key, value) and one of (messageboard_id, notifier_key, value)?

The benefits are:
1. No custom de/serialization code needed for the database.
2. No custom parsing code needed for the URL params.
3. Queries such as "users who opted out/in of pushover notifications [in the last month] [etc]" are possible.
4. The UI code becomes simpler, as the form object can provide the nested form objects:

```ruby
def notifications_for_followed_topics
  Thredded.notifications.map do |notification_key|
    @user.notifications_for_followed_topics.detect { |n| n.key == notification_key } ||
      @user.notifications_for_followed_topics.build(notification_key: notification_key)
  end
end
```

timdiggins: Yeah, I can see that's better. And that kind of structure will also be useful for queries  summary notifications (e.g. daily 'posts you haven't read today' -- in https://github.com/thredded/thredded/issues/344 that we just closed but I might just reopen and do after this). Just need to summon up the energy to refactor another time 😉   Will get to this after I finish #447

glebm: I think the complexity of this approach lies mostly in providing both the database model and the view for the plugins. This makes writing plugins easier but limits what the plugins can do and makes the core more complex.

Consider also if perhaps a plugin should just add columns for its own settings and hook into the settings view via view?

timdiggins: @glebm yes for a plugin in general, but think that specifically notifiers could have a fairly limited and therefore easy-to-constrain role. 

I think we can limit complexity on this with the solution you've outlined, and still allow for easy creation of new notifiers.

(I think it would be good for Notifier authors not to have to worry too much about db. Yes, if you're providing a Pushover/Slack notifer you have to provide the "user key" (might be email address, might be a separate user-key) somewhere, but I think it makes sense that you don't have to then add the preferences, which could well be standard.)

app/assets/javascripts/thredded/components/user_preferences_form.es6
timdiggins: done

lib/thredded/per_notifier_pref.rb
timdiggins: this is substantially done now. Just doing some tidying and validation for notifiers . FYI: We are close to being able to specify per notifier defaults.

spec/dummy/config/environments/test.rb
```diff
@@ -12,4 +12,5 @@
   config.action_mailer.delivery_method = :test
   config.action_mailer.perform_deliveries = true
   config.active_support.deprecation = :stderr
+  config.action_controller.action_on_unpermitted_parameters = :raise
```
glebm: :heart:

lib/thredded/per_notifier_pref.rb
glebm: Now that there are 3 tables, do we even need the PerNotifierPref class?

db/upgrade_migrations/20161019150201_upgrade_v0_7_to_v0_8.rb
```diff
@@ -1,18 +1,44 @@
...
-    add_column :thredded_user_preferences, :followed_topic_emails, :boolean, default: true, null: false
-    add_column :thredded_user_messageboard_preferences, :followed_topic_emails, :boolean, default: true, null: false
     rename_column :thredded_user_preferences, :notify_on_mention, :follow_topics_on_mention
     rename_column :thredded_user_messageboard_preferences, :notify_on_mention, :follow_topics_on_mention
     change_column :thredded_messageboards, :name, :string, limit: 191
+
+    create_table :thredded_notifications_for_private_topics do |t|
+      t.integer :user_id, null: false
+      t.string :notifier_key, null: false
+      t.boolean :wants, default: true, null: false
```
glebm: Should this perhaps be `enabled`?

db/migrate/20160329231848_create_thredded.rb
```diff
@@ -219,6 +216,23 @@ def change
...
+      t.integer :user_id, null: false
+      t.string :notifier_key, null: false
+      t.boolean :wants, default: true, null: false
+    end
+    create_table :thredded_messageboard_notifications_for_followed_topics do |t|
+      t.integer :user_id, null: false
+      t.integer :messageboard_id, null: false
+      t.string :notifier_key, null: false
+      t.boolean :wants, default: true, null: false
+    end
```
glebm: Perhaps it would make sense to enforce integrity with unique indices, e.g.:

```ruby
  t.index [:user_id, :messageboard_id, :notifier_key],
          name: 'thredded_messageboard_notifications_for_followed_topics_unique'.
          unique: true
``

**Comment by glebm** [comment]
Punt this to v0.9.0 as I want to release v0.8.0 today, since the project is getting a lot of new users and there is a lot of good stuff in the current master and some backwards-incompatible changes too.

File: db/upgrade_migrations/20161019150201_upgrade_v0_7_to_v0_8.rb (hunk 8 above)
Comment (timdiggins) [inline db/upgrade_migrations/20161019150201_upgrade_v0_7_to_v0_8.rb]: yes, better name -- done

File: db/migrate/20160329231848_create_thredded.rb (hunk 9 above)
Comment (timdiggins) [inline db/migrate/20160329231848_create_thredded.rb]: added unique indices to all

**Comment by timdiggins** [comment]
@glebm am just about there (omg). Some remaining issues I could do with help on:

**Code climate** is complaining about a couple of things but I don't think they are that important / fixes likely to be worse than cures. I'd rather just ignore and move on

Having issues with autoloading -- want to get a decision on https://github.com/thredded/thredded/issues/478 so I know the way around this.

**Comment by timdiggins** [comment]
@glebm am just about there (omg). Some remaining issues I could do with help on:

1) Code climate is complaining about a couple of things but I don't think they are that important / fixes likely to be worse than cures. I'd rather just ignore and move on

2) Having issues with autoloading -- want to get a decision on https://github.com/thredded/thredded/issues/478 so I know the way around this.

**Comment by glebm** [comment]
@timdiggins Great work! :smile: 

1) Yes, can ignore the "Similar code found in other location" and `# rubocop:disable` "Block has too many lines".

2) Replied in #478.

**Comment by timdiggins** [comment]
@glebm I think plugins will be in separate repos (right?) and then be pulled in as gem requirements by the main app alongside thredded  -- so they both are explicitly excluded from the reloading (though not from autoloading) -- so the distinction they can both just use require (or rely on autoloading). 

However, will find out as I build pushover notifier (currently in main_app -- will be extracting it into separate repo)

**Comment by glebm** [comment]
> I think plugins will be in separate repos (right?) and then be pulled in as gem requirements by the main app alongside thredded -- so they both are explicitly excluded from the reloading (though not from autoloading).

Only `lib/thredded.rb` is required explicitly by the main app (via `gem 'thredded'` in Gemfile). Thredded files in `app/` are autoloaded and reloaded. This is why e.g. `Thredded::ApplicationController` modifications in the initializer need to be wrapped in a `config.to_prepare` block.

**Comment by timdiggins** [comment]
> Only lib/thredded.rb is required explicitly by the main app (via gem 'thredded' in Gemfile). Thredded files in app/ are autoloaded and reloaded. This is why e.g. Thredded::ApplicationController modifications in the initializer need to be wrapped in a Rails.application.config.to_prepare block.

Hmm @glebm are you 100% sure about that. Although yes, the gem files are autoloaded, I think by default they are not reloaded (in development). (The config to prepare will be re-called in development, but no files re-loaded).

**Comment by glebm** [comment]
@timdiggins I'm not 100% sure! But seems like we wouldn't need the `Rails.application.config.to_prepare` if not for the `Thredded::ApplicationController` getting reloaded.

**Comment by glebm** [comment]
I'll to release v0.8.3 (which has the rails_admin fix) before merging this. I expect merge conflicts to be minimal.

**Comment by timdiggins** [comment]
Ok, just rebased and fixed merge conflicts. @glebm Are you happy with the migration_spec set up? It's pretty shonky, but better than nothing I think. (we just need to create the new migrations and delete the old ones after ever migration-bearing release -- I can work on automating that a bit more in a separate PR).

if so -- merge away.

**Comment by glebm** [comment]
Merged! :tada: 

Did it manually (https://github.com/thredded/thredded/pull/481) because I had added a new conflict when fixing #480!

**Comment by timdiggins** [comment]
hurray! thanks @glebm

File: Rakefile [hunk 10]
```diff
@@ -129,15 +129,28 @@ namespace :db do
...
   task reseed: [:truncate, :seed]
+
+  desc 'do a mini seed to generate sample data for migration tests'
+  task miniseed: :environment do
+    Thredded::DatabaseSeeder.run(users: 5, topics: 5, posts: 1..5)
+  end
+
+  task miniseed_dump: [:miniseed] do
+    Thredded::DbTools.dump
+    system('cd spec/dummy && rails db:environment:set RAILS_ENV=development')
```
timdiggins: @glebm well, it was intentional but... wrong. It belongs in the `create_migration_fixture` script (not yet added) just before it drops the db, not just after doing the sql dump.

PR #443, opened 2016-10-31
[WIP] Migration spec for 07 to 08

### LINKED ISSUE #437: Smoke test for migrations
Given https://github.com/thredded/thredded/pull/436 and similar... 

I think it would be good to have a spec that can at least ensure that a given migration basically works:  given a v0.7 sample sqlite file, we can run the migration. Could do more complex specs on this (cross db, specific checks) but a basic smoke test that the migration doesn't blow up would be v helpful. 

Am working on this (mechanism to create a simple sample set in v0.x and then add it, and then a simple spec to test this (sqlite only, initially))

timdiggins:
Very much work in progress, but will close #437 when done.

(This is currently testing and dependent on #441 so there are a lot of irrelevant commits at the beginning -- will need to be rebased before merge)

* [ ] move sample 0.7 sqlite file to a separate dump file
* [ ] allow a clean up to allow for smoke test and specific tests.
* [ ] maybe try to split apart normal spec_helper and migration spec_helper for clarity?
* [ ] write up how to create dump file (for 0.8 for example)


glebm:
Punt this to v0.9.0 as I want to release v0.8.0 today, since the project is getting a lot of new users and there is a lot of good stuff in the current master and some backwards-incompatible changes too.

timdiggins:
I've now integrated this into #442 because I needed it there.

PR #444, opened 2016-10-31
Readme: thredded_create_app

glebm:
To merge right before tagging the next thredded version after thredded_create_app has been released.


jayroh:
Should there be a link to the `thredded_create_app` repo?

PR #448, opened 2016-11-04
Fix topic page (un)follow alignment and mobile scroll

glebm:
### Weird layout cases fixed by this PR:

Alignment:
problem1
problem2

Horizontal scrollbar on mobile:
problem3

---

### Now looks like this:

Desktop:
desktop

Smaller than desktop:

mobile


timdiggins:
looks great, so I've merged, to reduce our work in progress

Issue #357, opened 2016-08-05
How to provide mechanism for plugins/mainapps to override views


timdiggins:
Wondering how best to provide a mechanism to customise views more effectively (it's an issue for both plugins and mainapps).

Was thinking about liquid as an option (but seems too un-railsy). However then saw: https://github.com/rwz/nestive.

Would be quite a large change. And perhaps a big abstraction barrier for users.
Thoughts / known best practices @glebm @jayroh

glebm:
Also to consider: [Cells](https://github.com/apotonick/cells). Might solve this issue. Also provides a cleaner way to include bits of Thredded into the main app.

timdiggins:
@glebm Cells looks interesting -- and in fact the whole of trailblazer looks good (hadn't seen this before now -- seems worth exploring). Have you ever used cells in production? trailblazer?

Just while I'm reflecting on this, Cells looks good for plugin authors. Seems like (possibly) a high bar for main-app users/customizers. But that's ok. What we need for main-app users is a gradient of customization: 
 1. drop in  as is 
 2. customize with initializer / layout
 3. override using view overrides / class overrides  (guided by readme docs)
 4. more complex / quasi plugin use (guided by some as yet unwritten plugin docs)

glebm:
I haven't used Cells before nor Trailblazers.

So I read this Cells tutorial: https://www.sitepoint.com/introduction-to-cells-a-better-view-layer-for-rails/ and https://www.sitepoint.com/cells-a-deeper-look-into-dependency-injection-and-testing/ and it seems like there is a fair bit of magic in Cells as well. Unsure what the multi-get caching story is (which we currently benefit from in Rails 5). It is also not quite clear how hooks would be implemented.

Perhaps for v1 we could provide a zero-magic API like this:

```ruby
# Thredded code
class Thredded::ViewHooks
  def initialize(view_context)
    @view_context = view_context
  end

  module Default
    def post_form_content_area(&block)
      # no-op hook by default
      @view_context.capture(&block)
    end
  end
  include Default
  
  def self.configure(modules)
     # called once at setup
     modules.each(include module)
  end
end

# Expose a new configuration option of an array of view hook modules
Thredded.view_hooks = [Thredded::ViewHooks::Default]
```

```ruby
module Thredded::ApplicationHelpers
  def view_hooks
    @view_hooks ||= Thredded::ViewHooks.new(self)
  end
end
```

```ruby
# Thredded view
[...]

PR #455, opened 2016-11-04
A draft implementation of view hooks.

glebm:
I wanted to have full IDE support but at the same time provide a nice API for plugin authors (as per the discussion in #357).

So it's not very DRY. @timdiggins @jayroh What do you think?

Usage:

```ruby
Thredded.view_hooks.post_form.content_text_area.config.before do
  'Hi'
end
```

The view hooks registration is done by plugins directly (as in above), so the order cannot be controller by the user. The plugins can however access the array directly, e.g.:

```ruby
Thredded.view_hooks.post_form.content_text_area.config.before.unshift(-> { 'hi' })
```

The plugins may alternatively expose a method to register its view hooks, leaving the plugins order up to the user. I'm not sure whether this is the right approach. 

Refs #357. ~~Still needs tests and docs.~~

### DIFF
--- a/app/views/thredded/posts_common/form/_content_field.html.erb
+++ b/app/views/thredded/posts_common/form/_content_field.html.erb
@@ -1,6 +1,6 @@
 <li>
   <%= form.label :content, content_label %>
-  <%= view_hooks.post_form.content_text_area do %>
+  <%= view_hooks.post_form.content_text_area.render do %>
       <%= render 'thredded/posts_common/form/before_content', form: form %>
       <%= form.text_area :content, {rows: 5, required: true} %>
       <%= render 'thredded/posts_common/form/after_content', form: form %>
--- a/app/views/thredded/posts_common/form/_content_field.html.erb
+++ b/app/views/thredded/posts_common/form/_content_field.html.erb
@@ -1,6 +1,8 @@
 <li>
   <%= form.label :content, content_label %>
-  <%= render 'thredded/posts_common/form/before_content', form: form %>
-  <%= form.text_area :content, { rows: 5, required: true }  %>
-  <%= render 'thredded/posts_common/form/after_content', form: form %>
+  <%= view_hooks.post_form.content_text_area do %>
+      <%= render 'thredded/posts_common/form/before_content', form: form %>
+      <%= form.text_area :content, {rows: 5, required: true} %>
+      <%= render 'thredded/posts_common/form/after_content', form: form %>
+  <% end %>
 </li>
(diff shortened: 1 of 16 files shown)


app/views/thredded/posts_common/form/_content_field.html.erb
```diff
@@ -1,6 +1,6 @@
 <li>
   <%= form.label :content, content_label %>
-  <%= render 'thredded/posts_common/form/before_content', form: form %>
-  <%= form.text_area :content, { rows: 5, required: true }  %>
-  <%= render 'thredded/posts_common/form/after_content', form: form %>
+  <%= view_hooks.post_form.content_text_area do %>
```
timdiggins: Can we have a transitional moment (maybe only during the preparation of the release version) where we have both before_content/after_content and the view hooks? Makes migration of my main_apps easier.

timdiggins:
Will need to give it a spin in practice before I can see what it's like for a plugin author. It seems straightforward enough. 

ideally the hooks order would at least be deterministic -- I think it would be a shame if it was not somehow predictable. 

Maybe what we should do is start playing with creating a couple of plugins and see how it might work.

app/views/thredded/posts_common/form/_content_field.html.erb
glebm: Done

glebm:
I've updated the docs. I think the order problem may be solvable by instructing plugin authors to expose a `MyThreddedPlugin.register` method instead of registering hooks on load, so that the user can control the order.

glebm:
I've changed the implementation / API slightly. It now uses a thread-local, but the big win is that "Find usages" now works for view hooks, finding both config and render calls.

If this looks good, I think we can merge.

jayroh:
How could this handle something like ... pulling in partials? Or larger complex blocks of html? Or interpolating data provided from controllers into views?

I'm trying to think of the base rails view patterns I would see in the rails stack.

glebm:
The hooks are all run in the context of the view, so they can access everything the view can access, including controller instance variables, the `render` method, etc.

I've just realised that I need to check if it can access locals passed into the partial calling the `hook.render`.

glebm:
Looks like we *can't* access `local_assigns` from these blocks, so we'll need to pass those in.

glebm:
Locals are now passed explicitly. PTAL.

lib/generators/thredded/install/templates/initializer.rb
```diff
@@ -99,3 +99,14 @@
 #     end
 #   end
 # end
+
+# ==> View hooks
+#
+# Customize the UI before/after/replacing individual components.
+# See the full list of view hooks by running:
+#
+#     $ grep view_hooks "$(bundle show thredded)"/**/*.html.erb.
```
timdiggins: nit: only works on zsh-like shells (** doesn't work on bash for example).
Alternative (or just state "on zsh:"):

```grep view_hooks -R --include *.html.erb "$(bundle show thredded)"```

timdiggins:
I'm keen to try this out. And ready for a merge. I like the docs, and the low maintenance list method using grep.

PR #458, opened 2016-11-06
Change Markdown parser to Kramdown

glebm:
github-markdown is not maintained nor supported


glebm:
Kramdown also supports code and math highlighting, so this opens doors to plugins.

I've just pushed a code highlighting plugin that relies on this PR: https://github.com/thredded/thredded-markdown_coderay (still needs docs and tests but it works!)
thredded-markdown_coderay

kivikakk:
I might suggest that http://commonmark.org/ might be a better target than `github-markdown`? (I say this as someone who works on both!) There's a gem at https://github.com/gjtorikian/commonmarker which should drop about as easily as Kramdown or anything else.

jayroh:
@glebm love the plugin, so I'm 👍 for this PR.

@kivikakk I wasn't aware of that implementation so it's certainly something to give a look at. Thanks for raising that up :)

jayroh:
Code looks good, Gleb. 👍 for squash and merge.

glebm:
@kivikakk The commomarker gem looks great. As this Kramdown integration is already working, I'm going to merge this for now, but going forward I think we should add commonmarker and make it the default, while *maybe* allowing kramdown as an option since kramdown is pure ruby.

Issue #371, opened 2016-08-10
remove any mention of incoming email feature


jayroh:
Last I looked that was ripped out completely, but it looks like the config generator still contains mentions of it.

PR #466, opened 2016-11-07
Remove config / code regarding incoming mail

jayroh:
Considering there is no infrastructure in the app to handle incoming
email the following code is, for all intents and purposes, noise.
"Bit-rot" as a term comes to mind. That being said - the gem can
function without the presence of this removed code.

Addresses issue #371


jayroh:
Only question I would have around this would be whether or not to guard against someone's existing app setting this attribute and stopping the app from booting.

What do you think about a deprecation message? Or are we moving forward with the operating assumption that this is pre-1.0 software.

glebm:
I think it's fine to just blow up since the fix is easy and we'll mention it in the release notes

PR #468, opened 2016-11-08
Attempt to fix onPageLoad

### LINKED ISSUE #464: onPageLoad callbacks may not always get called (?)
Some components are adding their PageLoad callbacks within a jquery callback. 

But this is (at least sometimes) happening after the PageLoad callbacks have been fired.

so maybe we need to check for this and fire them immediately if  

```javascript
if (onPageLoadFiredOnce && window.Thredded.DOMContentLoadedFired) {
  callbacks();
}
```

ref: https://github.com/thredded/thredded/commit/8e7e1d399a35006b03630ff18e484fd66620f9f9#commitcomment-19721263

I've possibly seen this happening in a branch, but it's all a bit async and non-deterministic

glebm:
Remove the use of jQuery.fn.ready, as it falls back to window.load if
DOMContentLoaded had already fired, but window.load fires much later.

This means we no longer have to distinguish between Turbolinks Classic
with vs without jQuery.Turbolinks.

@timdiggins Does this resolve #464?

### DIFF
--- a/app/assets/javascripts/thredded/core/on_page_load.es6
+++ b/app/assets/javascripts/thredded/core/on_page_load.es6
@@ -1,7 +1,6 @@
-(($) => {
+(() => {
   const isTurbolinks = 'Turbolinks' in window && window.Turbolinks.supported;
   const isTurbolinks5 = isTurbolinks && 'clearCache' in window.Turbolinks;
-  const isjQueryTurbolinks = 'turbo' in $.fn;
 
   let onPageLoadFiredOnce = false;
   const pageLoadCallbacks = [];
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/thredded/core/on_page_load.es6
```diff
@@ -19,31 +18,29 @@
...
-
   if (isTurbolinks5) {
     document.addEventListener('turbolinks:load', () => {
       triggerOnPageLoad();
     });
   } else {
-    // Turbolinks Classic with no jQuery.Turbolinks:
-    $(() => {
-      triggerOnPageLoad();
+    // Turbolinks Classic (with or without jQuery.Turbolinks):
```
timdiggins: this comment is wrong  -- this stanza represents Turbolinks Classic (with or without jQuery.Turbollinks) **or no Turbolinks**

glebm: Fixed

PR #471, opened 2016-11-12
Remove the messageboards.closed column

### LINKED ISSUE #407: remove default_scope from messageboard
We should remove the default_scope from messageboards and just add a scope `.open` or `.unclosed` to messageboards whenever we use them. the order should be handled now by #404 

_(the more I use default scopes the less I like them, as initially attractive as they are)_

glebm:
Fixes #407


glebm:
Merging this to avoid holding back the v0.8.0 release.

Issue #428, opened 2016-10-20
Encourage people to share they're using thredded


jayroh:
I find this is something the project has had very little luck with - and that's nudging people towards sharing that they're using thredded for their websites. I have a feeling there are more than several production deployments out there but ... we hear nothing about them.

One of the first things people do when they seed the database, or create a new messageboard, is see the first sample topic. What about updating that copy to link to the twitter account, link to the thredded team email ("let us know how you're using thredded"), maybe link to the issue tracker, etc.

What do you think?

glebm:
I've added some text there that can be improved in the future. Getting ready to release v0.8.0!

PR #473, opened 2016-11-13
Encourage people to share they're using thredded

glebm:
Adds links to Thredded GitHub, Gitter, and Twitter in the copy of the post generated when an admin creates a messageboards.

Refs #428


glebm:
Merging this to avoid holding back the v0.8.0 release. The copy can be improved later on.

PR #482, opened 2016-11-19
s/find_by_slug/friendly_find!/ and use everywhere

glebm:
This is to avoid confusion with the Rails dynamic finder methods

### DIFF
--- a/app/controllers/thredded/application_controller.rb
+++ b/app/controllers/thredded/application_controller.rb
@@ -76,7 +76,7 @@ module Thredded
     end
 
     def update_user_activity
-      return if messageboard.nil? || !signed_in?
+      return if !messageboard_or_nil || !signed_in?
 
       Thredded::ActivityUpdaterJob.perform_later(
         thredded_current_user.id,
@@ -88,10 +88,17 @@ module Thredded
       thredded_current_user
     end
 
+    # Returns the `@messageboard` instance variable.
+    # If `@messageboard` is not set, it first sets it to the topic with the slug or ID given by
+    # `params[:messageboard_id]`.
+    #
+    # @return [Thredded::Messageboard]
+    # @raise [Thredded::Errors::MessageboardNotFound] if the topic with the given slug does not exist.
     def messageboard
-      @messageboard ||= params[:messageboard_id].presence && Messageboard.friendly.find(params[:messageboard_id])
-    rescue ActiveRecord::RecordNotFound
-      raise Thredded::Errors::MessageboardNotFound
+      @messageboard ||= begin
+        fail Thredded::Errors::MessageboardNotFound unless params[:messageboard_id].presence
+        Messageboard.friendly_find!(params[:messageboard_id])
+      end
     end
 
     def messageboard_or_nil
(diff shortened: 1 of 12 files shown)


app/controllers/thredded/application_controller.rb
```diff
@@ -88,10 +88,17 @@ def pundit_user
       thredded_current_user
     end
 
+    # Returns the `@messageboard` instance variable.
+    # If `@messageboard` is not set, it first sets it to the topic with the slug or ID given by
```
timdiggins: docs seem odd here -- "topic" ?

app/controllers/thredded/application_controller.rb
```diff
@@ -88,10 +88,17 @@ def pundit_user
...
+    # `params[:messageboard_id]`.
+    #
+    # @return [Thredded::Messageboard]
+    # @raise [Thredded::Errors::MessageboardNotFound] if the topic with the given slug does not exist.
```
timdiggins: ... "messageboard" ... ?

Also why **are** we using MessageboardNotFound rather than the more normal (but less specific) ActiveRecord::NotFound -- can see it's a style across different models, but wondering what it gives us

app/controllers/thredded/application_controller.rb
```diff
@@ -88,10 +88,17 @@ def pundit_user
...
     def messageboard
-      @messageboard ||= params[:messageboard_id].presence && Messageboard.friendly.find(params[:messageboard_id])
-    rescue ActiveRecord::RecordNotFound
-      raise Thredded::Errors::MessageboardNotFound
+      @messageboard ||= begin
+        fail Thredded::Errors::MessageboardNotFound unless params[:messageboard_id].presence
```
timdiggins: Bit of a nit pick, but I feel a bit uncomfortable with this style (hard to read) (previous version was odd too IMO). 

Also -- not sure why, but the behaviour seems to change in this PR. Previously it was **not** raising MessageboardNotFound when messageboard_id empty, now it is...

Instead maybe extract the check to a `required_messageboard_id` or `messageboard_id!` (?!) method  which does `params[:messageboard_id].presence || fail Thredded::Errors::MessageboardNotFound` Will make this method super simple

app/controllers/thredded/application_controller.rb
glebm: Oops, copy-paste accident. Fixed.

app/controllers/thredded/application_controller.rb
glebm: We're a custom error so that main app can handle it in a custom way. Currently the default behaviour is to render `exception.message` for these custom errors (https://github.com/thredded/thredded/blob/ffebb25398fdf4de95b1cf14e70421c16f58d544/app/controllers/thredded/application_controller.rb#L17-L24). Maybe there is a better way to do this?

timdiggins:
@glebm regarding the errors: Feel like it would be less surprising if the errors extended (were specializations) of ActiveRecord::NotFound. However that's not part of this PR which is all good. Could look at that later if we care enough.

app/controllers/thredded/application_controller.rb
jayroh: > Also why are we using MessageboardNotFound rather than the more normal (but less specific) ActiveRecord::NotFound -- can see it's a style across different models, but wondering what it gives us

The generic exception is fine but there was a time when I wrote this that I wanted a specific error message to show up with regards to a *messageboard* not existing.

Issue #478, opened 2016-11-18
Not autoloading / reloading is a pain


timdiggins:
To fix rails admin (https://github.com/thredded/thredded/issues/280) 🎉 , we've stopped auto_loading lib (https://github.com/thredded/thredded/commit/181dfa1a9ef3e62018d6385fa7f6ad013ad8085c) 

I worry that having a mixture of autoloaded and non-autoloaded paths is likely to cause us problems -- a proliferation of explicit requires and require_dependency (for non-autoload and autoload respectively -- but easy to get this wrong when it's all Thredded code) which are then confusing.
We already have loads of explicit requires in the spec files. 

Also it's just a hassle to not auto reload.

But we could just move most of our /lib code to an /app/lib directory (autoloaded) and only leave the code in /lib that we are either planning to move out to a plugin anyway (a kind of halfway house), or is a generator or template or rake/thor task or the engine itself

timdiggins:
@glebm thoughts on this?

glebm:
Here are some thoughts per file:
1. `content_formatter.rb` - I think this must stay in lib, because its class attributes are modified in the initializers outside of `config.to_prepare` and these modification would be lost if it is reloaded.
2. `html_pipeline/` - can stay in lib, because it's a dependency of `content_formatter.rb` that isn't reloaded.
3. `view_hooks/` - can stay in lib or move to `app/view_hooks`.
4. `database_seeder.rb` - Must stay in lib, because we do not want to eager load it in production.
5. `errors.rb` - Must stay in lib because other files in lib depend on it.
6. `search_parser.rb`, `topics_search.rb` - These are actually form models / scopes, so should go into `app/`. But we're not changing these files much at all, so can stay in lib for now.

So, out of the current files in lib, only `view_hooks/`, `search_parser.rb`, and `topics_search.rb` can be moved to `app/`. The notifiers code should go into `app/notifiers/thredded`.

timdiggins:
SGTM. @glebm -- will get the notifier stuff done in #442 and then we can move the other ones as we necessary (given we have a plan, we don't have to do it all at once)

My feeling is we should try to minimize `require_dependency` (and require, though typically these are needed)-- they can proliferate without end.  To that end we should remove them as we move files.

My understanding for the distinction is:

* always use require for non-autoloaded files (and thus non-reloadable modules)
[...]

PR #483, opened 2016-11-19
Do not `require` from autoloaded files

glebm:
Following the findings in #478:
https://github.com/thredded/thredded/issues/478#issuecomment-261742503


timdiggins:
@glebm I think we've agreed this one is not the right path

PR #484, opened 2016-11-19
Do not use `require_dependency`

glebm:
`require_dependency` is unnecessary, as per the findings in
https://github.com/thredded/thredded/issues/478#issuecomment-261743820

However, when not using `require_dependency`, the absolute constant name
must be used (e.g. `Thredded::Post` instead of `Post`) whenever
there is a chance that the constant has not been loaded yet.
This is because if the `Post` constant is already loaded, autoload will
not get triggered for `Thredded::Post` when referencing just `Post`.

### DIFF
--- a/app/commands/thredded/autofollow_mentioned_users.rb
+++ b/app/commands/thredded/autofollow_mentioned_users.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require_dependency 'thredded/at_notification_extractor'
 module Thredded
   class AutofollowMentionedUsers
     def initialize(post)
--- a/app/controllers/thredded/moderation_controller.rb
+++ b/app/controllers/thredded/moderation_controller.rb
@@ -1,13 +1,11 @@
 # frozen_string_literal: true
-require_dependency 'thredded/moderate_post'
-require_dependency 'thredded/posts_page_view'
 module Thredded
   class ModerationController < Thredded::ApplicationController
     before_action :thredded_require_login!
     before_action :load_moderatable_messageboards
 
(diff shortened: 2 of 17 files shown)


timdiggins:
@glebm I think being explicit about nesting is a good idea with autoloading, even though I think your explanation may be very slightly off. There's a good post at http://urbanautomaton.com/blog/2013/08/27/rails-autoloading-hell/ that is still I think accurate with Rails 5. It would explain (I think) that the **second time** these are autoloaded it is the problem (not the first) -- because the first time, it autoloads them with the wrong dependency... (e.g. in https://github.com/thredded/thredded/pull/484/files#diff-c1ff35195c31a5d98e69b95d02ec8590 in the old way PostsPageView gets autoloaded as Thredded::ModerationController::PostsPageView first time, and second time is just missing).

All that said, I think this PR is totally correct. Would like to try it out one time later today, but just merge it if you need it.

glebm:
Both of our explanations are right, unfortunately :smile: 

If `::PostsPageView` or `::Thredded::ModerationController::PostsPageView` is defined at the time of loading `moderation_controller.rb`, then `const_missing` will not be triggered. :boom:

If none of `::Thredded::ModerationController::PostsPageView`, `Thredded::PostsPageView`, `::PostPageView` are defined, Rails will recursively try to find `thredded/moderation_controller/posts_page_view.rb`, then `thredded/posts_page_view.rb`, then `posts_page_view.rb` in all `autoload_paths`. if `thredded/moderation_controller/posts_page_view.rb` exists, :boom:.

PR #485, opened 2016-11-20
Disable buttons in forms on submit

glebm:
This prevents accidental double-submits.

Uses the jquery-ujs `data-disable-with`.

### DIFF
--- a/app/views/thredded/messageboard_groups/new.html.erb
+++ b/app/views/thredded/messageboard_groups/new.html.erb
@@ -16,7 +16,11 @@
         <%= f.text_field :name, required: true %>
       </li>
 
-      <li><%= f.submit t('thredded.messageboard_group.create'), class: 'thredded--form--submit' %></li>
+      <li>
+        <%= f.submit t('thredded.messageboard_group.create'),
+                     class: 'thredded--form--submit',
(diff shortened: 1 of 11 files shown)


jayroh:
Man, this is way overdue. I have just one comment but other than that, this is ready to go.

glebm:
@jayroh What is your one comment? :smile:

jayroh:
It didn't come through? Jesus Christ githubs new commenting ux is
maddening.

I had mentioned that the local var in one of the views could/should be
moved to where it's assigned as a value for a data attribute.

Responding via email now so I don't have it in front of me (at the movies
about to see fantastic beasts 😄)

glebm:
@jayroh I've moved the variable closer but it's a bit too long to inline it (especially since I've now replaced the `input` with a `<button>`).

Hope you've enjoyed the movie! We have high hopes for it but will be waiting for a Blu-Ray! :smile:

jayroh:
@glebm Fun movie! It was sorely needed considering what the world feels like these past few weeks. If you need something in the next month or so to distract you - go see it. And *Arrival* is pretty great too 😉

glebm:
@jayroh I know what you mean about the past few weeks. Thanks!

PR #494, opened 2016-11-24
Add global settings for auto-following

### LINKED ISSUE #372: Disable Automatic Subscription When Posting
Is it possible to disable the automatic subscription (auto-follow) when posting to a topic? I would prefer users to have to opt-in to following a topic via email.

glebm:
* Whether the user should get subscribed to a new topic they've created.
* Whether the user should get subscribed to a topic after posting in it.

Also, document (in the initializer) how to change the global default for
auto-follow on mention.

Resolves #372

### DIFF
--- a/app/models/concerns/thredded/post_common.rb
+++ b/app/models/concerns/thredded/post_common.rb
@@ -32,7 +32,7 @@ module Thredded
       Thredded::ContentFormatter.new(view_context, users_provider: users_provider).format_content(content)
     end
 
-    def first_in_topic?
+    def first_post_in_topic?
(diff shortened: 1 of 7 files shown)


jayroh:
Only one small nitpick otherwise these looks great: `first_in_topic` took me a second to determine what exactly was first. First POST or first REPLY. Why? I've frequented too many communities where "FIRST!" means first *reply*.

So - what to you think about being a little more specific and calling that method `first_post_in_topic` or `original_post_in_topic`?

jayroh:
🎉

PR #500, opened 2016-11-29
The banner and navigation role not needed

realCh4mp:


glebm:
Yep, as per the [HTML spec says](https://www.w3.org/TR/html5/dom.html#aria-usage-note) they are the default roles for these element and it is not recommended to add them explicitly:
> In the majority of cases setting an ARIA role and/or aria-* attribute that matches the default implicit ARIA semantics is unnecessary and not recommended as these properties are already set by the browser.

PR #501, opened 2016-12-04
Add pagination on top of the page

glebm:
We show quite a few posts per page, and it's not always clear which page you're on.

This adds the pagination bar to the `topics#show` and `private_topics#show` pages by default. View hooks are provided so it's easy to disable it.

---

thredded-pagination-top


glebm:
@timdiggins @jayroh Is this OK to merge?

glebm:
@jayroh @timdiggins I'm going to merge this so that we can focus on v0.10.0 without merge conflicts. :ship:  I'll address your feedback if any in a follow-up PR.

PR #502, opened 2016-12-09
Update thredded.scss

realCh4mp:
jumping scrollbar issue fix.
disabled by default but it's a good practice to enable when a website / forum is too short = no scrollbar.

### DIFF
--- a/app/assets/stylesheets/thredded.scss
+++ b/app/assets/stylesheets/thredded.scss
@@ -1,3 +1,7 @@
++/* prevents from website jumping
+ +html {overflow-y: scroll}
+ +*/
+
 @import "thredded/dependencies";
 @import "thredded/thredded";
 


app/assets/stylesheets/thredded.scss
```diff
@@ -1,3 +1,7 @@
++/* prevents from website jumping
+ +html {overflow-y: scroll}
+ +*/
```
jayroh: Is the fix supposed to be commented out, @realCh4mp?

glebm:
1. Thredded shouldn't be changing global styles (such as the `html` element).
2. Commented out code here is not useful as the user cannot change this file.

realCh4mp:
yeah maybe it should not be a part of thredded.

but it fixes the jumping scrollbar issue so I thought we could keep it as a hint there.

PR #508, opened 2016-12-24
Fix NoMethodError in Rails 5.0.1

unn4m3d:
Stacktrace is [here](https://gist.github.com/unn4m3d/628f1681ce4b1a4c7e6b863de4ce53a3)

### DIFF
--- a/app/view_models/thredded/posts_page_view.rb
+++ b/app/view_models/thredded/posts_page_view.rb
@@ -4,6 +4,7 @@ module Thredded
   class PostsPageView
     delegate :to_ary,
              :present?,
+             :to_a,
              to: :@post_views
     delegate :total_pages,
              :current_page,


jayroh:
Ha - weird. Good catch though!

@unn4m3d could you update the commit message with either (part of) the contents of the gist you linked to in your comment? Or move your comment's contents into the body of the commit message?

Giving some context in the commit would be super helpful

glebm:
Relevant Rails commit:
https://github.com/rails/rails/commit/87899cfcf0ada140764acf4393eb862698449f54

The Rails changelog entry is under 5.0.1.rc1: https://github.com/rails/rails/blob/v5.0.1/actionview/CHANGELOG.md#rails-501rc1-december-01-2016

glebm:
The Rails PR that introduced this backwards-incompatible change: https://github.com/rails/rails/pull/25912.

PR #514, opened 2017-01-16
Closes #512

### LINKED ISSUE #512: Private message e-mail notification
Hi! During testing e-mail notification for new private message I noticed strange bug - the message which comes by the e-mail is always the first message of the whole conversation. No matter who sends message, the notification always contains the first message of conversation.

soanvig:
Fix private_topic_mailer notification
Notification now contains newest post, instead of the first in whole conversation.


soanvig:
It's my first PR. I am not sure if I've done it correctly. I wanted to merge it into new branch on thredded, but it seems I cannot create one.

PR #515, opened 2017-01-16
Fix/private topic mailer, closes #512

### LINKED ISSUE #512: Private message e-mail notification

soanvig:
- Fix private_topic_mailer notification
- Notification now contains newest post, instead of the first in whole conversation.
- Filtering (sanitizing) content of HTML e-mail is unneccessary - content is filtered automatically (checked on app using Thredded), so all HTML will be escaped. I feel, that any other sanitizing (removing Markdown signs, or even the HTML) can have negative impact on displaying message in e-mail's inbox, resulting in being unable to read.

Closes issue #512 
I have used `.last` instead of `.order_newest_first.first`. It seems to work.


jayroh:
Correct me if I'm wrong but has the behavior for `first` and/or `last` changed through the last couple rails versions? I seem to remember there being a change over the last few years.

To that end I might recommend ordering the collection specifically, with SQL, limiting it to 1 record, and getting that one record. `posts.last` would return all the records and then pop the last ruby object, no?

soanvig:
Indeed, `.last` fetches `SELECT * FROM xxx` then pops the last item.

If we should use LIMIT?
http://stackoverflow.com/questions/34488/does-limiting-a-query-to-one-record-improve-performance

It seems, that yes, it will affect performance, as expected (if "last" message will be, for example, in the middle of the table).
I was basing on what was before, so I just changed ".first" to ".last".

Anyway, it can be done, but there - if we are using SQL query it definitely shouldn't be done in viewer.

app/views/thredded/private_topic_mailer/message_notification.html.erb
```diff
@@ -1,7 +1,5 @@
-<blockquote><%= @topic.posts.first.filtered_content(self) %></blockquote>
-<%                           # as this is only sent on first post, doesn't need to define order %>
+<blockquote><%= @topic.posts.last.content %></blockquote>
```
glebm: The post should just be passed directly from the place that creates this notification, i.e.:
https://github.com/thredded/thredded/blob/019281c256f0c8b23e737d58863af90a04d16260/app/notifiers/thredded/email_notifier.rb#L21

soanvig: Then we can apply @jayroh idea. I will do that :)

soanvig:
I sit to write that, and I came to a conclusion - *how* we should write SQL query to limit results?

We want to find the last (newest) post. Okey, so then: how do we know which one is the last? To be ensured, that selected post is last, we need to download all of posts and check.

Therefore it seems that limiting result to 1 will give no performance boost, because we need to download all posts anyway.
@jayroh

glebm:
@soanvig We shouldn't have to *find* anything. Simply pass the post ID to the mailer from https://github.com/thredded/thredded/blob/019281c256f0c8b23e737d58863af90a04d16260/app/notifiers/thredded/email_notifier.rb#L21.

soanvig:
Oh yes... I've forgot, that we the whole event is triggered after posting... Damn myself.

soanvig:
Okey, I did the push.
I've also changed HTML structure (according to https://developer.mozilla.org/samples/html/figure.html) to include author of message. Author is identified by e-mail, because this is the field, which should be always present.

app/mailers/thredded/private_topic_mailer.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 module Thredded
   class PrivateTopicMailer < Thredded::BaseMailer
-    def message_notification(private_topic_id, emails)
+    def message_notification(private_topic_id, post, emails)
```
glebm: Let's pass `post_id` here and look it up like this:

```ruby
@post = find_record Thredded::PrivatePost, post_id
```

This avoids serializing the entire record into job arguments and also ensures the latest version of the post is used when job is executed.

app/views/thredded/private_topic_mailer/message_notification.html.erb
```diff
@@ -1,7 +1,10 @@
-<blockquote><%= @topic.posts.first.filtered_content(self) %></blockquote>
-<%                           # as this is only sent on first post, doesn't need to define order %>
+<figure>
+    <figcaption><%= @post.user.email %> said:</figcaption>
+    <blockquote>
+        <%= @post.content %>
```
glebm: The `content` attribute is the raw markup. Instead, this should be:

```ruby
@post.filtered_content(self)
```

soanvig:
Let me work a while on mockup. I'm trying to solve how it works :D

app/views/thredded/private_topic_mailer/message_notification.text.erb
```diff
@@ -1,5 +1,5 @@
 <%= @post.user.email %> said:
-<%= @post.content %>
+<%= @post.filtered_content(self) %>
```
glebm: Here I think the raw markup is more appropriate, as this is the plain text version

glebm:
Almost there just need to address the comment above and update the preview as well!

soanvig:
I can't get Preview working with f7777dfe895c7a9bbdc7fc9ac71a79b04c650cce, because of
`Couldn't find Thredded::PrivatePost with 'id'=407>`
which is a bit weird, because Topic works, but Post nope.

soanvig:
@glebm, I will learn a lot during working with Thredded. Code tested locally (had to manually create `log/test.sqlite3.log`) and working.

PR #516, opened 2017-01-22
Preview posts and topics

### LINKED ISSUE #237: Preview post contents
Per github, waffle.io, discourse, et al ... would be nice to be able to preview the output of a post similarly.

Have something like `PreviewController#show` return some json with the generated html, maybe?

glebm:
This adds preview to Thredded.

If JavaScript is enabled, the UI is preview-as-you-type and looks like this:

preview screenshot

If JavaScript is disabled, instead there is a preview button that opens the preview in a new tab when clicked.

Resolves #237

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -6,8 +6,6 @@ engines:
       languages:
       - ruby
       - javascript
-  fixme:
-    enabled: true
   rubocop:
     enabled: true
 ratings:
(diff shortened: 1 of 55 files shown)


glebm:
@jayroh @timdiggins This should be good to go. :smile:

glebm:
@jayroh @timdiggins I'm going to merge this so that we can focus on v0.10.0 without merge conflicts. :ship:  I'll address your feedback if any in a follow-up PR.

Issue #488, opened 2016-11-21
Auto-follow new topics in specific messageboards?


zapnap:
I'm looking for a way to have all my users automatically follow (and be notified of) new topics in a given messageboard, ala Google Groups. This would need to be something they could opt out of, of course. Looks like this would need to be added as a auto_follow preference in thredded_user_messageboard_preferences and then exposed as a preference in preferences/edit, but thought I'd toss it out there for discussion before taking a stab at implementation. Any thoughts?

I could also see use cases for a similar feature at the meta level like "auto-follow new messageboards"...

glebm:
I can definitely see use cases for this, like a private company messageboard.
IMO the most-common use case would be `false` by default though (can be changed with `change_column_default` for a given installation though).

zapnap:
Agree 100%. I'll work up a PR.

zapnap:
@glebm how do you prefer db migrations in a PR? Just dumped into db/migrate for ultimate merging into an upgrade_migrations script at release?

glebm:
@zapnap Well, we usually just modify the "create" and "upgrade" migrations directly in the PR (although I'm not sure if it's the best approach).

glebm:
We may need to release v0.9.0 soon because it fixes a lot of minor bugs, so let's punt this to v0.10.0.

zapnap:
👍 I'll probably try to make some time to take a look this weekend. v0.10 seems like a good target. Thanks @glebm

glebm:
@zapnap Looks good!

PR #517, opened 2017-01-24
Auto-follow new topics

zapnap:
See #488 - by changing the column default specified in the README, users will (by default) auto-follow topics in all newly created messageboards. You can also togggle individual messageboard auto-follows (like an 'important notices' board) for specific message boards by running a data migration targeting the ThreddedUserMessageboardPreference model. Users can turn this functionality off at the global level or for particular boards.


thredded.gemspec
```diff
@@ -67,7 +67,6 @@ Thredded works with SQLite, MySQL (v5.6.4+), and PostgreSQL. See the demo at htt
   # dummy app dependencies
   s.add_development_dependency 'rails-i18n'
   s.add_development_dependency 'http_accept_language'
-  s.add_development_dependency 'mysql2'
```
glebm: We need this for testing on mysql2.

db/migrate/20161121231919_follow_topics_on_create.rb
```diff
@@ -0,0 +1,6 @@
+class FollowTopicsOnCreate < ActiveRecord::Migration
```
glebm: Let's move this to a new migration file in `db/upgrade_migrations/#######_upgrade_v0_9_to_v0_10.rb`, and update the [create_thredded](https://github.com/thredded/thredded/blob/master/db/migrate/20160329231848_create_thredded.rb) migration.

thredded.gemspec
zapnap: My bad -- didn't intend to commit this

db/migrate/20161121231919_follow_topics_on_create.rb
zapnap: Done!

db/migrate/20160329231848_create_thredded.rb
```diff
@@ -161,6 +161,7 @@ def change
     create_table :thredded_user_preferences do |t|
       t.references :user, null: false
       t.boolean :follow_topics_on_mention, default: true, null: false
+      t.boolean :auto_follow_topics, default: true, null: false
```
glebm: The default here and below should probably be `false`.

db/migrate/20170125033319_upgrade_v0_9_to_v0_10.rb
```diff
@@ -0,0 +1,6 @@
+class UpgradeV09ToV010 < ActiveRecord::Migration
```
glebm: This file should be in `db/upgrade_migrations`.

db/migrate/20160329231848_create_thredded.rb
zapnap: Good catch -- got it

db/migrate/20170125033319_upgrade_v0_9_to_v0_10.rb
zapnap: Moved.

app/commands/thredded/autofollow_users.rb
```diff
@@ -21,6 +26,10 @@ def autofollowers
 
     attr_reader :post
 
+    def include_auto_followers
+      post.messageboard.user_messageboard_preferences.auto_followers.map(&:user)
```
glebm: `UserPreference.auto_followers` should also be included here.

app/commands/thredded/autofollow_users.rb
```diff
@@ -1,18 +1,23 @@
...
 
+    def mentioned_users
+      @mentioned_users ||= Thredded::AtNotificationExtractor.new(post).run
+    end
+
     def autofollowers
-      autofollowers = Thredded::AtNotificationExtractor.new(post).run
+      autofollowers = (include_auto_followers + mentioned_users).uniq
       autofollowers.delete(post.user)
       exclude_those_opting_out_of_at_notifications autofollowers
```
glebm: The users that don't have read permissions for the topic should also be excluded (otherwise we may leak the contents in the notification email).

app/commands/thredded/autofollow_users.rb
```diff
@@ -1,18 +1,23 @@
...
-  class AutofollowMentionedUsers
+  class AutofollowUsers
     def initialize(post)
       @post = post
     end
 
     def run
       autofollowers.each do |user|
-        Thredded::UserTopicFollow.create_unless_exists(user.id, post.postable_id, :mentioned)
+        reason = mentioned_users.include?(user) ? :mentioned : :auto
```
glebm: Perhaps `autofollowers` should return a Hash of `user_id` to `reason` so we don't have to do this check here?

app/models/thredded/messageboard.rb
```diff
@@ -27,6 +27,14 @@ def ensure_position
       self.position ||= (created_at || Time.zone.now).to_i
     end
 
+    after_save :set_autofollow, on: :create
```
glebm: Setting global auto-follow preference to true will not update it for the existing messageboards with this approach.

How about we just check both the global and per-messageboards preferences when calculating the autofollowers in `AutofollowUsers`?

app/commands/thredded/autofollow_users.rb
glebm: The notification job now respects permissions, so this one is done.

zapnap:
How do you want to proceed here wrt the new work in #529 ?

glebm:
Rebase, address comments, and we'll merge this!

glebm:
Released v0.10.0 and moved this to v0.11.0 milestone, so the migration will need to be renamed.

zapnap:
Okay -- just haven't had a free moment lately to follow up on this.

glebm:
I've updated this in #554

Issue #62, opened 2015-01-26
i18n all the things


jayroh:
Another good point in issue #46 from one mr. @glebm 

> I18n everything. I wrote the [i18n-tasks gem](https://github.com/glebm/i18n-tasks) to make maintaining locale files super-easy.

TheDudeWithTheThing:
Oh wow, we use the i18n-tasks gem at work, highly recommend

glebm:
There is also a number of gems to extract plain strings from views and turn them into keys: https://github.com/glebm/i18n-tasks/wiki/Extract-plain-strings-to-i18n-keys.

glebm:
Current status: if you touch a view, please i18n it.

glebm:
Only a few more to go:

```bash
$ git grep "'[^']*[A-Z][^']*'" app/views
app/views/thredded/post_mailer/post_notification.html.erb:  <%= link_to 'View the conversation here', topic_url(@post.postable) %>.
app/views/thredded/posts/edit.html.erb:<% content_for :thredded_page_title, 'Edit Post' %>
app/views/thredded/private_topic_mailer/message_notification.html.erb:  <%= link_to 'Go here', @topic %> to view the conversation.
app/views/thredded/private_topics/new.html.erb:<% content_for :thredded_page_title, 'Create a New Private Topic' %>
app/views/thredded/shared/_page.html.erb:  document.addEventListener('DOMContentLoaded', function() {
app/views/thredded/theme_previews/show.html.erb:<% content_for :thredded_page_title, 'Theme Preview' %>
app/views/thredded/theme_previews/show.html.erb:    <li><%= link_to 'Theme Preview', theme_preview_path %></li>
app/views/thredded/theme_previews/show.html.erb:               placeholder:  'Start a New Topic' %>
app/views/thredded/theme_previews/show.html.erb:             placeholder:  'Start a New Topic' %>
app/views/thredded/topics/new.html.erb:<% content_for :thredded_page_title, 'Create a New Topic' %>
```

glebm:
Done!

jayroh:
👏👏👏👏👏👏

PR #518, opened 2017-01-24
Add support for polish language

soanvig:
Hi. For my application I wrote polish translation. Here goes localization file.

But what to do with texts which are not localized in i18n? I've overwritten them in my app, but in gem it's impossible to do for now.


glebm:
@soanvig Please add i18n for any text that hard-coded. Our long-term goal is to localize everything (see #62)!

soanvig:
This pull request is closed.
#519 is the correct one.

PR #519, opened 2017-01-24
Add polish localization file

soanvig:
Here goes localization for **polish** language. 

It should work out of the box (alongside with that plugin for jQ for displaying date, I don't remember the name. I had errors with it due to not enough pluralization options, but fixed it).


glebm:
This is great, thank you! :sparkles:

PR #532, opened 2017-02-18
Make a smoke test for otherwise randomly-occurring failure. [WIP]

timdiggins:
There's a randomly occurring failure when smileys are added, when js tests will try to pull in the emoji which doesn't exist (see below).

So far, this PR just makes a nice little smoke test (and tidies up FakeContent a bit) to highlight the problem (and potential future problems with other rich content).

Possible solutions:

* add emojis into dummy app in repo
* normally don't make posts so random / rich content  (and do some web mocking to fix this spec)
* always do some web mocking to grab these particular public/emojis/ requests

### DIFF
--- a/spec/features/thredded/user_views_a_topic_with_super_rich_content_spec.rb
+++ b/spec/features/thredded/user_views_a_topic_with_super_rich_content_spec.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+require 'spec_helper'
+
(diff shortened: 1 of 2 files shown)


glebm:
:+1: Sorry, didn't notice the [WIP] in the title and merged.

> Possible solutions:

Perhaps add just the two emojis that the FakeContent can generate? This way we still test the emoji filter, without bloating the repo.

Issue #362, opened 2016-08-07
Topics in index should link to most recently read page, rather than first page


timdiggins:
Given that we store the page of the most recently read page, shouldn't we be linking to this page in the topics#index ?

glebm:
I think we are. It also works for me on https://thredded.org/thredded/main-board.

timdiggins:
@glebm agreed, have understood where it happens in the view model now. 

I was thinking (in addition to linking to the page as we are doing) we should link to the anchor of the last read post if possible. This anchor should be calculated stored in UserTopicReadState.

timdiggins:
from #533: 

> And I think it would be good to highlight the unread posts when you first view them (when they are transitioning from read-> unread as it were), like we highlight unread topics

@jayroh I know you've had some thoughts about this -- maybe about doing it in Javascript w/scrollspy for example ( I'm less keen on this, even though it is less accurate, seems like it could be nightmarish to support)

jayroh:
@timdiggins Yeah your estimation is right on. This is the "best" way to track all of this but it's no easy task as there's a lot to juggle. On topics#show - the topic, page, and post

Then of course there's topics#index - marking read or un-read on all of those.

JS is the best way to do this and in terms of support, as long as it's well tested it should be relatively manageable.

timdiggins:
I'm in favour of a simple approach as the next step:

* when viewer views (requests the html of) the last page of posts , then mark all these as read  (via controller) the last page of posts (as roughly as present -- though we ignore what page is being read)

* add a marker saying "unread posts below here" in the gap between read and unread posts

* when linking to a topic page, link to: the last post (specific `#post-2312` anchor) if all read, or to the top of the page (no anchor) if none read, or to the "unread posts marker" (`#unread-posts` anchor) if some but not all are read.

timdiggins:
decision on https://github.com/thredded/thredded/issues/535 is relevant to putting a "unread posts below here" in the UI.

glebm:
> I'm in favour of a simple approach as the next step:

+1

> add a marker saying "unread posts below here" in the gap between read and unread posts

#536 is a pre-requisite for adding the marker, so I'll merge it tomorrow if there are no objections.

> when linking to a topic page, link to: the last post (specific #post-2312 anchor) if all read, or to the top of the page (no anchor) if none read, or to the "unread posts marker" (#unread-posts anchor) if some but not all are read.

+1

PR #533, opened 2017-02-20
mark as unread

EllaNancyFay:


glebm:
This adds a "Mark as unread" button for every post like this:

> mark-as-unread


Perhaps we should have just one button under the topic header instead?

E.g.:

> mark-as-unread-suggestion

The button would mark the last post in the topic as unread.

On desktop this might need to use flexbox to look nice.

timdiggins:
@glebm agreed in current Thredded, unless a topic goes onto multiple pages, there's no difference between a topic being totally unread and a topic having only the last post unread. However with a post with multiple pages, and moreover if we implement https://github.com/thredded/thredded/issues/362, then there is a difference. (And I think it would be good to highlight the unread posts when you first view them, like we highlight unread topics)

But maybe the text should say "Mark unread from here" (like gmail does for example

gmail_-_red56_uk
).


That said, if we are going to have "mark as unread" being available on every post, it does need to set the correct `page` of the previous post for the read state -- so that's a todo.

EllaNancyFay:
@glebm @timdiggins - comments taken on board. In order for this to be coherent I feel the best course of action would to implement [#362](https://github.com/thredded/thredded/issues/362). Then progress this PR so:

- if it's the first time the topic is being viewed the user is taken to the top of the page 
- if the user's viewed the topic and not marked a post as unread they're taken to the last post
- if the user's viewed the topic and marked a post as unread they're taken to that post (with some indication in the UI that posts are unread from here)

I'll crack on for now. Give me a shout if there's any further comments

glebm:
I think having a "Mark unread from here" button under every post creates a lot of visual noise for what will be a rarely used feature. We need to consider carefully how to do this while keeping the UI clean.

> with some indication in the UI that posts are unread from here

Are you thinking of something like a separator, e.g.: "New posts since your last visit"?
Or an indicator (e.g. a blue dot) next to unread posts?

EllaNancyFay:
@glebm agreed 'Mark unread from here' is a bit of a mouthful and adds unnecessary noise to the UI. How about popping the actions into a nifty menu? Something like @timdiggins gmail example could work well. It also allows for further actions to be added if and when needed (i.e. direct response to post like the WhatsApp quote reply feature could be cool). 

For the UI indication I'm thinking something similar to [Slack](https://www.howtogeek.com/wp-content/uploads/2015/06/img_5575fdc92854d.png) but reading 'Unread posts' along with adding a class (thredded--unread--post or thredded--read--post) so people can easily add their own CSS.

glebm:
@EllaNancyFay Sounds good!

If we are going to add a menu in the post header, perhaps we can move the other actions to the header or menu well (edit & delete). Kind of like the GitHub comment header?

It'd be nice if the dropdown is implemented without JavaScript (e.g. by changing the `display` of children on parent `&:focus, &:active`).

By the way, `PostView#cache_key` will need to include the read state for this to work with the current caching mechanism.

> For the UI indication I'm thinking something similar to Slack

:+1:

EllaNancyFay:
@glebm I'm struggling a bit with how to input the separator when the posts are rendered using a collection. I could move away from the collection and iterate through each post then add the separator where 
```if post.read_state != previous_post.read_state```
However this would do away with the collection cache... Any suggestions warmly welcomed

**Comment by glebm** [comment]
@EllaNancyFay `render` has a `spacer_template` argument! http://api.rubyonrails.org/classes/ActionView/PartialRenderer.html

**Comment by glebm** [comment]
@EllaNancyFay Initially I had thought we could use the `spacer_template` for this but that's not possible (the `spacer_template` is rendered only once [here](https://github.com/rails/rails/blob/92703a9ea5d8b96f30e0b706b801c9185ef14f0e/actionview/lib/action_view/renderer/partial_renderer.rb#L322)).

Perhaps we can implement our own caching method similar to how Rails does it [here](https://github.com/rails/rails/blob/92703a9ea5d8b96f30e0b706b801c9185ef14f0e/actionview/lib/action_view/renderer/partial_renderer/collection_caching.rb#L12):

So a `cached_collection_render_to_strings(collection)` method that would return a list or a hash of
`[ [item, "rendered html"], ... ]`.

We can then only cache the post contents (the only bit that's slow), removing the need for the complex `cache_key` method.

This seems like something that belongs in Rails core, so maybe could be merged eventually.

**Comment by EllaNancyFay** [comment]
@glebm I've moved the the actions into a drop down menu. Think it's looking good. 
screen shot 2017-02-28 at 14 50 15

What you're suggesting sounds like a good course of action for the navigating to last read post / inserting a separator above it. Let's split it out from this PR so one of us can pick it up in the coming weeks. 

Also code climate has picked up a couple of duplication issues, but I'm not sure how important they are to fix?

File: app/assets/stylesheets/thredded/components/_post.scss [hunk 1]
```diff
@@ -4,6 +4,71 @@
   @include thredded-media-mobile {
     margin-bottom: $thredded-base-spacing;
   }
+  .thredded--post--dropbtn .thredded--post--three-dot-menu {
```
glebm: 1. Can we remove the `.thredded--post--dropbtn` wrapper, as it doesn't seem necessary?
2. Let's rename `.thredded--post--three-dot-menu` to `.thredded--post--dropdown--toggle`.

app/assets/stylesheets/thredded/components/_post.scss
```diff
@@ -4,6 +4,71 @@
...
+    @media print {
+      display: none;
+    }
+  }
+
+  .thredded--post--actions {
+    display: none;
+    position: absolute;
+    right: 0;
+    background-color: #f9f9f9;
```
glebm: This doesn't work with the dark theme. Let's add a `$thredded-overlay-background` variable, and set the default to

app/assets/stylesheets/thredded/components/_post.scss
```diff
@@ -4,6 +4,71 @@
...
+    min-width: 170px;
+    box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.2);
+    z-index: 1;
+    cursor: pointer;
+  }
+
+  .thredded--post--actions a {
```
glebm: Let's use a single class here, such as `.thredded--post--dropdown--actions--item`.

app/assets/stylesheets/thredded/components/_post.scss
```diff
@@ -4,6 +4,71 @@
...
+    font-size: $thredded-font-size-small;
+    color: $thredded-text-color;
+    //color: $thredded-action-color;
+    text-decoration: none;
+    //transition: $thredded-action-transition;
+    padding: 5px 10px;
```
glebm: Let's use the Material dimensions from here: https://material.io/guidelines/components/menus.html#menus-specs

app/assets/stylesheets/thredded/components/_post.scss
```diff
@@ -4,6 +4,71 @@
...
+    }
+
+    &:active,
+    &:focus {
+      outline: none;
+    }
+
+  }
+
+  .thredded--post--dropdown:hover .thredded--post--actions {
```
glebm: Please rename `.thredded--post--actions` to `.thredded--post--dropdown--actions`.

app/assets/stylesheets/thredded/components/_post.scss
```diff
@@ -4,6 +4,71 @@
...
+    display: block;
+  }
+
+  .thredded--post--dropdown:hover .thredded--post--dropbtn .thredded--post--three-dot-menu  {
+    color: $thredded-action-hover-color;
+  }
 }
```
glebm: Please do not use nested selectors when it is possible to use single selectors instead.
With the renamings I've suggested above it should be possible to un-nest the entire block above.

app/views/thredded/posts_common/_mark_as_unread_link.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= link_to(t('thredded.topics.mark_as_unread'), post.mark_unread_path, method: :post, class:
```
glebm: Please use `button_to` instead, so that it works without JavaScript:

```erb
<%= button_to t('thredded.topics.mark_as_unread'), post.mark_unread_path,
              form: {class: 'thredded--post--mark-as-unread'} %>
```

app/controllers/thredded/posts_controller.rb
```diff
@@ -37,8 +37,22 @@ def destroy
                     notice: I18n.t('thredded.posts.deleted_notice')
     end
 
+    def mark_as_unread
+      authorize post, :read?
+      post.mark_as_unread(thredded_current_user, post.page(user: thredded_current_user))
```
glebm: This does not work for private messages (I'm guessing the `page` method needs to be updated)

app/views/thredded/posts_common/_actions.html.erb
```diff
@@ -1,11 +1,24 @@
-<% if post.can_update? %>
-  <%= link_to t('thredded.posts.edit'), post.edit_path, class: 'thredded--post--edit' %>
-<% end %>
+<div class='thredded--post--dropdown'>
```
glebm: This currently displays the dropdown toggle even if there are no actions available to the user.

I can think of two extension-friendly ways of fixing this:

1. With CSS: Move the toggle outside and after the actions container, and set:
    ```scss
    .thredded--post--dropdown--actions:empty + .thredded--post--dropdown--toggle {
      display: none;
    }
    ```
2. With HTML: render the actions to a string via `capture` first, and then check that the result is not blank before rendering the dropdown.

Either one is fine by me

EllaNancyFay:
@glebm - thanks for the review, was very insightful! Have made changes as per your comments :)

glebm:
Thanks!

PR #536, opened 2017-03-05
Implement render_collection_to_strings_with_cache

glebm:
With this we will multi-get only the post contents, and *not* cache the surrounding post UI.

Advantages:

* Allows state-dependent post UI as there is no need to worry about caching it anymore.
* Less cache load, as only the contents are cached. Different post views, such as a user post vs a post in a topic, re-use the same cached content.
* No need for the complex extension-unfriendly `PostView#cache_key` method.
* Bonus: multi-get caching even on Rails 4.

Disadvantage:

The `render_collection_to_strings_with_cache` implementation is complex.
It does use only the public  API of ActionView though.
Hoping for something like this to be included in ActionView itself
(https://groups.google.com/forum/#!topic/rubyonrails-core/Cx0GXz_k7Dc).

This idea comes from the discussion at
https://github.com/thredded/thredded/pull/533#issuecomment-282907746.

### DIFF
--- a/app/controllers/thredded/preferences_controller.rb
+++ b/app/controllers/thredded/preferences_controller.rb
@@ -4,8 +4,7 @@ module Thredded
     before_action :thredded_require_login!,
                   :init_preferences
 
-    def edit
-    end
+    def edit; end
 
     def update
       if @preferences.save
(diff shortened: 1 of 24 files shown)


glebm:
I've pushed this to Heroku master to see how it performs.

PR #539, opened 2017-03-07
Add Thredded.routes_id_constraint configuration option

jvoss:
Added Thredded.routes_id_contraint option to allow the user to specify the page constraint ID in lieu of forcing positive integers. Useful in cases where the underlying database is using a different ID type such as UUID. Default remains a positive integer. Additionally changed Thredded::PrivateTopicForm#normalized_user_ids to map to the string value of the user ID to support UUIDs (however this may not be the best option). I was unable to replicate your test environment, so I am unable to fully test these changes, however they are working without issue within my application.

### DIFF
--- a/app/forms/thredded/private_topic_form.rb
+++ b/app/forms/thredded/private_topic_form.rb
@@ -64,7 +64,7 @@ module Thredded
 
     def topic_categories
       if category_ids
-        ids = category_ids.reject(&:empty?).map(&:to_i)
+        ids = category_ids.reject(&:empty?)
         Category.where(id: ids)
       else
         []
@@ -82,8 +82,8 @@ module Thredded
     def normalized_user_ids
       user_ids
         .reject(&:empty?)
-        .map(&:to_i)
-        .push(user.id)
+        .map(&:to_s)
+        .push(user.id.to_s)
         .uniq
     end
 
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -2,8 +2,7 @@
 Thredded::Engine.routes.draw do # rubocop:disable Metrics/BlockLength
   resource :theme_preview, only: [:show], path: 'theme-preview' if %w(development test).include? Rails.env
 
-  positive_int = /[1-9]\d*/
-  page_constraint = { page: positive_int }
+  page_constraint = { page: Thredded.routes_id_constraint }
```
glebm: This is a page number for pagination, not an ID.

app/forms/thredded/private_topic_form.rb
```diff
@@ -82,7 +82,7 @@ def private_user_ids
     def normalized_user_ids
       user_ids
         .reject(&:empty?)
-        .map(&:to_i)
+        .map(&:to_s)
         .push(user.id)
```
glebm: This needs to be converted `to_s` as well, otherwise the `uniq` below won't work.

jvoss:
Added updates as requested.

PR #541, opened 2017-03-07
Updated the 'signed_in?' method to accept an unused, optional 'scope' parameter…

pak9rabid:
…, to make it compatible with the devise_security_extension gem

### DIFF
--- a/app/controllers/thredded/application_controller.rb
+++ b/app/controllers/thredded/application_controller.rb
@@ -44,7 +44,15 @@ module Thredded
       send(Thredded.current_user_method) || NullUser.new
     end
 
-    def signed_in?(scope=nil)
+    # When used with the devise_security_extension gem (https://github.com/phatworx/devise_security_extension),
+    # the gem's authentication implementation calls the 'signed_in?' method with a parameter named 'scope'.
+    # The original Thredded implementation of the 'signed_in?' method does not accept any parameters, so when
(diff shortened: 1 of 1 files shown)


glebm:
The `signed_in?` method is a method only used by Thredded controllers. How exactly does it make it compatible with that gem?

pak9rabid:
From what I can tell, devise_security_extensions is calling the version provided by thredded's application_controller.rb, and since it doesn't support the 'scope' argument that it's passing in, the app crashes with a "wrong number of arguments (given 1, expected 0)" error.  Here's a screenshot of the error I was getting before making this change:

screen shot 2017-03-08 at 11 04 11 am

glebm:
OK, please add a comment with an explanation and point to the line in the devise_security_extensions gem where it happens.

Also, all Ruby code is formatted and linted by rubocop, and it must report no issues. See https://github.com/thredded/thredded#ruby.

pak9rabid:
The requested changes have been made

glebm:
After considering it a bit, I've figured that the most robust way to solve it is to avoid conflicts altogether, and renamed `signed_in?` to `thredded_signed_in?` in #543.

We had done a similar change in the past, renaming `current_user` to `thredded_current_user` in #176.

glebm:
This should now be fixed on master!

pak9rabid:
Awesome, I appreciate yall's responsiveness on this!

pak9rabid:
Tested it out, works great.  Thanks!

PR #542, opened 2017-03-08
Incorporate user_display_name_method into autocomplete search results

jvoss:
Allows for custom user_display_name's to be visible in the drop down when autocomplete searches for users. Default functionality remains unchanged, however when a user has specified a user_display_name_method, it will append the output to the search drop down for friendly user identification.

### DIFF
--- a/app/controllers/thredded/autocomplete_users_controller.rb
+++ b/app/controllers/thredded/autocomplete_users_controller.rb
@@ -6,10 +6,11 @@ module Thredded
     def index
       authorize_creating PrivateTopicForm.new(user: thredded_current_user).private_topic
       users = params.key?(:q) ? users_by_prefix : users_by_ids
+
       render json: {
         results: users.map do |user|
           { id:         user.id,
-            name:       user.send(Thredded.user_name_column),
+            name:       Thredded.display_name.call(user),
             avatar_url: Thredded.avatar_url.call(user) }
         end
       }
--- a/lib/thredded.rb
+++ b/lib/thredded.rb
@@ -46,6 +46,7 @@ module Thredded
     :autocomplete_min_length,
     :active_user_threshold,
     :avatar_url,
+    :display_name,
     :email_from,
     :email_outgoing_prefix,
     :layout,
@@ -113,6 +114,10 @@ module Thredded
   def self.notifiers=(notifiers)
     notifiers.each { |notifier| BaseNotifier.validate_notifier(notifier) }
     @@notifiers = notifiers # rubocop:disable Style/ClassVars
+  end
+
+  def self.display_name(user)
+    [user.send(Thredded.user_name_column), user.send(Thredded.user_display_name_method)].uniq.join(' - ')
   end
 
   def self.user_display_name_method
--- a/lib/thredded.rb
+++ b/lib/thredded.rb
@@ -98,6 +98,10 @@ module Thredded
   self.autocomplete_min_length = 2
   self.routes_id_constraint = /[1-9]\d*/
 
+  self.display_name = ->(user) do
+    [user.send(Thredded.user_name_column), user.send(Thredded.user_display_name_method)].uniq.join(' - ')
+  end
+
   # @return [Thredded::AllViewHooks] View hooks configuration.
   def self.view_hooks
     instance = Thredded::AllViewHooks.instance
@@ -114,10 +118,6 @@ module Thredded
   def self.notifiers=(notifiers)
     notifiers.each { |notifier| BaseNotifier.validate_notifier(notifier) }
     @@notifiers = notifiers # rubocop:disable Style/ClassVars
-  end
-
-  def self.display_name(user)
-    [user.send(Thredded.user_name_column), user.send(Thredded.user_display_name_method)].uniq.join(' - ')
   end
 
   def self.user_display_name_method


app/controllers/thredded/autocomplete_users_controller.rb
```diff
@@ -6,17 +6,27 @@ class AutocompleteUsersController < Thredded::ApplicationController
...
-        results: users.map do |user|
-          { id:         user.id,
-            name:       user.send(Thredded.user_name_column),
-            avatar_url: Thredded.avatar_url.call(user) }
-        end
+          results: users.map do |user|
+            { id:         user.id,
+              name:       display_name(user),
+              avatar_url: Thredded.avatar_url.call(user) }
+          end
```
jayroh: Looks like CI is failing with rubocop complaining about the indentation here.

app/controllers/thredded/autocomplete_users_controller.rb
```diff
@@ -6,17 +6,27 @@ class AutocompleteUsersController < Thredded::ApplicationController
...
 
+    # Defines how the name is displayed in the autocomplete search results
+    def display_name(user)
+      if Thredded.user_display_name_method == :to_s
+        user.send(Thredded.user_name_column)
+      else
+        "#{user.send(Thredded.user_name_column)} - #{user.send(Thredded.user_display_name_method)}"
+      end
+    end
+
```
jayroh: This isn't too heavy so I could see leaving this in the controller as it is, but extracting this out to a small service class that does this work would increase testability and would also NOT increase the surface area of the controller class.  Something like `Thredded::DisplayName`. This method could get removed and the hash above could call it with something like

```ruby
{ id:         user.id,
  name:       Thredded::DisplayName.new(user).to_s,
  avatar_url: Thredded.avatar_url.call(user) }
```

Actually, now that I'm looking at it. This could take the lead of `avatar_url` and be a configurable/overridable Proc.

```ruby
{ id:         user.id,
  name:       Thredded.display_name.call(user),
  avatar_url: Thredded.avatar_url.call(user) }
```

app/controllers/thredded/autocomplete_users_controller.rb
```diff
@@ -6,17 +6,27 @@ class AutocompleteUsersController < Thredded::ApplicationController
...
+              avatar_url: Thredded.avatar_url.call(user) }
+          end
       }
     end
 
     private
 
+    # Defines how the name is displayed in the autocomplete search results
+    def display_name(user)
+      if Thredded.user_display_name_method == :to_s
```
glebm: I'm fine with adding the ability to customize this, but this check should either be `Thredded.user_display_name_method == Thredded.user_name_column`, or alternatively no check just:

```ruby
[user.send(Thredded.user_name_column), user.send(Thredded.user_display_name_method)].uniq.join(' - ')
```

app/controllers/thredded/autocomplete_users_controller.rb
```diff
@@ -6,17 +6,27 @@ class AutocompleteUsersController < Thredded::ApplicationController
...
       authorize_creating PrivateTopicForm.new(user: thredded_current_user).private_topic
       users = params.key?(:q) ? users_by_prefix : users_by_ids
+
       render json: {
-        results: users.map do |user|
-          { id:         user.id,
-            name:       user.send(Thredded.user_name_column),
-            avatar_url: Thredded.avatar_url.call(user) }
-        end
+          results: users.map do |user|
```
glebm: rubocop

app/controllers/thredded/autocomplete_users_controller.rb
glebm: This is not the display name, this is only meant to be used for displaying the autocomplete results in @-mentions and private message recipients, so I don't think `Thredded.display_name` would be the right name for the proc.

Also, maybe we don't need this to be a setting. It's probably rare to want to customize it, and one can always override the controller method if they need to.

jvoss: I do not think it would be rare at all if the display name is different from the user name column. Are you suggesting I just override this locally and forget about the pull request entirely?

glebm: No, I think we're all confused :)

1. There is already a `Thredded.user_display_name_method` (not the same as `user_name_column`). It is used in certain places such as before each post.
2. What you want is a different display *only for the autocomplete*. What I'm saying to @jayroh is that it should definitely *not* be called `display_name` as that would be very confusing.
3. For your PR, I think it makes sense to do the thing I had suggested originally, and *not* add a new setting, but let's wait for @jayroh's to get back on this.

jayroh: yeah @glebm is right. I didn't tie my previous suggestion with what already exists in the app. I like the idea of a setting with a default, but in this case, maybe it's not such a big deal to just throw it in the controller.

It just has a bit of a non-SRP smell to it, is all.

What do you guys think?

jvoss: So what I did for now is just fork off into another branch to update the [controller](https://github.com/jvoss/thredded/blob/bluesky/app/controllers/thredded/autocomplete_users_controller.rb) for my project. I wound up creating a method in the controller (#autocomplete_display_name) that provides the original functionality I thought you guys might like. 

I'm sure there is further tweaking (as in a configuration option) but basically as it stands, if the user is using anything other than #to_s for the display name method it will include their custom display method in addition to their forum user name.

Take a look at that and let me know what you guys think. If you have suggestions on how you might like it included (if at all) I'm certainly open to suggestions.

PR #547, opened 2017-03-12
Remove the email-specific post_notifications table

glebm:
We now track all notifications via user_post_notifications, so this
table is redundant.

Removing it simplifies the notifier code, reducing it to the
responsibility of only delivering stuff, and not deciding who to notify.


jayroh:
👍 Good to go!

timdiggins:
LGTM 👍

PR #550, opened 2017-03-14
Email styles

glebm:
1. Add `_email.scss` with some email styles.
2. Add roadie-rails to the demo app and use it for email previews.
3. Update readme.


jayroh:
@glebm aside related to the emails - if you have the default email templates in front of you, could you stick a few screenshots in the wiki?

glebm:
The emails can be previewed at https://thredded.org/emails/en/thredded__post_mailer_preview-post_notification. I'll link that from the wiki.

PR #554, opened 2017-03-18
Auto-follow messageboard preference follow-up

glebm:
See #488 - by changing the column default specified in the README, users
will (by default) auto-follow topics in all newly created messageboards.
You can also togggle individual messageboard auto-follows (like an
'important notices' board) for specific message boards by running a data
migration targeting the ThreddedUserMessageboardPreference model. Users
can turn this functionality off at the global level or for particular
boards.

### DIFF
--- a/app/commands/thredded/autofollow_users.rb
+++ b/app/commands/thredded/autofollow_users.rb
@@ -1,18 +1,23 @@
 # frozen_string_literal: true
 module Thredded
-  class AutofollowMentionedUsers
+  class AutofollowUsers
     def initialize(post)
       @post = post
     end
 
     def run
       autofollowers.each do |user|
-        Thredded::UserTopicFollow.create_unless_exists(user.id, post.postable_id, :mentioned)
+        reason = mentioned_users.include?(user) ? :mentioned : :auto
+        Thredded::UserTopicFollow.create_unless_exists(user.id, post.postable_id, reason)
       end
     end
 
+    def mentioned_users
+      @mentioned_users ||= Thredded::AtNotificationExtractor.new(post).run
+    end
+
     def autofollowers
-      autofollowers = Thredded::AtNotificationExtractor.new(post).run
+      autofollowers = (include_auto_followers + mentioned_users).uniq
       autofollowers.delete(post.user)
       exclude_those_opting_out_of_at_notifications autofollowers
     end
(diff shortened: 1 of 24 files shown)


zapnap:
This looks fantastic! Thanks @glebm 

Apologies that I haven't had more free cycles to contribute lately and thanks again for cleaning up my mess :)

glebm:
I'm wondering how the global setting for this should behave with regards to the per-messageboard setting.

In the current version of this PR, this is what happens:

| Global | Board | Follow? |
| :-:  | :-: | :-: |
| ✅ | ✅ | ✅|
| ✅ | ❌ | ❌|
| ❌ | ✅ | ✅|
| ❌ | ❌ | ❌|

As per the above, the global setting (`user_preference#auto_follow_topics?`) does not have any effect with this setup. It is currently only used as the default value for the board setting (`user_messageboard_preference#auto_follow_topics?`). The settings change independently so long as the records exist in the database.

This seems potentially confusing.

zapnap:
Your interpretation is the correct one, I think. At least that was the intent. The global setting is really just a default. Open to better ideas if folks find that confusing.

glebm:
I think this would work if changing the global setting would also change all the messageboard ones.

So if I change global from false to true, all the existing board preferences would be updated.

Also, current version of this PR does not actually use the user's global preference as the default, need to fix and add more tests.

glebm:
I've implemented the stuff in the comment above. Ready to review / merge if everyone agrees with this behaviour.

app/commands/thredded/autofollow_users.rb
```diff
@@ -0,0 +1,56 @@
...
+    #   UserMessageboardPreference#auto_follow_topics? && UserPreference#auto_follow_topics?
+    # If the `user_preference` for a given does not exist, its default value is used.
+    # @return [Enumerable<Thredded.user_class>]
+    def auto_followers
+      user_board_prefs = post.messageboard.user_messageboard_preferences.each_with_object({}) do |ump, h|
+        h[ump.user] = ump
+      end
+      Thredded.user_class.includes(:thredded_user_preference)
+        .select(Thredded.user_class.primary_key)
+        .find_each(batch_size: 50_000).select do |user|
```
glebm: One way to avoid checking every user would be to ensure a `Thredded::UserMessageboardPreference` always exists for each user (currently it is created lazily when necessary).

However, this only takes on the order of ~100ms / 100k users, and it happens in a background job, so it's OK to do it leave as is for now.

zapnap: 👍

PR #556, opened 2017-03-20
Add russian localization file

xuliaP:
may be add gem 'babosa' for friendly slug?


xuliaP:
https://github.com/norman/babosa
Currently supported languages include:

Bulgarian
Danish
German
Greek
Macedonian
Norwegian
Romanian
Russian
Serbian
Spanish
Swedish
Ukrainian

jayroh:
So this would be *instead of* friendly_id? And mostly for the reason of non-ascii characters?

xuliaP:
in addition

config/locales/ru.yml
```diff
@@ -0,0 +1,381 @@
...
+      mark_as_unread: Пометить как непрочитанное
+      not_following: не отслеживаю
+      locked:
+        label: Блокировать
+      sticky:
+        label: Закрепить
+    users:
+      currently_online: Онлайн
+      user_started_topic_html: Сделана запись
+  activerecord:
```
glebm: All the keys from here and below are [part of rails-i18n](https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/ru.yml) and we should not be duplicating them here.

config/locales/ru.yml
```diff
@@ -0,0 +1,381 @@
+---
+ru:
+  thredded:
+    content_moderation_states:
+      content_blocked_notice: Блокировано
+      content_blocked_notice_with_record_html: Блокировано %{moderator} %{time_ago}
+    email_notifier:
+      by_email: по почте
```
glebm: Perhaps this should say "по электронной почте" (not sure though).

config/locales/ru.yml
```diff
@@ -0,0 +1,381 @@
...
+    moderation:
+      approve_btn: Одобрить
+      block_btn: Заблокировать
+      pending:
+        empty:
+          content: Все посты промодерированы.
+          title: Хорошая работа!
+      post_approved_html: Пост одобрен %{moderator} %{time_ago}.
+      post_blocked_html: Пост заблокирован %{moderator} %{time_ago}.
+      post_deleted_notice: Этот пост удален.
```
glebm: nit: "удалён" here and below

config/locales/ru.yml
```diff
@@ -0,0 +1,381 @@
...
+        other: Начато %{count} тем
+      user_posted_in_topic_html: "%{user_link} сделал запись в %{topic_link}"
+      user_since_html: Пользователь был %{time_ago}
+      user_started_topic_html: "%{user_link} начал %{topic_link}"
+      unfollow: Не следить за темой
+      unfollowed_notice: Вы не следите за темой
+      mark_as_unread: Пометить как непрочитанное
+      not_following: не отслеживаю
+      locked:
+        label: Блокировать
```
glebm: "Блокировать" is already used for moderation. How about we call this "Закрыть" instead? It seems more appropriate as "закрыть тему" is a common expression.

glebm:
I'm not sure whether we should add babosa as a dependency or just document how to add it instead.

Babosa does not work for all languages, e.g. for Japanese one might want to not use babosa, and process with [mojinizer](https://github.com/ikayzo/mojinizer) instead.

I think we should perhaps just document the usage of babosa with Thredded, and add a config setting, such as:

```ruby
Thredded.slugifier = -> input { input.parameterize }
```

```ruby
class Thredded::Post < ActiveRecord::Base
  ...

  protected

  def normalize_friendly_id(input)
    Thredded.slugifier.call(input.to_s)
  end
end
```

config/locales/ru.yml
xuliaP: email - international word, and i think dont translate it

config/locales/ru.yml
xuliaP: ok..i fix it

xuliaP:
sorry

config/locales/ru.yml
glebm: Я имел в виду точку над "ё", а не кавычки

glebm:
@xuliaP The review comments need to be addressed and the tests need to pass before this can be merged. You can use Google Translate for translating the new keys into other languages (`GOOGLE_TRANSLATE_API_KEY=your-api-key i18n-tasks translate-missing`).

xuliaP:
насчет ё..сейчас уже не используется (what country are you from?) maybe i add only what i changed in all yml(english world)?

glebm:
> насчет ё..сейчас уже не используется (what country are you from?) 

OK. I am from Russia but I haven't lived there for 10 years.

> maybe i add only what i changed, in all yml(english word)? and naturally my translate is not perfect..but i hope...more followers check translate and correct it

OK, sure.

PR #563, opened 2017-03-29
UI: Show search on the left, user nav on the right

### LINKED ISSUE #561: Nav: flip the order
Usually, the user name is displayed in the top-right.
It makes sense for private messages and notifications to be close to the user name.

Perhaps we should swap search and user nav, i.e. move Search to the left and user nav to the right?

glebm:
The username / profile link is usually rendered on top right, so it makes sense to keep user nav (notification settings, private messages) close.

thredded-nav-flip

Fixes #561


jayroh:
Beautiful! :heart:

PR #568, opened 2017-04-10
Always return 404 for missing messageboards

### LINKED ISSUE #557: Missing Template in the thredded controller for formats other than HTML
An ActionView::MissingTemplate occurred in topics#index:

Missing template thredded/error_pages/not_found with {:locale=>[:en], :formats=>[:json], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :jbuilder]}. Searched in:
* "/var/app/current/app/views"
* "/opt/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/devise-4.2.0/app/views"
* "/opt/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/thredded-0.10.0/app/views"
* "/opt/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/kaminari-core-1.0.1/app/views"
* "/opt/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/devise-i18n-1.1.1/app/views"
------------------------------- Request:

* URL        : https://www.myforumlink.com/forum/composer.json
* HTTP Method: GET
* IP address : *
* Parameters : {"controller"=>"thredded/topics", "action"=>"index", "messageboard_id"=>"composer", "format"=>"json"}
* Timestamp  : 2017-03-19 08:16:02 UTC
* Server : *
* Rails root : /var/app/current
* Process: 24539

Seems like there are some people trying to find out information about the core of the server and application. Looks like the error handler for thredded controllers can only deal with HTML format. I am seeing this on topic#show and topic#show, possibly others.

cbortz:
Always return a 404 in `TopicsController#index` when the `messageboard` isn't found, regardless of request format (HTML, JSON, etc).

Fixes #557

### DIFF
--- a/app/controllers/thredded/application_controller.rb
+++ b/app/controllers/thredded/application_controller.rb
@@ -20,7 +20,7 @@ module Thredded
                 Thredded::Errors::UserNotFound do |exception|
       @error   = exception
       @message = exception.message
-      render template: 'thredded/error_pages/not_found', status: :not_found
+      render template: 'thredded/error_pages/not_found', status: :not_found, formats: [:html]
     end
 
     rescue_from Pundit::NotAuthorizedError,
--- a/spec/controllers/thredded/topics_controller_spec.rb
+++ b/spec/controllers/thredded/topics_controller_spec.rb
@@ -13,8 +13,7 @@ module Thredded
       allow(controller).to receive_messages(
         topics:        [@topic],
         cannot?:       false,
-        the_current_user:  user,
-        messageboard:  @messageboard
+        the_current_user:  user
       )
     end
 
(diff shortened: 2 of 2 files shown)


glebm:
This should only happen if the client indicates HTML as the request content type. If the client requests anything else, we should return 406 instead.

Perhaps simply disable all the other request formats in the routes?

cbortz:
@glebm sounds good to me. I'll update this later today, and update the specs if necessary. Cheers!

cbortz:
@glebm I began implementing this I think this could use some further thought & discussion. Here are some of my thoughts:

- There are certainly more routes where someone may try to request an unsupported format. Which routes should default to this behavior?
  + Topic#index
  + Topics#show ?
  + ?
- Format constraints placed in the routes will return 404 when not matched, rather than 406
- How would a user customize this behavior?
  + Would they need to check the routes file into their repository?
  + What if I want my topics page to respond to JSON?
  + What if I want my 404 page to return JSON?

---

With all of these questions, I'm starting to think this is a non-issue. Let's ignore any of the proposed fixes for a moment...

In a production environment, the current codebase would return a 500 with a JSON request, whether the messageboard is found or not, because JSON is an unsupported format either way. This is technically correct, but not semantically. I think this is standard behavior though.

It might be the simplest to document errors and supported formats. If someone wants to use JSON, they could just implement something like [Jbuilder](https://github.com/rails/jbuilder) and check those files into their repository.

What do you think?

PR #576, opened 2017-04-20
Make topic slugs unique across messageboards

### LINKED ISSUE #573: Redirect to canonical topic url if messageboard slug is wrong
A topics may be moved to another messageboard. Need to redirect instead of 404, so that the old URLs still work.

glebm:
This allows us to correctly redirect to the new topic URL when the topic's messageboard has changed.

Resolves #573

Needs code review!

### DIFF
--- a/app/models/thredded/topic.rb
+++ b/app/models/thredded/topic.rb
@@ -22,8 +22,7 @@ module Thredded
 
     extend FriendlyId
     friendly_id :slug_candidates,
-                use:            [:history, :reserved, :scoped],
-                scope:          :messageboard,
+                use: %i(history reserved),
                 # Avoid route conflicts
                 reserved_words: ::Thredded::FriendlyIdReservedWordsAndPagination.new(%w(topics))
 
@@ -153,7 +152,8 @@ module Thredded
     def slug_candidates
       [
         :title,
-        [:title, '-topic'],
+        [:title, '-', messageboard.try(:name)],
+        [:title, '-', messageboard.try(:name), '-topic']
       ]
     end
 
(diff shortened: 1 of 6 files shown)


app/models/thredded/topic.rb
```diff
@@ -22,8 +22,7 @@ class Topic < ActiveRecord::Base # rubocop:disable Metrics/ClassLength
 
     extend FriendlyId
     friendly_id :slug_candidates,
-                use:            [:history, :reserved, :scoped],
-                scope:          :messageboard,
+                use: %i(history reserved),
                 # Avoid route conflicts
```
timdiggins: do the words "Avoid route conflicts" fill anyone else with horror?

timdiggins:
Having a look at the PR, but this allows me to ask a question about FriendlyId. 

How committed are we to the current "Friendly" urls. 

I so much prefer trello's style of friendly urls (id plus descriptive but irrelevant text) which are automatically unique and don't really require much special treatment. Anyone who looks at URLs a lot will understand the trello ones. And many people who don't understand the trello ones won't bother to look at the URLs at all.

I just feel FriendlyId's (unique, historical, id-hiding) url is causing a lot of work and some danger (see https://github.com/thredded/thredded/pull/576/commits/a196e550e59b61866f8445c3338e3f4f6fa4bfc5#r113638127 for example). Is it really worth this effort. (kind of obvious which way I fall on this, but what do you think @glebm, @jayroh

glebm:
I think it's worth the effort due because these URLs are slightly more SEO-friendly. They're hard to get right, but I think with this PR we will have nailed it.

> And many people who don't understand the trello ones won't bother to look at the URLs at all.

Not in the browser URL bar, but the URLs are also displayed in Google search results. Compare:

a) The full URL is displayed:
plforums-search-result
b) Part of the URL is hidden because of the numeric ID taking up space:
discourse-search-result

timdiggins:
good answer @glebm -- ok, I can see why this makes sense to public forum users.

PR #594, opened 2017-05-30
Added Require User Authentication Section to readme

nextgenappsllc:
Added Require User Authentication section to permissions portion of the readme to clarify a simple way to add internal discussions boards secluded from the public.

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,6 +539,16 @@ Currently, the default behaviour is to render an error message with an appropria
 layout. You may want to override the handling for `Thredded::Errors::LoginRequired` to render a login form instead.
 For an example of how to do this, see the initializer.
 
+### Require User Authentication (private forum)
+
+To enforce users being autheticated before using thredded you can the appropriate `before_action` to `Thredded::ApplicationController`.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -539,6 +539,16 @@ Currently, the default behaviour is to render an error message with an appropria
 layout. You may want to override the handling for `Thredded::Errors::LoginRequired` to render a login form instead.
 For an example of how to do this, see the initializer.
 
+### Require User Authentication (private forum)
+
+To enforce users being autheticated before using thredded you can the appropriate `before_action` to `Thredded::ApplicationController`.
+
+If you are using Devise and your model name is "User" then you could put the following in the "thredded.rb" initializer.
```
glebm: This is not the right way to do it, as it depends on devise.
Here is how to do this in a way that does not depend on the auth framework:

```ruby
# config/initializers/thredded.rb
Rails.application.config.to_prepare do
  Thredded::ApplicationController.before_action :thredded_require_login!
end
```

nextgenappsllc: Ah this is much better, thanks!

PR #612, opened 2017-08-16
Fix for module qualified user class

embmicro:
While trying to integrate Thredded with Spree I set Thredded.user_class to "Spree::User" which caused Thredded.user_class_name to be "spree/user" instead of "user" breaking most methods. This was the fix that seems to be working for me. I'm pretty new to Ruby/Rails so not sure if this is the best way to fix it.

### DIFF
--- a/lib/thredded.rb
+++ b/lib/thredded.rb
@@ -143,7 +143,7 @@ module Thredded
     unless user_class_name.is_a?(String)
       fail "Thredded.user_class must be set to a String, got #{user_class_name.inspect}"
     end
-    @@user_class_name = user_class_name # rubocop:disable Style/ClassVars
+    @@user_class_name = user_class_name.demodulize # rubocop:disable Style/ClassVars
   end
 
   # @return [Class<Thredded::UserExtender>] the user class from the host application.


glebm:
`"spree/user"` is an `underscore`d version of `Spree::User`, however `Thredded.user_class_name` is never underscored as far as I can tell.

Where did you see `spree/user`?

lib/thredded.rb
```diff
@@ -143,7 +143,7 @@ def self.user_class=(user_class_name)
     unless user_class_name.is_a?(String)
       fail "Thredded.user_class must be set to a String, got #{user_class_name.inspect}"
     end
-    @@user_class_name = user_class_name # rubocop:disable Style/ClassVars
+    @@user_class_name = user_class_name.demodulize # rubocop:disable Style/ClassVars
```
glebm: This is incorrect as it may no longer refers to the same class. E.g. `"Spree::User".demodulize` is "User" (which could be `::User` etc).

embmicro: The problem was because I didn't override _standalone.html.erb which does use Thredded.user_class_name.underscore. Overriding this and removing the underscore calls in the initializer made it work without the change.

So an actual fix may be to add demodulize to these two places.

glebm:
Fixed in the commit above

PR #615, opened 2017-09-02
Ensure users are notified by all notifiers, not just the first 

### LINKED ISSUE #540: Behaviour with multiple notifiers.
In a situation with multiple notifiers, only the first notifier will be used because of #529 

This means that the notifiers are (currently) effectively a choice.

However the UI (and previous behaviour) allowed users to choose multiple notifiers if they wanted. (it turns out some people actually want this).

Two easy options:

1) add notifier key to UserPostNotification (and add this to the uniqueness criterion)
2) change UI to only allow one Notifier to run (reducing functionality)

any other options?

timdiggins:
fixes #540.

Note that in rare cases (I think only when background jobs are misconfigured or have an unreliable lock) this implementation might result in double notification.

### DIFF
--- a/app/commands/thredded/notify_following_users.rb
+++ b/app/commands/thredded/notify_following_users.rb
@@ -11,10 +11,10 @@ module Thredded
     def run
       Thredded.notifiers.each do |notifier|
         notifiable_users = targeted_users(notifier)
-        notifiable_users = notifiable_users.select do |user|
-          # Create a notification for the user.
+        notifiable_users.each do |user|
+          # Record idempotently that the notification happened
           # If a notification was already created (from another thread/process),
-          # this will return false due to the unique constraint on the table
+          # this won't create another notification, but will renotify (too bad)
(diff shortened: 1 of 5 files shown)


glebm:
How could it result in double notification?

`Thredded::UserPostNotification.create_from_post_and_user` should be atomic

timdiggins:
@glebm It could happen if there are two instances running `NotifyFollowingUsers` at precisely the same time. One would create the `UserPostNotification`, and the other would silently allow it to happen. But it's only going to happen vanishingly rarely and only (I beleive) as a result of a misconfiguration of the background job processor (which should have locks for precisely this kind of thing)

timdiggins:
@glebm to be precise about what I mean by "double notification" -- sending two identical emails (or push notifications or whatever). But I don't think we should worry about this really.

glebm:
@timdiggins Got it. I have an idea for an easy fix, merging this first.

PR #616, opened 2017-09-02
move BaseNotifier into /lib so it can be used by plugins

timdiggins:


timdiggins:
@glebm I'm trying to share my pushover notifier as an example, but I need access to BaseNotifier (for validation in the specs in fact). I think if it's in /app I can't access it. I know you've had lots of painful experience with moving things from autoloaded to non autoloaded locations. The specs are now passing, but wanted to run it past you.

glebm:
This might break auto-loading, but if it's necessary then so be it

PR #629, opened 2017-10-12
A simple message for depicting if a topic is locked

akashaggarwal7:
I'd first of all like to thank you for this amazing project. It must be my lucky day to have come across it. I love how the code base is so neat and clean.

While exploring, I noticed that it wouldn't let a user know if a topic is locked or not and just simply doesn't render a form. I added this- 
image

### DIFF
--- a/app/assets/stylesheets/thredded/components/_topics.scss
+++ b/app/assets/stylesheets/thredded/components/_topics.scss
@@ -172,3 +172,10 @@
     display: none;
   }
 }
+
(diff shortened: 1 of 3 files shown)


glebm:
Thanks! :+1: I've made a few adjustments and released v0.13.6.

akashaggarwal7:
Glad you accepted it, I'll continue making more changes over time because I'm really thrilled to work on this. We can get talking somewhere if you'd like and discuss ideas :)

PR #635, opened 2017-10-29
Messageboard locking feature

akashaggarwal7:
The feature lets administrators to lock a messageboard.
If a messageboard is locked, an icon gets displayed that messageboard category and users may not post new topics. Users may still reply to a topic if the topic is unlocked.

screen shot 2017-10-29 at 7 42 30 pm


akashaggarwal7:
screen shot 2017-10-29 at 7 42 30 pm

app/assets/stylesheets/thredded/components/_messageboard.scss
```diff
@@ -43,6 +43,15 @@
   margin-bottom: $thredded-small-spacing / 2;
 }
 
+.threadded--messageboard--icon {
+  fill: currentColor;
```
glebm: Please replace this line with

```scss
@extend %thredded--icon;
```

This makes it easier for the user to customize all the icons.

app/models/thredded/user_permissions/write/all.rb
```diff
@@ -9,7 +9,8 @@ module All
 
         # @return [ActiveRecord::Relation<Thredded::Messageboard>] messageboards that the user can post in
         def thredded_can_write_messageboards
-          Thredded::Messageboard.all
+          # Thredded::Messageboard.all
+          Thredded::Messageboard.where(locked: false)
```
glebm: I believe it's better to revert the changes in this file and do the check on the `MessageboardPolicy` level.

This is because we must filter `PostPolicy#create?` based on this method, but we want to allow posting to the existing topics.

This filtering already happens via `MessageboardPolicy#post?` but for this we'll replace that with `@user.thredded_can_write_messageboards.include?(@messageboard)` in `PostPolicy`.

app/models/thredded/user_permissions/write/all.rb
```diff
@@ -18,7 +19,8 @@ module ClassMethods
           # @param _messageboards [Array<Thredded::Messageboard>]
           # @return [ActiveRecord::Relation<Thredded.user_class>] users that can post to the given messageboards
           def thredded_messageboards_writers(_messageboards)
```
glebm: This method was not actually used anywhere so I've removed it in https://github.com/thredded/thredded/commit/29be9c4f884a061001779692805a8e0cb93c210b.

app/policies/thredded/post_policy.rb
```diff
@@ -26,7 +26,7 @@ def initialize(user, post)
     end
 
     def create?
-      @user.thredded_admin? || !@post.postable.locked? && messageboard_policy.post?
+      !@user.is_a?(Thredded::NullUser) && (@user.thredded_admin? || !@post.postable.locked?)
```
glebm: The `NullUser` check is not necessary because `NullUser` also implements the permissions methods.

Please replace with:

```ruby
def create?
  @user.thredded_admin? ||
    !@post.postable.locked? &&
    # Users are allowed to post in unlocked topics of a locked messageboard
    # only if they would be allowed to post if it weren't locked.
    @user.thredded_can_write_messageboards.include?(@post.messageboard)
end
```

app/views/thredded/topics/index.html.erb
```diff
@@ -9,10 +9,12 @@
   </div>
 
   <%= content_tag :section, class: 'thredded--main-section thredded--topics', 'data-thredded-topics' => true do %>
+    <% if policy(messageboard).post? %>
```
glebm: This check is unnecessary because `@new_topic` will only be set in the controller if `policy(messageboard).post?` is `true`.

Please revert the changes in this file.

config/locales/en.yml
```diff
@@ -59,6 +59,8 @@ en:
         no_group: No Group
         title_label: Name
         update_btn_submitting: :thredded.form.update_btn_submitting
+        locked:
```
glebm: Please used the same format as for the other fields.
You can update the locale files by running:

```bash
i18n-tasks mv thredded.messageboard.form.locked.label thredded.messageboard.form.locked_label
```

config/locales/ru.yml
```diff
@@ -58,6 +58,8 @@ ru:
         no_group: Не группировать
         title_label: Название форума
         update_btn_submitting: :thredded.form.update_btn_submitting
+        locked:
+          label: Заблокировать тему
```
glebm: This translation says "Lock topic", please use "Заблокировать" instead.

app/policies/thredded/post_policy.rb
```diff
@@ -26,7 +26,7 @@ def initialize(user, post)
     end
```
glebm: Please replace the **`MessageboardPolicy`**`#post?` method with:

```ruby
    def post?
      @user.thredded_admin? ||
        (!@messageboard.locked? || moderate?) &&
        @user.thredded_can_write_messageboards.include?(@messageboard)
    end
```

app/views/thredded/messageboards/_messageboard.html.erb
```diff
@@ -19,6 +19,8 @@
                 user: messageboard.last_user.thredded_display_name %>
         </p>
       <% end %>
+      <br>
```
glebm: Perhaps it will look better with the icon in the top right corner?
Will also save space on mobile.

app/views/thredded/messageboards/_messageboard.html.erb
```diff
@@ -19,6 +19,8 @@
...
+      <%= inline_svg('thredded/lock.svg', class: 'threadded--messageboard--icon') if messageboard.locked %>
```
glebm: This tag should have a `title` attribute for accessibility. How about:

> This messageboard is locked. Only moderators can create new topics here.

app/assets/stylesheets/thredded/components/_messageboard.scss
```diff
@@ -43,6 +43,15 @@
   margin-bottom: $thredded-small-spacing / 2;
 }
 
+.threadded--messageboard--icon {
```
glebm: There is a typo here, should be **thre**~a~**dded--messageboard--icon**.

config/locales/en.yml
akashaggarwal7: The only other place `Locked` is used is for topics, and that has 2 child nodes `label` and `message` so this was actually following the format. Do you still want me to push the change you suggested?

app/models/thredded/user_permissions/write/all.rb
akashaggarwal7: I'm leaving this unchanged so you could resolve this merge conflict.

app/views/thredded/messageboards/_messageboard.html.erb
akashaggarwal7: I'm not too sure about this. Moving it to far right would destroy the symmetry. Your space point is valid though. I too thought about it but I traded that off because this was the most pleasing for me. Perhaps we could use another icon or something else for providing visual information?

I tried this, didn't find it appealing either: 
image

app/views/thredded/messageboards/_messageboard.html.erb
akashaggarwal7: Good point, and that message seems suitable. I didn't change this yet since we'd need a translation node name for it.

app/policies/thredded/post_policy.rb
akashaggarwal7: This is done, you commented on the wrong line.

akashaggarwal7:
@glebm I added the added/changed most of the stuff and replied to your previous comments where I need your opinion.

app/models/thredded/user_permissions/write/all.rb
glebm: Please rebase so that there are no merge conflicts. As a result, there should be no changes made to this file.

app/views/thredded/messageboards/_messageboard.html.erb
glebm: How about moving it right after the title (on the same line)?

config/locales/en.yml
glebm: Well in this case the key is under `messageboard.form` but OK either way. We may want to add a message on the form explaining how a locked messageboard behaves later.

app/views/thredded/messageboards/_messageboard.html.erb
```diff
@@ -19,6 +19,8 @@
                 user: messageboard.last_user.thredded_display_name %>
         </p>
       <% end %>
+      <br>
+      <%= inline_svg('thredded/lock.svg', class: 'thredded--messageboard--icon') if messageboard.locked %>
```
glebm: How about `thredded.messageboard.locked_notice` key for the tooltip?

app/views/thredded/messageboards/_messageboard.html.erb
akashaggarwal7: I think it'd be out of symmetry too. That was the first thing that I tried because it was the easiest.

config/locales/en.yml
akashaggarwal7: Ok

app/views/thredded/messageboards/_messageboard.html.erb
akashaggarwal7: Cool, just left to decide for the placement and the translations. Should I push with the latest screenshot that I posted?

app/views/thredded/messageboards/_messageboard.html.erb
glebm: OK, let's go with the one in the screenshot then!

akashaggarwal7:
@glebm Everything looks great to me. I rebased and cleaned my commits as well: https://github.com/akashaggarwal7/thredded/commits/master

glebm:
Looks good! 👍 

This will need to wait until the next patch release, as this change has a database migration so we'll need to bump the version to v0.14.0, and I want to release some bugfixes for the last v0.13.x release.

I'll also have to change the migration after merging because we handle migrations differently (`create_thredded` always has everything in one migration, and `upgrade_migrations` are provided for upgrading from one minor version to the next one).

akashaggarwal7:
Okay! Can't wait for it :)

glebm:
@akashaggarwal7 Merged! I've had to make some fixes and minor changes, please have look.

akashaggarwal7:
@glebm Great, thanks! I liked how you separated the messageboard meta partial in d46e8465d4f30a19d3bea1e8fdb63110230b6e82.

akashaggarwal7:
@glebm The feature isn't reflected on https://thredded.org/thredded. Perhaps we need a restart there and maybe you'd want to lock a messageboard for demo.

PR #644, opened 2017-11-14
Allow development with bundler 1.16.0

timdiggins:
Before v1.16.0, eval_gemfile used path relative to Bundler root.
Now it seems to be using the path relative to the gemfile.

### DIFF
--- a/spec/gemfiles/rails_4_2.gemfile
+++ b/spec/gemfiles/rails_4_2.gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
 gemspec path: '../../'
-eval_gemfile './shared.gemfile'
+eval_gemfile '../../shared.gemfile'
 gem 'rails', '~> 4.2.5'
 gem 'rails-ujs'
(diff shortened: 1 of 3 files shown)


timdiggins:
@glebm this looks like it's necessary for everything on travis now (because of bundler 1.16.0). But it means that it won't work on bundler <1.15.4

Weirdly getting some random db errors on travis though.

glebm:
> Weirdly getting some random db errors on travis though.

We started getting these recently from Travis but there have not been any changes that touch the database. I have no idea why.

timdiggins:
I have had a look -- it's because of race conditions with js specs.

although random it's always
 'User creates new topic' => 'scenario 'with title and content', js: true 

and the problem is caused by: 

```

def new_topic
    sign_in
    messageboard = create(:messageboard)
    PageObject::Topics.new(messageboard)
end
# then
scenario "something" do
  visit new_topic
end
```

Because new_topic is being created just before visiting it, there's a race.

so easy fix is to create new_topic **before** signing in. 

I think might be better if the fixtures are in `let!`s because we can then ensure all the lets are done before the scenario.
Of course there is then repetition of sign_in. But that's genearlly ok -- it's part of the scenario.
But I'm not 100% enamored of the DSL approach in the feature specs here -- it's a lot to wade through, and I'd prefer just to read the raw capybara DSL cause I then don't have to deal with  constructs from different DSLs (capybara vs thredded-features). However this is a separate discussion!. Will push a separate fix in a PR

PR #645, opened 2017-11-14
Make @mention dropdown not match when trailing space.

### LINKED ISSUE #643: Errors with @mention selector (users without spaces)
If you use the @ username selector and you select a username **without** a space (NB takes some work with the demo data to do this, as only Joe (default log in) has no spaces in his name -- so login as some other user, and then type `@Joe` in the post message), then:

* when you click the username or press return the drop down does not disappear.
* if you click or press return a second time, then the javascript hangs locking that page.

timdiggins:
Otherwise you don't get rid of the dropdown when you select a user who
doesn't have a space in it. fixes #643
It then prevents tripping into a bug
in textcomplete: https://github.com/yuku-t/textcomplete/issues/115
(because when `@Joe ` is inserted without this patch then it matches again
and then the reinsertion if you select again will be an identical
replacement and then js hangs (haven't found out why).

### DIFF
--- a/app/assets/javascripts/thredded/components/mention_autocompletion.es6
+++ b/app/assets/javascripts/thredded/components/mention_autocompletion.es6
@@ -1,7 +1,8 @@
 //= require thredded/components/user_textcomplete
 
 const ThreddedMentionAutocompletion = {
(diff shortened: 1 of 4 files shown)


timdiggins:
I'm merging this as the only failure was a read timeout for web-content (rich content smoke test)

PR #646, opened 2017-11-14
Fix random travis errors due to js race conditions

timdiggins:


timdiggins:
@glebm see discussion in previous PR https://github.com/thredded/thredded/pull/644#issuecomment-344320488

timdiggins:
Nope. I got this wrong. Still working on fixing this. it's very odd and looks like a leftover from a previous test

timdiggins:
Okay, I think I've got it -- the error comes from the previous js test which is continuing on and then failing (race between end of browsing and database cleaning) -- the error doesn't surface until the next js test. I've separated out the non-js and js tests to make this clearer (I think we should keep that): https://travis-ci.org/thredded/thredded/jobs/302118809

Now to fix it...

timdiggins:
The premise of this PR is wrong. Closing and moving discussion to https://github.com/thredded/thredded/issues/648. See also https://stackoverflow.com/questions/47303275/server-error-from-previous-capybara-js-test-surfacing-in-next-test

Issue #648, opened 2017-11-15
Fix travis error with capybara js


timdiggins:
Initially this looked random, but it is deterministic and weird.

Have written this up fully: https://stackoverflow.com/questions/47303275/server-error-from-previous-capybara-js-test-surfacing-in-next-test

timdiggins:
Fixed by https://github.com/thredded/thredded/commit/3088bb68f62fb051b8fc94b12d79b277c7317772 (updating chromedriver)

PR #649, opened 2017-11-15
[WIP] Try to elucidate/fix travis js failures

timdiggins:
https://github.com/thredded/thredded/issues/648


glebm:
I wonder if `post.start_quote` actually clicks "Quote" twice somehow.
Perhaps because [`open_post_actions`](https://github.com/thredded/thredded/blob/00a0173ef5e9e2c2387d1e97bdeb07adaf5b5982/spec/support/features/page_object/post.rb#L52) performs an actual click, and there is a slight overlap here:

quote

glebm:
@timdiggins Yep that seems to be it! #651
Thank you for doing all the work to elucidate the error, could not have figured it out otherwise!

glebm:
Nope, spoke too soon, that wasn't it :(

PR #664, opened 2017-12-06
Namespace MessageBoard in post.rb

tiii:


jayroh:
Was this causing a problem somewhere?

tiii:
Not in my case - but for consistency reasons I would suggest you prefix all thredded models. And it will cause trouble if you have an own Messageboard model in your app.

tiii:
I guess i'll close this again, since I can't reproduce it.

jayroh:
Nah I think it's still worth pulling in. I asked out of curiosity in the off chance it was causing issues 😁

jayroh:
Feel free to reopen if you'd like

tiii:
I just realized the other relations have different names and rails does namespaced model lookup in relations. No need to change this :ok_hand:

jayroh:
Cool. Thanks for the added context :)

On Wed, Dec 6, 2017 at 9:07 AM, Titus Ramczykowski <notifications@github.com

PR #666, opened 2017-12-06
Add german locales

tiii:
Now (finally) I got the time to open the PR for german locales. They are far from perfect, but it's better than nothing I guess. :wink:

PS: While translating I realized some inconsistensies with the translations of private topics. Sometimes they are translated as ```private topic``` other times they are translated as ```private conversations```.


jayroh:
Niiiiiice

PR #673, opened 2018-01-10
Allow disabling user links

metalelf0:
This commit changes the user link partial to render a span element
instead of a link when the user_path lambda is evaluated to nil.

Therefor, it allows disabling all user links by setting the `user_path`
lambda to always return nil.

This can be useful if, for any reason, you only want to link to some
user profiles (just write the lambda so that it returns only valid
links, and nil otherwise), or if you want to completely disable the
profile link feature.

### DIFF
--- a/spec/views/thredded/user/link_spec.rb
+++ b/spec/views/thredded/user/link_spec.rb
@@ -29,7 +29,7 @@ RSpec.describe 'partial: thredded/users/link' do
     HTML
   end
 
-  it 'with nil user path' do
+  it 'with nil user path', thredded_reset: [:@@user_path] do
     Thredded.user_path = ->(_user) { nil }
     render_partial build_stubbed(:user, id: 5, name: 'joel')
     expect(rendered).to eq <<-HTML.strip_heredoc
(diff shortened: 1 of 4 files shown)


metalelf0:
@glebm : I suppose specs are failing because of this example I added in `spec/views/thredded/user/link_spec.rb`:

```
  it 'with nil user path' do
    Thredded.user_path = ->(_user) { nil }
    render_partial build_stubbed(:user, id: 5, name: 'joel')
    expect(rendered).to eq <<-HTML.strip_heredoc
      <span class='thredded--username'>joel</span>
    HTML
  end
```

This has the side effect of changing the `Thredded.user_path` lambda for the rest of the spec suite. I tried using the `with_thredded_setting` around block as used elsewhere but it's not working for this specific setting. Any hint about how I can solve this?

spec/views/thredded/user/link_spec.rb
```diff
@@ -28,4 +28,12 @@ def render_partial(user)
       <em>#{I18n.t('thredded.null_user_name')}</em>
     HTML
   end
+
+  it 'with nil user path' do
```
glebm: How about `it 'with nil user path', thredded_reset: [:@@user_path] do`?

metalelf0:
@glebm thanks, that worked :)

jayroh:
ha - I didn't even know `thredded_reset` existed! Good to know

PR #683, opened 2018-02-21
Protect against onebox errors

### LINKED ISSUE #682: Rarish bugs in onebox (can bring down demo)
and in testing it I broke the demo. Will remind myself how to reseed heroku, and fix it.

The bug (which I haven't validated without thredded yet) is when onebox is trying to wrap a pdf which has a broken link.

The bug occurs with the following URL, which is NOT broken, but whose https:// version **is** broken. So it looks like onebox is trying to grab a different url (which is a bug in itself) -- probably only when served from https -- it seems like onebox (or is it possibly thredded) is maybe trying to deal with mixed content errors, but in a dangerous way...

http://www.poettker.com/files/pdf/Produktkatalog_170912.pdf
vs
https://www.poettker.com/files/pdf/Produktkatalog_170912.pdf

Traceback (from thredded v0.13.8): https://gist.github.com/timdiggins/3eba95ba752092e3648b9a3a932df41b

timdiggins:
Fixes #682 

@glebm Are we ok to use Rails.logger ? Mine was first example of it (in webapps I'd use some kind of exception notifier (Rollbar or Bugsnag or whatever, but Rails.logger is our best bet I presume for the engine)

### DIFF
--- a/lib/thredded/html_pipeline/onebox_filter.rb
+++ b/lib/thredded/html_pipeline/onebox_filter.rb
@@ -75,7 +75,7 @@ module Thredded
           preview.to_s.strip
         end
       rescue => e
-        puts "Onebox error for #{url}: #{e} "
+        Rails.logger.error("Onebox error for #{url}: #{e}")
         url
       end
 
--- a/lib/thredded/html_pipeline/onebox_filter.rb
+++ b/lib/thredded/html_pipeline/onebox_filter.rb
@@ -74,7 +74,7 @@ module Thredded
         else
           preview.to_s.strip
         end
-      rescue => e
+      rescue StandardError => e
         Rails.logger.error("Onebox error for #{url}: #{e}")
         <<~HTML
           <p><a href="#{ERB::Util.html_escape(url)}" target="_blank" rel="nofollow noopener">#{ERB::Util.html_escape(url)}</p>
--- a/lib/thredded/html_pipeline/onebox_filter.rb
+++ b/lib/thredded/html_pipeline/onebox_filter.rb
@@ -74,6 +74,9 @@ module Thredded
         else
           preview.to_s.strip
         end
+      rescue => e
+        puts "Onebox error for #{url}: #{e} "
+        url
       end
 
       def onebox_options(_url)
--- a/lib/thredded/html_pipeline/onebox_filter.rb
+++ b/lib/thredded/html_pipeline/onebox_filter.rb
@@ -76,7 +76,9 @@ module Thredded
         end
       rescue => e
         Rails.logger.error("Onebox error for #{url}: #{e}")
-        url
+        <<~HTML
+          <p><a href="#{ERB::Util.html_escape(url)}" target="_blank" rel="nofollow noopener">#{ERB::Util.html_escape(url)}</p>
+        HTML
       end
 
       def onebox_options(_url)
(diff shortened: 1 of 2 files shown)


timdiggins:
re: codeclimate I'm deliberately rescuing everything because we're trying to guard against Onebox doing bad stuff. it's a RuntimeError right now, I could just catch that?. or maybe we should be testing that certain known-good oneboxes are still working (by stopping caching on a few of our tests).

lib/thredded/html_pipeline/onebox_filter.rb
```diff
@@ -74,6 +74,9 @@ def render_onebox(url)
         else
           preview.to_s.strip
         end
+      rescue => e
+        Rails.logger.error("Onebox error for #{url}: #{e}")
+        url
```
glebm: We should render a link here:

```ruby
<<~HTML
  <p><a href="#{ERB::Util.html_escape(url)}" target="_blank" rel="nofollow noopener">#{ERB::Util.html_escape(url)}</p>
HTML
```

spec/lib/thredded/content_formatter_spec.rb
```diff
@@ -108,6 +108,28 @@ def parsed_html(html)
       end
     end
 
+    context 'with no onebox caching' do
+      around do |example|
+        onebox_views_cache = Thredded::HtmlPipeline::OneboxFilter.onebox_views_cache
+        onebox_data_cache = Thredded::HtmlPipeline::OneboxFilter.onebox_data_cache
+        Thredded::HtmlPipeline::OneboxFilter.onebox_views_cache = ActiveSupport::Cache::MemoryStore.new
+        Thredded::HtmlPipeline::OneboxFilter.onebox_data_cache = ActiveSupport::Cache::MemoryStore.new
+        example.run
```
glebm: I'm not sure whether the code after this line runs in case of a test failure. It might be necessary to use `ensure` here.

lib/thredded/html_pipeline/onebox_filter.rb
```diff
@@ -74,6 +74,9 @@ def render_onebox(url)
         else
           preview.to_s.strip
         end
+      rescue => e
```
glebm: This implicitly rescues `StandardError`. Perhaps doing it explicitly will fix the CodeClimate / Rubocop warning:

```ruby
rescue StandardError => e
```

Otherwise we can just `# rubocop:disable` it.

spec/lib/thredded/content_formatter_spec.rb
timdiggins: Nope. it's part of the `around` guarantee to trap exceptions. https://relishapp.com/rspec/rspec-core/v/3-7/docs/hooks/around-hooks#an-around-hook-continues-to-run-even-if-the-example-throws-an-exception

lib/thredded/html_pipeline/onebox_filter.rb
timdiggins: ✔️

glebm:
`Rails.logger` is fine by the way. There is no generic mechanism to notify via an exception notifier in Rails yet that I know of.

timdiggins:
PS: This is possibly not exciting enough to merit attention in our Changelog?

glebm:
I'll add it to the changelog under Fixes in the next release, because errors from Onebox may be a common issue (I had some of those too for the Amazon onebox and for the Twitter one).

timdiggins:
ugh sorry I merged when it was red in travis. how did I miss that. fixing now.

PR #688, opened 2018-03-16
test case: User loaded before i18n

lukesarnacki:
This PR adds failing test scenario which shows that `User` class is loaded before `I18n` is loaded.

### DIFF
--- a/spec/dummy/app/models/user.rb
+++ b/spec/dummy/app/models/user.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class User < ActiveRecord::Base
-  validates :name, presence: true
+  validates :name, presence: { message: I18n.t('presence_validation_message', locale: :en) }
 
   def to_s
     fail 'Deliberately failing so we can test'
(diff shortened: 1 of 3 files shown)


glebm:
That's not exactly a bug. Even if I18n was available at that point, the code you've posted is wrong, because it will be evaluated only once at startup, outside of the context of the current locale in the request.

lukesarnacki:
@glebm I knew I should have done more complex example here ;)

Like I wrote in the issue, I know this is not how you would write the code, but it was the simplest way to show this.

In my particular case, I have code which uses locales passed explicitly like so:

```ruby
I18n.available_locales.map do |locale|
  I18n.t(:foo, locale)
end
```

This is then used in `inclusion` validation. I wanted to point out general problem, not that this particular test scenario fails.

lukesarnacki:
@glebm I changed my test scenario to have `locale: :en` passed explicitly - maybe this will be less confusing.

lukesarnacki:
I will close this as I think it cannot be changed without changing the way thredded setup is done.

PR #695, opened 2018-04-30
Jump to the first unread post when navigating to a topic

glebm:
Removes the page column from read states and instead calculates the first unread page and the last read page dynamically in a single monster query that nevertheless is fast when scoped to a user and a list of topics.

Takes into account post policy scopes when calculating page information.

### DIFF
--- a/app/models/thredded/null_user_topic_read_state.rb
+++ b/app/models/thredded/null_user_topic_read_state.rb
@@ -13,5 +13,13 @@ module Thredded
     def post_read?(_post)
       false
     end
+
+    def first_unread_post_page
+      nil
+    end
+
+    def last_read_post_page
+      1
+    end
   end
 end
--- a/app/views/thredded/posts_common/_before_first_unread_post.html.erb
+++ b/app/views/thredded/posts_common/_before_first_unread_post.html.erb
@@ -1,0 +1,7 @@
+<div class="thredded--before-first-unread-post<%= ' unread--before-first-unread-post--first-in-page' if post.first_in_page? %>">
+  <%= view_hooks.posts_common.before_first_unread_post.render(self, post: post) do %>
+    <% unless post.first_in_page? %>
+      <span id="unread"></span>
+    <% end %>
+  <% end %>
+</div>
--- a/spec/models/thredded/user_topic_read_state_spec.rb
+++ b/spec/models/thredded/user_topic_read_state_spec.rb
@@ -4,12 +4,7 @@ require 'spec_helper'
 
 module Thredded
   describe UserTopicReadState, '#post_read?(post)' do
-    let(:read_state) do
-      create(:user_topic_read_state,
-             postable: create(:topic),
-             user: create(:user),
-             read_at: 1.day.ago)
-    end
+    let(:read_state) { create(:user_topic_read_state, read_at: 1.day.ago) }
 
     it 'is true when post.created_at > read_at' do
       post = create(:post, created_at: 2.days.ago)
@@ -24,6 +19,34 @@ module Thredded
     it 'is false when post.created_at < read_at' do
       post = create(:post, created_at: 1.minute.ago)
       expect(read_state.post_read?(post)).to be_falsey
+    end
+  end
+
+  describe UserTopicReadState, '.include_first_unread' do
+    let(:topic) { create(:topic, with_posts: 5) }
+    let(:posts) { topic.posts.to_a.sort_by(&:created_at) }
+
+    it 'returns the first unread post page and the last read page' do
+      create(:user_topic_read_state, postable: topic, read_at: posts[1].created_at)
+      expect(UserTopicReadState.with_page_info(posts_per_page: 3).to_a)
+        .to(contain_exactly(have_attributes(first_unread_post_page: 1, last_read_post_page: 1)))
+      expect(UserTopicReadState.with_page_info(posts_per_page: 2).to_a)
+        .to(contain_exactly(have_attributes(first_unread_post_page: 2, last_read_post_page: 1)))
+      expect(UserTopicReadState.with_page_info(posts_per_page: 1).to_a)
+        .to(contain_exactly(have_attributes(first_unread_post_page: 3, last_read_post_page: 2)))
+    end
+
+    it 'page info when there are no unread posts' do
+      create(:user_topic_read_state, postable: topic, read_at: posts[-1].created_at)
+      expect(UserTopicReadState.with_page_info(posts_per_page: 3).to_a)
+        .to(contain_exactly(have_attributes(first_unread_post_page: nil, last_read_post_page: 2)))
+    end
+
+    it 'respects the given posts_scope' do
+      create(:user_topic_read_state, postable: topic, read_at: posts[0].created_at)
+      posts_scope = Thredded::Post.where.not(id: posts[1].id)
+      expect(UserTopicReadState.with_page_info(posts_per_page: 1, posts_scope: posts_scope).to_a)
+        .to(contain_exactly(have_attributes(first_unread_post_page: 2, last_read_post_page: 1)))
     end
   end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
   Spoilers are focusable and are activated on mousedown, spacebar, or enter. They can also be nested.
   Markup is configurable via `Thredded::SpoilerTagFilter.spoiler_tags`.
   [#701](https://github.com/thredded/thredded/pull/701)
+* Jump to the first unread post when navigating to a topic.
+  [#695](https://github.com/thredded/thredded/pull/695)
 
 ## Fixed
 
@@ -21,7 +23,7 @@
 * Post IP tracking removed from core because it requires explicit consent under GDPR.
   [#705](https://github.com/thredded/thredded/pull/705)
 
-**NB: If updating to this version from 0.14.x, you **must** copy and run the upgrade migration after updating the gem:
+**NB**: If updating to this version from 0.14.x, you **must** copy and run the upgrade migration after updating the gem:
 
 ```console
 cp "$(bundle show thredded)"/db/upgrade_migrations/20180110200009_upgrade_thredded_v0_14_to_v0_15.rb db/migrate
--- a/app/commands/thredded/mark_all_read.rb
+++ b/app/commands/thredded/mark_all_read.rb
@@ -8,14 +8,8 @@ module Thredded
 
       unread_topics.each do |topic|
         last_post = topic.posts.order_oldest_first.last
-        total_pages = topic.posts.page(1).total_pages
 
-        Thredded::UserPrivateTopicReadState.touch!(
-          user.id,
-          topic.id,
-          last_post,
-          total_pages
-        )
+        Thredded::UserPrivateTopicReadState.touch!(user.id, topic.id, last_post)
       end
     end
   end
--- a/app/controllers/thredded/posts_controller.rb
+++ b/app/controllers/thredded/posts_controller.rb
@@ -55,8 +55,7 @@ module Thredded
 
     def mark_as_unread
       authorize post, :read?
-      page = post.page(user: thredded_current_user)
-      post.mark_as_unread(thredded_current_user, page)
+      post.mark_as_unread(thredded_current_user)
       after_mark_as_unread # customization hook
     end
 
--- a/app/controllers/thredded/private_posts_controller.rb
+++ b/app/controllers/thredded/private_posts_controller.rb
@@ -54,8 +54,7 @@ module Thredded
 
     def mark_as_unread
       authorize post, :read?
-      page = post.page
-      post.mark_as_unread(thredded_current_user, page)
+      post.mark_as_unread(thredded_current_user)
       after_mark_as_unread # customization hook
     end
 
--- a/app/controllers/thredded/private_topics_controller.rb
+++ b/app/controllers/thredded/private_topics_controller.rb
@@ -34,9 +34,7 @@ module Thredded
       @posts = Thredded::TopicPostsPageView.new(thredded_current_user, private_topic, page_scope)
 
       if thredded_signed_in?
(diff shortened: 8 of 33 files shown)


glebm:
`EXPLAIN` statements for the monster query:

Postgres:

```
EXPLAIN for: SELECT "read_states_with_first_unread".* FROM (SELECT "thredded_user_topic_read_states"."id", "unread_posts"."id" AS first_unread_post_id, COUNT("read_posts"."id") / 3 + 1 AS first_unread_post_page FROM "thredded_user_topic_read_states" INNER JOIN "thredded_topics" ON "thredded_topics"."id" = "thredded_user_topic_read_states"."postable_id" LEFT OUTER JOIN (SELECT "thredded_posts".* FROM "thredded_posts") "unread_posts" ON "unread_posts"."postable_id" = "thredded_topics"."id" AND "unread_posts"."created_at" > "thredded_user_topic_read_states"."read_at" LEFT OUTER JOIN (SELECT "thredded_posts".* FROM "thredded_posts") "read_posts" ON "read_posts"."postable_id" = "thredded_topics"."id" AND "read_posts"."created_at" <= "thredded_user_topic_read_states"."read_at" INNER JOIN "thredded_user_topic_read_states" "id_and_first_unread" ON "thredded_user_topic_read_states"."id" = "id_and_first_unread"."id" GROUP BY "thredded_topics"."id", "unread_posts"."id", "unread_posts"."created_at", "thredded_user_topic_read_states"."id" ORDER BY "unread_posts"."created_at" DESC) read_states_with_first_unread
                                                                QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------
 Subquery Scan on read_states_with_first_unread  (cost=113.63..123.47 rows=246 width=24)
   ->  GroupAggregate  (cost=113.63..121.01 rows=246 width=40)
         Group Key: thredded_posts.created_at, thredded_topics.id, thredded_posts.id, thredded_user_topic_read_states.id
         ->  Sort  (cost=113.63..114.25 rows=246 width=40)
               Sort Key: thredded_posts.created_at, thredded_topics.id, thredded_posts.id, thredded_user_topic_read_states.id
               ->  Hash Left Join  (cost=66.50..103.86 rows=246 width=40)
                     Hash Cond: (thredded_topics.id = thredded_posts_1.postable_id)
                     Join Filter: (thredded_posts_1.created_at <= thredded_user_topic_read_states.read_at)
                     ->  Hash Left Join  (cost=54.92..88.29 rows=246 width=40)
                           Hash Cond: (thredded_topics.id = thredded_posts.postable_id)
                           Join Filter: (thredded_posts.created_at > thredded_user_topic_read_states.read_at)
                           ->  Hash Join  (cost=43.35..72.72 rows=246 width=24)
                                 Hash Cond: (id_and_first_unread.id = thredded_user_topic_read_states.id)
                                 ->  Seq Scan on thredded_user_topic_read_states id_and_first_unread  (cost=0.00..22.30 rows=1230 width=8)
                                 ->  Hash  (cost=40.27..40.27 rows=246 width=24)
                                       ->  Hash Join  (cost=10.90..40.27 rows=246 width=24)
                                             Hash Cond: (thredded_user_topic_read_states.postable_id = thredded_topics.id)
                                             ->  Seq Scan on thredded_user_topic_read_states  (cost=0.00..22.30 rows=1230 width=24)
                                             ->  Hash  (cost=10.40..10.40 rows=40 width=8)
                                                   ->  Seq Scan on thredded_topics  (cost=0.00..10.40 rows=40 width=8)
                           ->  Hash  (cost=10.70..10.70 rows=70 width=24)
                                 ->  Seq Scan on thredded_posts  (cost=0.00..10.70 rows=70 width=24)
                     ->  Hash  (cost=10.70..10.70 rows=70 width=24)
                           ->  Seq Scan on thredded_posts thredded_posts_1  (cost=0.00..10.70 rows=70 width=24)
(24 rows)
```

MySQL:

```
EXPLAIN for: SELECT `read_states_with_first_unread`.* FROM (SELECT `thredded_user_topic_read_states`.`id`, `unread_posts`.`id` AS first_unread_post_id, COUNT(`read_posts`.`id`) / 3 + 1 AS first_unread_post_page FROM `thredded_user_topic_read_states` INNER JOIN `thredded_topics` ON `thredded_topics`.`id` = `thredded_user_topic_read_states`.`postable_id` LEFT OUTER JOIN (SELECT `thredded_posts`.* FROM `thredded_posts`) `unread_posts` ON `unread_posts`.`postable_id` = `thredded_topics`.`id` AND `unread_posts`.`created_at` > `thredded_user_topic_read_states`.`read_at` LEFT OUTER JOIN (SELECT `thredded_posts`.* FROM `thredded_posts`) `read_posts` ON `read_posts`.`postable_id` = `thredded_topics`.`id` AND `read_posts`.`created_at` <= `thredded_user_topic_read_states`.`read_at` INNER JOIN `thredded_user_topic_read_states` `id_and_first_unread` ON `thredded_user_topic_read_states`.`id` = `id_and_first_unread`.`id` GROUP BY `thredded_topics`.`id`, `unread_posts`.`id`, `unread_posts`.`created_at`, `thredded_user_topic_read_states`.`id` ORDER BY `unread_posts`.`created_at` DESC) read_states_with_first_unread
+----+-------------+---------------------------------+------------+--------+-------------------------------------------------------------------------------------------+---------+---------+---------------------------------------------------------------+------+----------+----------------------------------------------------+
| id | select_type | table                           | partitions | type   | possible_keys                                                                             | key     | key_len | ref                                                           | rows | filtered | Extra                                              |
+----+-------------+---------------------------------+------------+--------+-------------------------------------------------------------------------------------------+---------+---------+---------------------------------------------------------------+------+----------+----------------------------------------------------+
|  1 | PRIMARY     | <derived2>                      | NULL       | ALL    | NULL                                                                                      | NULL    | NULL    | NULL                                                          |    9 |    100.0 | NULL                                               |
|  2 | DERIVED     | thredded_user_topic_read_states | NULL       | ALL    | PRIMARY                                                                                   | NULL    | NULL    | NULL                                                          |    1 |    100.0 | Using temporary; Using filesort                    |
|  2 | DERIVED     | thredded_topics                 | NULL       | eq_ref | PRIMARY                                                                                   | PRIMARY | 8       | thredded_gem_test.thredded_user_topic_read_states.postable_id |    1 |    100.0 | Using index                                        |
|  2 | DERIVED     | id_and_first_unread             | NULL       | eq_ref | PRIMARY                                                                                   | PRIMARY | 8       | thredded_gem_test.thredded_user_topic_read_states.id          |    1 |    100.0 | Using index                                        |
|  2 | DERIVED     | thredded_posts                  | NULL       | ALL    | index_thredded_posts_on_postable_id,index_thredded_posts_on_postable_id_and_postable_type | NULL    | NULL    | NULL                                                          |    3 |    100.0 | Using where; Using join buffer (Block Nested Loop) |
|  2 | DERIVED     | thredded_posts                  | NULL       | ALL    | index_thredded_posts_on_postable_id,index_thredded_posts_on_postable_id_and_postable_type | NULL    | NULL    | NULL                                                          |    3 |    100.0 | Using where; Using join buffer (Block Nested Loop) |
+----+-------------+---------------------------------+------------+--------+-------------------------------------------------------------------------------------------+---------+---------+---------------------------------------------------------------+------+----------+----------------------------------------------------+
```

timdiggins:
Nice work -- I've only skimmed. I'll need to brush up on my arel to be able to read the code... 

But, wouldn't it be much simpler to have an dom-id in the view which is (say) #unread? we know what id the last read one is, so we just jump to that dom-id on the last page of the topic.

That unread could even have a stylistic presence when there are posts after it (like a little thin muted line with muted UNREAD pulled-right above it. It would then help to make the "mark unread from here" much easier to understand. I had planned to do that, but I got distracted, then life got in the way. Of course when there are no unread posts, you'd want to not output this. However that's probably easy to do. 

However, maybe I've missed something about why you needed to do it this way. (thinking aloud: you'd have to work out what page the first unread post is on -- but we know what page the last read post is on, so we should be able to calculate (i think) whether that's at the end of the page.

PS as a minor point -- is is time to consider how long rails 4.2 should be supported in the core thredded?

timdiggins:
The other thing I like about #unread is that conceptually it is a permalink for 'what i need to read' and remains true (and not tied to state).

I'm not sure, but would there be mileage in having an action for a topic (a pseudo-page) called `unread` 

so that a link like: https://thredded.org/thredded/main-board/i-assume-you-need-something-desperately-main-board-topic/unread would just redirect to 
https://thredded.org/thredded/main-board/i-assume-you-need-something-desperately-main-board-topic/page-2#unread

and then if might be useful in some asynchronous contexts (e.g. emailing, but maybe also multiple tabs) where you want to resolve the the link at click-time, rather than write-time.

glebm:
> cause it might be a lot of work to do the other variant, which could be considered as an optimization

I'm gonna try to implement your suggestion before merging. It shouldn't be too much work.

Query-cost wise I think it's acceptable, and much small than the other stuff we do (e.g. Ruby code rendering views).

glebm:
@timdiggins Done!

New query is faster, and we now have a partial that is rendered before the first unread post. It currently simply contains a `<span id="unread"></span>` unless the post is also first in the current page.

The new query is:

```
SELECT
  "thredded_user_topic_read_states".*,
  id_and_first_unread."first_unread_post_page"
FROM "thredded_user_topic_read_states"
LEFT OUTER JOIN (
                  SELECT
                    "thredded_user_topic_read_states"."id",
                    COUNT("read_posts"."id") / 1 + 1 AS first_unread_post_page
                  FROM "thredded_user_topic_read_states"
                  INNER JOIN (
                               SELECT "thredded_posts".*
                               FROM "thredded_posts"
                               WHERE "thredded_posts"."id" != 2) "read_posts"
                    ON "read_posts"."postable_id" = "thredded_user_topic_read_states"."postable_id" AND
                       "read_posts"."created_at" <= "thredded_user_topic_read_states"."read_at"
                  INNER JOIN "thredded_topics"
                    ON "thredded_user_topic_read_states"."postable_id" = "thredded_topics"."id" AND
                       "thredded_topics"."last_post_at" > "thredded_user_topic_read_states"."read_at"
                  GROUP BY "thredded_user_topic_read_states"."id") id_and_first_unread
  ON "thredded_user_topic_read_states"."id" = id_and_first_unread."id"
```

Explain:

```
Hash Right Join  (cost=157.04..160.80 rows=97 width=44)
  Hash Cond: (thredded_user_topic_read_states_1.id = thredded_user_topic_read_states.id)
  ->  HashAggregate  (cost=153.86..155.31 rows=97 width=16)
        Group Key: thredded_user_topic_read_states_1.id
        ->  Hash Join  (cost=61.86..152.26 rows=320 width=16)
              Hash Cond: (thredded_posts.postable_id = thredded_user_topic_read_states_1.postable_id)
              Join Filter: (thredded_posts.created_at <= thredded_user_topic_read_states_1.read_at)
              ->  Seq Scan on thredded_posts  (cost=0.00..71.70 rows=1815 width=24)
                    Filter: (id <> 2)
              ->  Hash  (cost=61.46..61.46 rows=32 width=32)
                    ->  Hash Join  (cost=3.18..61.46 rows=32 width=32)
                          Hash Cond: (thredded_topics.id = thredded_user_topic_read_states_1.postable_id)
                          Join Filter: (thredded_topics.last_post_at > thredded_user_topic_read_states_1.read_at)
                          ->  Seq Scan on thredded_topics  (cost=0.00..56.61 rows=61 width=16)
                          ->  Hash  (cost=1.97..1.97 rows=97 width=24)
                                ->  Seq Scan on thredded_user_topic_read_states thredded_user_topic_read_states_1  (cost=0.00..1.97 rows=97 width=24)
  ->  Hash  (cost=1.97..1.97 rows=97 width=36)
        ->  Seq Scan on thredded_user_topic_read_states  (cost=0.00..1.97 rows=97 width=36)
```

glebm:
Now the query is only required to calculate the page correctly.

One alternative would be to re-calculate the `page` attribute of read state when a post is deleted or its moderation state changes. Perhaps this is a better approach.

app/views/thredded/posts_common/_before_first_unread_post.html.erb
```diff
@@ -0,0 +1,3 @@
+<% unless post.first_in_page? %>
+  <span id="unread"></span>
```
timdiggins: @glebm is there a reason we have to drop this if it's first_in_page? Maybe we could add a class if it's first in page? 

I can imagine wanting to style it differently if it's first in page, but I still would want to be able to have something visible to show that the page is unread. If you think of someone getting used to having a "unread" divider, they might then be expecting it, and searching for it if it's not present, causing confusion.

spec/models/thredded/user_topic_read_state_spec.rb
```diff
@@ -27,6 +22,34 @@ module Thredded
...
+      create(:user_topic_read_state, postable: topic, read_at: posts[1].created_at)
+      expect(UserTopicReadState.include_first_unread_page(posts_per_page: 3).to_a)
+        .to(contain_exactly(have_attributes(first_unread_post_page: 1)))
+      expect(UserTopicReadState.include_first_unread_page(posts_per_page: 2).to_a)
+        .to(contain_exactly(have_attributes(first_unread_post_page: 2)))
+      expect(UserTopicReadState.include_first_unread_page(posts_per_page: 1).to_a)
+        .to(contain_exactly(have_attributes(first_unread_post_page: 3)))
+    end
+
+    it 'returns nil post ID if there are no unread posts' do
```
timdiggins: nit: "it returns nil post **page**..."

app/models/thredded/null_user_topic_read_state.rb
```diff
@@ -13,5 +13,13 @@ def read?
     def post_read?(_post)
       false
     end
+
+    def first_unread_post_id
```
timdiggins: this is now redundant I think

glebm:
Comments addressed and:

1. Both first unread and last pages are now calculated dynamically.
2. The query is now very fast when scoped to a user and a list of topics (as is the case usually).

PR #696, opened 2018-05-08
Render posts concurrently

glebm:
Posts rendering sometimes involves network IO (e.g. for oneboxes), so uncached loads can be very slow.

This PR mitigates the issue by rendering the posts concurrently.

### DIFF
--- a/lib/thredded/collection_to_strings_with_cache_renderer.rb
+++ b/lib/thredded/collection_to_strings_with_cache_renderer.rb
@@ -1,15 +1,23 @@
 # frozen_string_literal: true
 
 require 'action_view/renderer/abstract_renderer'
+
 module Thredded
   class CollectionToStringsWithCacheRenderer < ActionView::AbstractRenderer
+    # The default number of threads to use for rendering.
+    mattr_accessor :render_threads
+    self.render_threads = 50
+
     # @param view_context
-    # @param collection [Array<T>]
-    # @param partial [String]
-    # @param expires_in [ActiveSupport::Duration]
+    # @param [Array<T>] collection
+    # @param [String] partial
+    # @param [ActiveSupport::Duration] expires_in
+    # @param [Integer] render_threads the number of threads to use for rendering. This is useful even on MRI ruby
+    #   for IO-bound operations.
+    # @param [Hash] locals
     # @return Array<[T, String]>
     def render_collection_to_strings_with_cache( # rubocop:disable Metrics/ParameterLists
-      view_context, collection:, partial:, expires_in:, locals: {}, **opts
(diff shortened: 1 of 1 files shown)


lib/thredded/collection_to_strings_with_cache_renderer.rb
```diff
@@ -52,12 +61,25 @@ def collection_cache
...
 
     # @return [Array<String>]
-    def render_partials(view_context, collection:, **opts)
+    def render_partials(view_context, collection:, render_threads:, **opts)
       return [] if collection.empty?
-      partial_renderer = ActionView::PartialRenderer.new(@lookup_context)
-      collection.map do |item|
-        partial_renderer.render(view_context, opts.merge(object: item), nil)
+      num_threads = [render_threads, collection.size].min
+      if num_threads == 1
```
timdiggins: is there an performance overhead to starting the Threads -- if so, it's plausible that this might want to be `if num_threads <= RENDER_THREADS_THRESHOLD` <- where that's a small number.

glebm: Starting even 50 threads is basically free on MRI ruby:

```ruby
require 'benchmark'
puts Benchmark.measure { Array.new(50) { |i| Thread.start { i*i } }.map(&:value) }
#=>  0.002865   0.005927   0.008792 (  0.004879)
```

While it may be different on JRuby, I doubt it's anywhere near the cost of rendering a post.

glebm:
> It might be worth having a test that just tests that you get the posts rendered in the correct order.

We already have plenty of tests that will fail if posts get rendered out-of-order, and the code is straightforward for now, so I lazily skipped on the tests.

> Are you sure this speeds it up? even in the (possibly pathological) case where there are no i/o dependent issues (e.g. onebox turned off)?

On MRI ruby, because of the GIL, it *doesn't* speed it up in the scenario where rendering is done purely in Ruby with no IO involved. However, the overhead of starting 50 threads is minimal. The speed up with even just a few oneboxes is **massive**.

On JRuby threads do run in parallel so there is speed up (assuming JRuby thread creation overhead is low enough).

> Do you know what the memory impact of having many threads is likely to be? (i don't really get how memory allocation happens in multithreaded ruby -- is it largely shared (copy on write) in which case there is very little memory required, or is it more complex) -- I presume the new thread's memory is reclaimed when it ends (and joins the main thread), but while it's going if you then encounter memory limits, that might be very undesirable in some platforms.

Well threads in Ruby are real threads, so they share memory directly (unlike forks which do use a COW mechanism due to how the OS performs forking). Any additional memory used by rendering should be no more than a few hundred KiB per thread (I haven't measured, this is just a guess).

> it's not clear how to override the number of threads -- you could redefine / predefine the constant in a monkey patch in your main app, 

Good point, I've made it a module attribute.

PR #701, opened 2018-05-27
Feature: Spoiler tags

glebm:
Spoiler tags can be added using `<spoiler></spoiler>` or `[spoiler][/spoiler]` markup (configurable via `Thredded::SpoilerTagFilter.spoiler_tags`).

Supported out of the box for any markup processor.
Spoilers are focusable and are activated on mousedown, spacebar, or enter.
They can also be nested.
Their accessibility role is "figure" and they have a dynamically updated `aria-expanded` attribute.

### DIFF
--- a/config/locales/zh-CN.yml
+++ b/config/locales/zh-CN.yml
@@ -126,6 +126,8 @@ zh-CN:
         update_btn_submitting: :thredded.form.update_btn_submitting
       pending_moderation_notice: 你的回复将在审核之后公开发表
       quote_btn: 引用
+      spoiler_summary: 剧透 - 点击显示。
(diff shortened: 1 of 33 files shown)


config/locales/zh-CN.yml
```diff
@@ -126,6 +126,8 @@ zh-CN:
         update_btn_submitting: :thredded.form.update_btn_submitting
       pending_moderation_notice: 你的回复将在审核之后公开发表
       quote_btn: 引用
+      spoiler_summary: 剧透 - 点击显示。
+      spoiler_summary_for_email: 扰流板 - 选择下面的内容来查看：
```
glebm: @moondcr These have been translated via Google Translate, PTAL.

PR #716, opened 2018-06-30
Send private message URL helper

### LINKED ISSUE #704: "Send private message" link (in demo and generated app)
Expose functionality from Thredded to make it easy to create a link that either goes to an existing private topic or a new private topic page.

At this point we should also introduce a `Thredded.private_messages_enabled` configuration option.

~~It would also be nice to have a link to start a new private topic with the same participants from an existing private topic.~~

glebm:
Adds a helper that returns a private message URL for the two given
users. If a thread already exists between the two users, returns the URL
to that thread.

Fixes #704

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -57,6 +57,9 @@ Style/CollectionMethods:
 Style/Documentation:
   Enabled: false
 
+Style/Lambda:
+  EnforcedStyle: literal
+
 Style/PredicateName:
   Exclude:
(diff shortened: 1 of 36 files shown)


ishields:
This is awesome! Exactly what I need.  Can't wait to get a new release for this feature.

ishields:
@glebm when is 15.4 schedule for release?  If it's a ways out, does it make sense to get a minor release before with this change?

PR #721, opened 2018-07-18
Translations for sign in / sign out.

descovi:


simonini:
Note:

I have problems with run test locally. 

When I _bundle_ I get an error because of pg (i don't use it in my mac).

Maybe solved with 

` bundle install --without development production`

But after i get another error when use rspec:

`require': cannot load such file -- rspec/core/formatters/progress_formatter`

Maybe the readme miss something?

Thanks 👍

glebm:
> I have problems with run test locally.

You will need the development headers for sqlite3, mysql, and postgres.
If you use homebrew, for postgresql I think it's simply:

```
brew install libpq
```

If that doesn't work, try:

```
brew install postgres
```

timdiggins:
thanks @descovi for this -- I've rebased and fixed the i18n-tasks complaint in #924

PR #736, opened 2018-10-01
Sass -> SassC

glebm:
The Ruby version of Sass is deprecated, see http://sass.logdown.com/posts/7081811


timdiggins:
@glebm warning that sassc-rails and sass-rails have some coexistance problems (particularly for libraries that explicitly `require sass-rails`.

One of these is rails-admin (of course) but there seems to be workarounds for this. Another is font-awesome-rails -- which in v5 version is using sassc (and not sassc-rails AFAICT) but in its v4 version is explicitly requiring sass-rails. And upgrading from v4-v5 is a bit of a faff.

glebm:
sassc-rails v2.0.0 is out, merging this, please report issues if any.

PR #746, opened 2018-10-12
Order users by name in autocomplete

### LINKED ISSUE #744: User autocompletion should sort
When autocompleting users for a drop down, I think it should sort them 

private_messages___thredded_demo

main_board___thredded_demo

timdiggins:
closes #744

### DIFF
--- a/app/controllers/thredded/autocomplete_users_controller.rb
+++ b/app/controllers/thredded/autocomplete_users_controller.rb
@@ -30,6 +30,7 @@ module Thredded
       if query.length >= Thredded.autocomplete_min_length
         DbTextSearch::CaseInsensitive.new(users_scope, Thredded.user_name_column).prefix(query)
           .where.not(id: thredded_current_user.id)
+          .order(Thredded.user_class.arel_table[Thredded.user_name_column].lower.asc)
           .limit(MAX_RESULTS)
       else
         []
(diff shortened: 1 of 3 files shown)


app/controllers/thredded/autocomplete_users_controller.rb
```diff
@@ -30,6 +30,7 @@ def users_by_prefix
       if query.length >= Thredded.autocomplete_min_length
         DbTextSearch::CaseInsensitive.new(users_scope, Thredded.user_name_column).prefix(query)
           .where.not(id: thredded_current_user.id)
+          .order(Thredded.user_class.arel_table[Thredded.user_name_column].lower.asc)
```
glebm: We'll get better results if we use the appropriate case-insensitive method for each database, especially for things like the German "ß", the Turkish "ı", etc.

Let's add a method to `db_text_search` instead?
Using it could look something like this:

```ruby
case_insensitive = DbTextSearch::CaseInsensitive.new(users_scope, Thredded.user_name_column)
case_insensitive.prefix(query)
  .where.not(id: thredded_current_user.id)
  .order(case_insensitive.column_for_order(:asc))
```

timdiggins: https://github.com/thredded/db_text_search/pull/5

PR #751, opened 2018-10-13
Add follow icon to unread topics count.

timdiggins:
Also allow for easier icon management with `define_svg_icons` and `inline_svg_once`.
More can be done with those (e.g. creating a helper to both define and then use an icon).

### DIFF
--- a/app/assets/stylesheets/thredded/layout/_user-navigation.scss
+++ b/app/assets/stylesheets/thredded/layout/_user-navigation.scss
@@ -20,3 +20,10 @@
 .thredded--user-navigation--unread-topics--followed-count {
   @extend %thredded--nav-tabs--item--badge;
 }
+.thredded--unread-topics--followed-icon {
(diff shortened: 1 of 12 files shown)


timdiggins:
I think it looks better. 
unread-count-has-icon

It makes me think it might be easier to read at a glance if both moderation and private messages had icons of their own. private messages uses an envelope icon sometimes I think. Not sure about moderation.

timdiggins:
@glebm do you know why the `title: nil` were on the `inline_svg` in definitions. I can't see any reason/rendered difference.

glebm:
> @glebm do you know why the title: nil were on the inline_svg in definitions. I can't see any reason/rendered difference.

No reason, I copy-pasted this from:
https://github.com/thredded/thredded/blob/dab4abc701f9e3b75f9b566b3e769c2401a10eeb/app/views/thredded/topics/index.html.erb#L7

PR #752, opened 2018-10-13
create and use `svg_icon` helper which defines icon if needed and then uses it.

timdiggins:
Issues I've just thought of:

* will this mess with caching (definition should probably be hoiked outside of cachable content)
* is "svg_icon" the right name for this.

### DIFF
--- a/app/helpers/thredded/icon_helper.rb
+++ b/app/helpers/thredded/icon_helper.rb
@@ -2,6 +2,17 @@
 
 module Thredded
   module IconHelper
+    def svg_icon(filename, **args)
+      svg = content_tag :svg, **args do
+        content_tag :use, '', 'xlink:href' => "##{thredded_icon_id(filename)}"
+      end
+      if (definition = define_svg_icons(filename))
+        definition + svg
+      else
+        svg
+      end
+    end
+
     def define_svg_icons(*filenames)
       return if filenames.blank?
       sb = filenames.map do |filename|
--- a/app/helpers/thredded/icon_helper.rb
+++ b/app/helpers/thredded/icon_helper.rb
@@ -2,7 +2,7 @@
 
 module Thredded
   module IconHelper
-    def svg_icon(filename, **args)
+    def shared_inline_svg(filename, **args)
       svg = content_tag :svg, **args do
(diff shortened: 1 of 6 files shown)


app/helpers/thredded/icon_helper.rb
```diff
@@ -2,6 +2,17 @@
 
 module Thredded
   module IconHelper
+    def svg_icon(filename, **args)
```
glebm: `shared_inline_svg`?

glebm:
> will this mess with caching (definition should probably be hoiked outside of cachable content)

It would, but we don't cache anything except post contents and I don't think we will.

app/helpers/thredded/icon_helper.rb
timdiggins: Yes good. I’ll do that then merge this into the other branch

timdiggins: Yes good. I’ll do that then merge this into the other branch

jayroh:
There are several rails helper gems that provide a helper like this I believe. That's not to say we shouldn't or couldn't have our own - but to warn against any possible clashes between method names.

I'm in my phone currently so I'm not able to do a lot of digging but a look through Ruby toolbox for the more prevailantly used gems might be a good guard against that

timdiggins:
@jayroh I've looked at https://www.ruby-toolbox.com/search?utf8=%E2%9C%93&q=svg and the only thing in the same space I can find is inline_svg, which this is wrapping

Also, FYI I looked into our svgs inline but in a separate (browser-cachable) document https://css-tricks.com/svg-use-with-external-reference-take-2/. This helper puts us on the way to doing that, but unfortunately because svg use doesn't support CORS, the document with the svgs in it would have to be served by the same host as the web page that "uses" them  -- which would exempt CDNs, and might thus cause more problems than it solves. (but i still could imagine this working -- but we'd want to link to a compiled (and digested) asset (probably an html file), but using the current host rather than the asset_url. I don't know whether that's possible, but in any case out of scope for the thredded project per se).

glebm:
> Also, FYI I looked into our svgs inline but in a separate (browser-cachable) document

Another issue is that it would cause flash of unstyled content (missing SVGs) on the first load (IMO for no good reason given how tiny these SVGs are).

PR #753, opened 2018-10-13
Can specify next page when creating new topic.

### LINKED ISSUE #619: Configure next page after creating a topic
In the interconnection between thredded and a main app, I want to be able to configure what the TopicsController sends you to after creating the topic.

Currently it always goes to the messageboard the topic is created in. In some usages I want the next page to be the recently created topic. In other usages I want the page to be a specific url (often but not always the page before the new topic form). 

I think the best way to do this is to add a hidden field called something like `next_page`. This can be either: 
* a path beginning with / (not a URL -- too many security concerns there)
* the string "topic"
* the string "messageboard" (or empty string <-- the default) 

The new topic form also takes an optional param `next_page` which populates this hidden field. 

Then Topics#create redirects to this param

I'm planning on working on this in next day or so. Shout if any thoughts @glebm or feel free to just throw in your points when we get to the PR

timdiggins:
Pass next_page param with new_topic_url. 

`next_page` can be:

*  "messageboard" (or empty same as current behaviour) to display messageboard topics after creating topic
*  "topic" to display topic after creating topic
* a path starting with / to redirect to this specific path. 

closes #619.

### DIFF
--- a/app/controllers/thredded/topics_controller.rb
+++ b/app/controllers/thredded/topics_controller.rb
@@ -88,7 +88,7 @@ module Thredded
       @new_topic = Thredded::TopicForm.new(new_topic_params)
       authorize_creating @new_topic.topic
       if @new_topic.save
-        redirect_to messageboard_topics_path(messageboard)
+        redirect_to next_page_after_create(params[:next_page])
       else
         render :new
       end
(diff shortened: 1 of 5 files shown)


app/controllers/thredded/topics_controller.rb
```diff
@@ -141,6 +141,19 @@ def unfollow
 
     private
 
+    def next_page_after_create(next_page)
+      case next_page
+      when 'messageboard', '', nil
+        return messageboard_topics_path(messageboard)
+      when 'topic'
+        messageboard_topic_path(messageboard, @new_topic.topic)
+      when %r{\A/[^/]\S+\Z}
```
glebm: Should be `\z` (`\Z` is something like `\n?\z`)

PR #763, opened 2018-10-28
json actions for mark_as_read and mark_as_unread for post

timdiggins:
There are some "api" endpoints (for ajax calls from mainapp or extensions) that I need and think maybe other people might need, so I'm could pull these into Thredded core.

However this may be unneeded.

However I also think (as I've illustrated in this PR) that we ought to not use the hierarchical routes unless they are being shown to the user / search engines (ie. GET routes to display things). This would make the routes file a bit easier to navigate and make the url-generation functions and paths more straightforward

ie.
```
mark_as_unread_post_path(post)
#  /posts/:id/mark_as_unread(.:format)
```

rather than
```
mark_as_unread_messageboard_topic_post_path(post.messageboard, post.postable, post)
# /:messageboard_id/:topic_id/:id/mark_as_unread(.:format)  
```

Unless there are objections, we can start moving the paths across (and changing the helpers to point to the new paths), keeping the old more complex paths till the next minor bump when we can drop them.

### DIFF
--- a/app/controllers/thredded/posts_controller.rb
+++ b/app/controllers/thredded/posts_controller.rb
@@ -7,6 +7,7 @@ module Thredded
     include Thredded::NewPostParams
 
     helper_method :topic
+    before_action :assign_messageboard_for_actions, only: %i[mark_as_read mark_as_unread]
     after_action :update_user_activity
 
     after_action :verify_authorized
@@ -94,6 +95,10 @@ module Thredded
       Thredded::Topic
         .where(messageboard: messageboard)
         .friendly_find!(params[:topic_id])
+    end
+
+    def assign_messageboard_for_actions
+      @messageboard = post.postable.messageboard
     end
 
     def post
--- a/app/controllers/thredded/posts_controller.rb
+++ b/app/controllers/thredded/posts_controller.rb
@@ -58,7 +58,7 @@ module Thredded
       authorize post, :read?
       UserTopicReadState.touch!(thredded_current_user.id, post)
       respond_to do |format|
-        format.html { redirect_to request.referer || post_path(post, user: thredded_current_user) }
+        format.html { redirect_back fallback_location: post_path(post, user: thredded_current_user) }
         format.json { render(json: { read: true }) }
       end
     end
--- a/app/controllers/thredded/posts_controller.rb
+++ b/app/controllers/thredded/posts_controller.rb
@@ -53,10 +53,22 @@ module Thredded
                     notice: I18n.t('thredded.posts.deleted_notice')
     end
 
+    def mark_as_read
+      authorize post, :read?
+      UserTopicReadState.touch!(thredded_current_user.id, post)
+      respond_to do |format|
+        format.html { redirect_to request.referer || post_path(post, user: thredded_current_user) }
+        format.json { render(json: { read: true }) }
+      end
+    end
+
     def mark_as_unread
       authorize post, :read?
       post.mark_as_unread(thredded_current_user)
-      after_mark_as_unread # customization hook
+      respond_to do |format|
+        format.html { after_mark_as_unread } # customization hook
+        format.json { render(json: { read: false }) }
+      end
     end
 
     def quote
--- a/app/controllers/thredded/private_posts_controller.rb
+++ b/app/controllers/thredded/private_posts_controller.rb
@@ -51,10 +51,22 @@ module Thredded
                     notice: I18n.t('thredded.posts.deleted_notice')
     end
 
+    def mark_as_read
+      authorize post, :read?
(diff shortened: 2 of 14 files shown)


glebm:
For HTML routes, having deep routes allows us to show the correct navigation on "Not Found" (e.g. if a topic wasn't found, at least we indicate which messageboard it is).

For JSON / action type routes, I agree that deep routes are unnecessary.

config/routes.rb
```diff
@@ -84,11 +84,19 @@
...
           get 'quote'
-          post 'mark_as_unread'
+          post 'mark_as_unread' # deprecated
         end
       end
     end
   end
 
+  # flat urls for anything which is non-visible to users & search engines
+  resources :posts, only: %i[] do
```
glebm: Alternatively, we could have this block under `resources :topics` but with `shallow: true` (https://edgeguides.rubyonrails.org/routing.html#shallow-nesting).

Also, we could scope all such routes under e.g. `/action/` so that they are easy to block with `robots.txt`.

config/routes.rb
```diff
@@ -84,11 +84,19 @@
         resource :preview, only: [:update], controller: 'post_previews'
         member do
           get 'quote'
-          post 'mark_as_unread'
+          post 'mark_as_unread' # deprecated
```
glebm: Perhaps we can just remove it, because most users would be using this method on PostView to generate the URL:

https://github.com/thredded/thredded/blob/235d0a9026c9f8248297c0fca13e655aca893a75/app/view_models/thredded/post_view.rb#L62

config/routes.rb
```diff
@@ -84,11 +84,19 @@
         resource :preview, only: [:update], controller: 'post_previews'
         member do
```
glebm: We should do this for private posts as well, for consistency.

config/routes.rb
timdiggins: I think it's a little easier to read and rationalize about without shallow, but this might be just a question of taste (I find our routes file rather heavy reading presently).

None of these routes will be `get` I think, so is the robots.txt blocking `/action` prefix really needed?

timdiggins: (thinking more). But there might be some clarity to having all of these routes under `action/` (in terms of logging, but also logical separation from the routes that can generate 404s) -- would then need add it to list of reserved words for messageboards (https://github.com/thredded/thredded/blob/master/app/models/thredded/messageboard.rb#L9).

glebm: > I think it's a little easier to read and rationalize about without shallow, but this might be just a question of taste (I find our routes file rather heavy reading presently).

I don't mind either way. If we want to scope them under `action/`, they'll have to be separate anyway.

> would then need add it to list of reserved words for messageboards 

Yes.

timdiggins:
@glebm I've covered private posts and have moved the paths under .../action.
 
I've started some release notes on this, but other than maybe needing to rename one of the messageboards (could possibly create a migration), it's not very interesting.

I'm not sure about codeclimate's recommendation to dry up posts and private posts controller. I think there are some benefits to keeping these separate but duplicated (at least for now). But maybe you can think of a way to dry this up without making it obscure.

I've also removed a redundant action https://github.com/thredded/thredded/pull/763/commits/d9d218c1b985613aa5a86aee97fc2edffdd0e4a9

app/controllers/thredded/posts_controller.rb
```diff
@@ -53,10 +54,22 @@ def destroy
                     notice: I18n.t('thredded.posts.deleted_notice')
     end
 
+    def mark_as_read
+      authorize post, :read?
+      UserTopicReadState.touch!(thredded_current_user.id, post)
+      respond_to do |format|
+        format.html { redirect_to request.referer || post_path(post, user: thredded_current_user) }
```
glebm: Please use:

```ruby
redirect_back fallback_location: post_path(post, user: thredded_current_user)
```

We have `redirect_back` polyfilled for Rails 4 here:

https://github.com/thredded/thredded/blob/235d0a9026c9f8248297c0fca13e655aca893a75/app/controllers/thredded/application_controller.rb#L66-L73

app/controllers/thredded/private_posts_controller.rb
```diff
@@ -52,10 +51,22 @@ def destroy
                     notice: I18n.t('thredded.posts.deleted_notice')
     end
 
+    def mark_as_read
+      authorize post, :read?
+      UserPrivateTopicReadState.touch!(thredded_current_user.id, post)
+      respond_to do |format|
+        format.html { redirect_to request.referer || post_path(post, user: thredded_current_user) }
```
glebm: `redirect_back` here as well

PR #768, opened 2018-11-17
Update MacOS installation instructions for chromedriver

cbortz:
```
Error: No available formula with the name "chromedriver"
It was migrated from homebrew/core to homebrew/cask.
You can access it again by running:
  brew tap homebrew/cask
And then you can install it by running:
  brew cask install chromedriver
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -628,7 +628,7 @@ On Mac, run:
 
 ```bash
 brew cask install chromium
-brew install chromedriver
+brew cask install chromedriver
 ```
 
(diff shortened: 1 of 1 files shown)


cbortz:
In the future, would y'all prefer the `[ci skip]` flag be included in the commit message for PRs like this one (where only the README is updated)? Or do you prefer to run the entire test suite regardless?

glebm:
@cbortz `[ci skip]` is fine for commits like this, no preference either way.

jayroh:
Thanks @cbortz!

PR #775, opened 2018-11-25
A view hook for customizing Topic title on topic#show

timdiggins:


timdiggins:
@glebm I think I might be the first person other than you to create a new ViewHook. I think I've done it right, but would be good to check that the idea is carried through correctly in terms of terminology etc.

I'll want to merge relatively soon anyway (tomorrow probably) but if you come up with any issues afterwards can address those in a new PR before release (not urgent for release)

PR #776, opened 2018-11-27
Add notes about usage of email_from in initializer and readme

timdiggins:


glebm:
Thank you! This came up a few times here and on gitter.

PR #782, opened 2018-12-16
Cache @-mentioned users when rendering posts

### LINKED ISSUE #771: More performant user provider for AtMentionFilter
As mentioned in #770, currently mentioned users are being fetched per post (and in parallel if `render_threads` > 1). There is a notion of a `user-provider`, but there's no way to currently provide them. Two ways to improve this from a performance point of view:

1) with parallel execution scan all posts in advance and aggregate users from posts and pre-fetch them within a shared UserProvider -- maybe this is a new step in the filters (prefetch). However this will mean an unnecessary scan and prefetch when the posts are already cached. However perhaps this is not such a big deal.

2) Alternatively (but this works with serial execution only - wouldn't work with render_threads > 1) don't provide in advance but have the same user provider across each post that fetches them when needed, and caches the results for future posts.  (This is quite a perf gain as within a thread the same users often get mentioned and rementioned).

glebm:
I went with caching instead of pre-scanning, because fully correct scanning would require us to run most of the content formatter (all the way up to `AtMentionFilter`), which can be slow.

Fixes #771

### DIFF
--- a/app/commands/thredded/at_notification_extractor.rb
+++ b/app/commands/thredded/at_notification_extractor.rb
@@ -16,7 +16,8 @@ module Thredded
       Thredded::HtmlPipeline::AtMentionFilter.new(
         html,
         view_context: view_context,
-        users_provider: ->(user_names) { @post.readers_from_user_names(user_names).to_a }
+        users_provider: ::Thredded::UsersProvider,
+        users_provider_scope: @post.readers
       ).mentioned_users
     end
   end
--- a/app/helpers/thredded/application_helper.rb
+++ b/app/helpers/thredded/application_helper.rb
@@ -85,7 +85,8 @@ module Thredded
     # @param content_partial [String]
     def render_posts(posts, partial: 'thredded/posts/post', content_partial: 'thredded/posts/content', locals: {})
       posts_with_contents = render_collection_to_strings_with_cache(
-        partial: content_partial, collection: posts, as: :post, expires_in: 1.week
+        partial: content_partial, collection: posts, as: :post, expires_in: 1.week,
+        locals: { options: { users_provider: ::Thredded::UsersProviderWithCache.new } }
       )
       render partial: partial, collection: posts_with_contents, as: :post_and_content, locals: locals
     end
--- a/app/models/concerns/thredded/post_common.rb
+++ b/app/models/concerns/thredded/post_common.rb
@@ -33,20 +33,14 @@ module Thredded
 
     # @param view_context [Object] the context of the rendering view.
     # @return [String] formatted and sanitized html-safe post content.
-    def filtered_content(view_context, users_provider: ->(names) { readers_from_user_names(names) }, **options)
-      Thredded::ContentFormatter.new(view_context, users_provider: users_provider, **options).format_content(content)
+    def filtered_content(view_context, users_provider: ::Thredded::UsersProvider, **options)
+      Thredded::ContentFormatter.new(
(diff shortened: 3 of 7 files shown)


timdiggins:
I like extracting out the separate `Thredded::UsersProvider` class and removing the `Post# readers_from_user_names` (which really doesn't belong there).

However, are you convinced this will actually speed things up? When I looked into this last, taking the out of the box situation where (I believe) the number of render threads is the same as the default number of posts per page... then the users would (mostly) be fetched in parallel anyway, meaning the user-cache would unlikely to be read very much.


Anyway, it might be good to be able to have a hook (or just instructions if it's a case of monkeypatching -- though I can't figure out how to do cleanly) to change the UserProvider implementation. (e.g. from default UserProviderWithCache to UserProvider ). 

----

More generally, (this is probably a separate question - and only marginally relevant to this PR) -- I don't know why we care so much about the **Post's** readers when displaying -- isn't it more the **Topic**'s readers, or in fact the Messageboard's readers -- because I think the Thredded model is that the Messageboard is the level at which read-permissions would be enforced. Some posts may be not readable by all of the messageboard's readers (due to premoderation settings) but that shouldn't change how the mentions are calculated for display purposes (though they would be for notification purposes - however the notification should probably only happen when the post is published anyway? Notification for the purposes of premoderation is a bit different to normal notification -- or should be).

Anyway if that was the case then we could actually do something like 
`users_provider = UsersProvider.new(messageboard)`
and then 
`users_provider.readers_for_user_names(user_names)`

But I don't think that would be needed for this PR at all -- I'm just trying to establish if my thinking regarding permissions are correct.

timdiggins:
>> However, are you convinced this will actually speed things up? When I looked into this last, taking the out of the box situation where (I believe) the number of render threads is the same as the default number of posts per page... then the users would (mostly) be fetched in parallel anyway, meaning the user-cache would unlikely to be read very much.

> All the threads share the same cache instance and access to the cache is behind a mutex, so a user is never fetched twice.

Ok -- I guess I was thinking that if the timing before fetching the users (in each thread) was relatively short (say 1 ms or 2), and the fetch was *relatively* long (say 5 ms), then the db accesses are likely to be in parallel. My assumptions may be wrong here, and/or my use case is different (quick renders, slower db).

timdiggins:
>> Anyway, it might be good to be able to have a hook (or just instructions if it's a case of monkeypatching -- though I can't figure out how to do cleanly) to change the UserProvider implementation. (e.g. from default UserProviderWithCache to UserProvider ).

> What use case do you have in mind for changing the implementation?

Prefetching :) 
But also would be cool to be able to try it out with and without the user cache and see if it had any higher order difference.

But it's not really important. I think I was getting a bit theoretical with my comments in responses to this PR. Sorry if not helpful! :)

glebm:
> Ok -- I guess I was thinking that if the timing before fetching the users (in each thread) was relatively short (say 1 ms or 2), and the fetch was relatively long (say 5 ms), then the db accesses are likely to be in parallel. 

Compared to not caching at all, yeah, but as you mentioned elsewhere it's common for mentions to be repeated across posts. Rendering is also quite slow.

I think overall caching is a win also because it removes the non-obvious requirement of a larger database connection pool (something I've hit recently).

glebm:
> > What use case do you have in mind for changing the implementation?

> Prefetching :)

Prefetching would require some major changes, so might be tricky with monkey patching!

PR #793, opened 2019-02-22
Feature: Reporting posts to moderation for review

drusepth:
This PR adds the ability for users to report posts to moderation. It re-uses the existing moderation queue already in place.

Questions:

* Will sticking things in the existing moderation queue instantly hide those topics for forums that require moderation before post visibility? This seems undesirable.

* Is "logged-in users" the best policy requirement for who can or cannot report posts?

Other notes:

* I think, ideally, the best UX after reporting would be to replace the "Report Post" menu item with a greyed-out "Reported Post" to signify the user has already reported that post. I can't think of a way to track this without requiring a database migration, so I fell back on just having a notice display informing the user that they've successfully reported the post. (Another option is greying out the report option if `moderation_state` is `pending_moderation`, but I don't think it's a good idea to expose to _other_ users which posts have/haven't been reported).

This was a quick implementation because some forums I run on thredded very frequently request this feature, and I'm definitely open to revisions and feedback. 

Thanks for everything you do, guys. :)

Screenshots:

2019-02-22-151203_609x390_scrot

### DIFF
--- a/app/controllers/thredded/posts_controller.rb
+++ b/app/controllers/thredded/posts_controller.rb
@@ -81,10 +81,10 @@ module Thredded
       authorize_reading post
       post.update(moderation_state: 'pending_moderation')
       respond_to do |format|
-        format.html {
+        format.html do
           redirect_back fallback_location: post_path(post, user: thredded_current_user),
                         notice: I18n.t('thredded.posts.reported_post')
-        }
+        end
         format.json { render(json: { reported: true }) }
       end
     end
--- a/app/controllers/thredded/posts_controller.rb
+++ b/app/controllers/thredded/posts_controller.rb
@@ -7,7 +7,7 @@ module Thredded
     include Thredded::NewPostParams
 
     helper_method :topic
-    before_action :assign_messageboard_for_actions, only: %i[mark_as_read mark_as_unread]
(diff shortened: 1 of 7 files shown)


glebm:
> Will sticking things in the existing moderation queue instantly hide those topics for forums that require moderation before post visibility? This seems undesirable.

Yeah, we wouldn't want that.

> Is "logged-in users" the best policy requirement for who can or cannot report posts?

Perhaps, but there should still be other controls in place (e.g. maximum number of reports per day per user) as with this design a single bad actor can fill up the moderation queue by reporting every post on the forum.

> I can't think of a way to track this without requiring a database migration

Given the other consideration, a fully implemented reporting system will likely require a database migration anyway.

---

There are many ways to design / implement reporting.

For example, one may want to have a form that allows the user to specify a reason why they're reporting a post and a custom message.

It may also be desirable to have a separate moderation queue for reports.

I suggest implementing this feature as a plugin.

The core is meant to be lean and only have essential / unambiguous features, and provide extension points to plugins for other functionality.

Any missing extension points can be added to the core, whatever the plugin requires.

drusepth:
Thanks for the feedback. I agree a plugin seems to be the way to go here, so I'll go ahead and close this PR. Thanks!

Issue #795, opened 2019-03-02
Blocked topic


abarbuc:
Hi, 

This may be a basic question and from my search in the previous topics I could not find an answer. Apologies if it was covered and I am missing it. 

When a moderator blocks a topic how can this be displayed to the author? I see other authors do not see it and the moderator sees it as blocked but for the original author displays as normal.  Also is there a way when blocked to provide feedback why it was blocked?

Thank you for the great functionality as well, great work.

glebm:
> I see other authors do not see it and the moderator sees it as blocked but for the original author displays as normal.

This is intentional as blocking is mainly intended for spam.

If you want to add a setting for this, feel free to send a PR.

Note that if a user is "pending moderation" and you block their first topic, all their subsequent topics are blocked automatically (this is not the case if the user is "approved").

> Also is there a way when blocked to provide feedback why it was blocked?

Nothing specific to blocking. You can simply reply to the topic though.

abarbuc:
Thank you for the prompt reply and the clarifications. 

Is there to reply to a single blocked post? I see there is a quote option but that does not maintain the blocked properties of the initial message. 

I have a second issue with creating topics. Locally everything seems to work ok in terms of creating Topics. But once deployed I can create one topic (if the Browser History is cleared) and I can not create a second topic. In the developer tools console I see these errors:

> Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' https: 'nonce-pDnQi0nWruwDOwTeEY31EQ=='". Either the 'unsafe-inline' keyword, a hash ('sha256-7cGCBh58ly6eVRh4wmX6H0WwhYbFAKpZaPiOoKa+nEE='), or a nonce ('nonce-...') is required to enable inline execution.
> 
> An invalid form control with name='topic[content]' is not focusable.
> 

The Chrome version I am testing with is Version 72.0.3626.119. I was also testing in Firefox 65.0.1  and same thing. 

Thank you for taking the time to answer.

glebm:
> Refused to execute inline script because it violates the following Content Security Policy directive

It's what it says, you're setting a CSP header that doesn't allow Thredded to execute its inline script. Add the hash as above to your CSP header.

> I can not create a second topic

Likely caused by the above.

> An invalid form control with name='topic[content]' is not focusable.

I'm guessing this will be fixed once you fix the CSP header.

glebm:
For the CSP issue, I've added nonce to the inline scripts on master. Try it out.

[...]

PR #797, opened 2019-03-03
Add CSP nonce to inline script tags if CSP is enabled on Rails v5.2+

glebm:
Refs https://github.com/thredded/thredded/issues/795#issuecomment-468979746


jayroh:
This is awesome @glebm! Learned something new today

glebm:
So did I!

PR #800, opened 2019-03-10
Improve Rails 6 support

sudara:
This PR removes 2 instances of the invalid (and unused) option `:on` for `before_save` callbacks in models.

Without this change, the following error is raised in Rails 6: 
`Unknown key: :on. Valid keys are: :if, :unless, :prepend`

Please note that the `:on` option was a no-op before Rails 6, so this PR does not change any behavior. 

However, Rails 6 introduced stricter option checking for model callbacks:  https://github.com/rails/rails/pull/30919

These were the only changes required to get Thredded happy on Rails 6 for http://github.com/sudara/alonetone — Impressive!🥇

### DIFF
--- a/app/models/thredded/messageboard.rb
+++ b/app/models/thredded/messageboard.rb
@@ -24,7 +24,7 @@ module Thredded
     validates :name, uniqueness: true, length: { within: Thredded.messageboard_name_length_range }, presence: true
     validates :topics_count, numericality: true
     validates :position, presence: true, on: :update
-    before_save :ensure_position, on: :create
+    before_save :ensure_position
 
     def ensure_position
       self.position ||= (created_at || Time.zone.now).to_i
--- a/app/models/thredded/messageboard_group.rb
+++ b/app/models/thredded/messageboard_group.rb
@@ -10,7 +10,7 @@ module Thredded
     scope :ordered, -> { order(position: :asc, id: :asc) }
     validates :name, presence: true, uniqueness: true
     validates :position, presence: true, on: :update
-    before_save :ensure_position, on: :create
+    before_save :ensure_position
 
     def ensure_position
       self.position ||= Time.zone.now.to_i


glebm:
These callbacks are only supposed to be invoked once, when creating the record. Can you please change it to do that? Thanks!

glebm:
Nevermind, it's fine for these to run on update as well as they're idempotent.

sudara:
Just a heads up that I'm also currently looking into `CollectionToStringsWithCacheRenderer` which looks like it needs some love for Rails 6:

alonetone | Don't Panic! The White Theme FAQ  2019-03-15 20-57-30

Not sure yet if I will succeed, it gets complex!

sudara:
`PartialRenderer#render` now returns an object (instead of a string). I've temporarily got our app happy on Rails 6 with this, but obviously it's not sustainable or backwards compatible, so we'll have to revisit: https://github.com/thredded/thredded/commit/f6aa8cefbe4ad7d31cb99c6955691d85191495e8

glebm:
Thank you for your investigation. Can we just do this conditionally based on `Rails::VERSION::MAJOR >= 6`?

sudara:
Yeah, totally possible to switch on version, but I felt reluctant to implement without better understanding the class. Seems like it would need a tad bit of refactoring to support a switch like that. I'm also curious if this class could be the cause for some memory bloat we've been seeing...

glebm:
@sudara This class uses a thread pool to render posts in a topic.
Multi-threading here improves performance despite the GIL because rendering sometimes involves a network call (e.g. for the oneboxes).

> I'm also curious if this class could be the cause for some memory bloat we've been seeing...

It does trade off a bit of memory for speed. You can find out how much by disabling multi-threading like this in the initializer:

```ruby
Thredded::CollectionToStringsWithCacheRenderer.render_threads = 1
```

PR #811, opened 2019-05-10
Reporting posts

drusepth:
This feature would be better built as a plugin as per suggested but I figured I'd open this PR to track changes until a plugin is built.


drusepth:
Oops, I didn't mean to make this PR in the main repo!

PR #813, opened 2019-05-25
Fixed strange crash with route not found when using url_for

igorkasyanchuk:


glebm:
Can you reproduce this with a test? I couldn't reproduce it here nor on any of my websites that use Thredded.

igorkasyanchuk:
@glebm  I was not able to reproduce in dummy app.
I'm using threadded as engine in my multi-lingual app. It's mounted on "/en/forum" path. 
It's different from how your dummy app works.

app/views/thredded/posts_common/actions/_quote.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= link_to t('thredded.posts.quote_btn'), url_for(post.quote_url_params),
+<%= link_to t('thredded.posts.quote_btn'), post.quote_url,
```
timdiggins, line 1: @glebm I'm just re-reviewing this one (after a looong while! It would be good to just close it or) and I realise that this is in fact the only case where we use a bare `url_for` (rather than Thredded::UrlsHelper.url_for) and it might be really hard work to find the case where this is a problem -- might be better to avoid using url_for just defensively?

but if you disagree, we should close this anyway.

glebm: Sure, though if we extract it into a method let's name it something other than `quote_url` because there is already `quote_path` which is something else.

timdiggins:
url_for removed in #913

PR #819, opened 2019-06-20
Fix post_moderation_records user reference type

mauri:
The references to post_user and moderator, both Thredded.user_class, are not using the type of the user class id. This prevents saving the actual relation if a column type like uuid is used as primary key.


glebm:
Thanks!

PR #823, opened 2019-07-25
Fix #822 Join thredded_topics before referencing it.

### LINKED ISSUE #822: ActiveRecord::StatementInvalid in Thredded::MessageboardsController#index
Hi,

On a Rails 6 rc2 project, when attempting to access forum index, I'm encountering the following error :

`PG::UndefinedTable: ERROR: missing FROM-clause entry for table "thredded_topics" `

The generated SQL is:

> "SELECT \"thredded_messageboards\".* FROM \"thredded_messageboards\" LEFT OUTER JOIN \"thredded_user_topic_read_states\" ON \"thredded_messageboards\".\"id\" = \"thredded_user_topic_read_states\".\"messageboard_id\" AND \"thredded_user_topic_read_states\".\"postable_id\" = \"thredded_topics\".\"id\" AND \"thredded_user_topic_read_states\".\"user_id\" IS NULL AND \"thredded_user_topic_read_states\".\"unread_posts_count\" = 0 INNER JOIN \"thredded_topics\" ON \"thredded_topics\".\"messageboard_id\" = \"thredded_messageboards\".\"id\" WHERE (\"thredded_topics\".\"moderation_state\" != 2 OR \"thredded_topics\".\"user_id\" IS NULL) GROUP BY \"thredded_messageboards\".\"id\""

I tried to find a solution but quickly got lost with all those arel tables in [this class method](https://github.com/thredded/thredded/blob/master/app/models/thredded/messageboard.rb#L127)

toommz:


app/models/thredded/messageboard.rb
```diff
@@ -128,8 +128,13 @@ def unread_topics_counts(user:, topics_scope: Thredded::Topic.all)
         messageboards = arel_table
         read_states = Thredded::UserTopicReadState.arel_table
         topics = topics_scope.arel_table
-        joins(:topics).merge(topics_scope).joins(
-          messageboards.outer_join(read_states).on(
+        topics_ids = topics_scope.pluck(:id)
```
glebm: This is now 2 queries which is not ideal.

I'd prefer that we figure out why it stopped working in Rails 6.2.rc2 and either file a bug in Rails or fix it on our side properly.

glebm: Filed a bug in Rails

Issue #810, opened 2019-05-07
Feature request: view messageboards by messageboard_group


shelmire:
Basically have a show view for messageboard_groups and links to them instead of the plain titles.

This is a request for an app I've been working on. I've got a rough draft pr that you can see here (I will just go ahead and use this personally if there's no interest in this feature here): https://github.com/shelmire/thredded/pull/1/files

(made a slight error here, the where clause in the show action should be where(group: params[:id]) not id: params[:id])

I've been extremely lazy about this because I didn't want to get into messing with the views, but perhaps a maintainer would be interested in finishing this up.

Content of that pr is useable under MIT license as the main project is.

jayroh:
Could you post a screenshot of what that would look like? I'm having a tough time wrapping my head around what this is for.

shelmire:
Sure, I've got this working in a staging environment. Main forum index looks the same, except that the group names are now links. Click on a link and you'll just see the messageboards associated with that group. This lets us link directly to groups of forums, rather than being stuck with either all or one. Functionally, my application has many Projects, and I want to provide a link to the several forums associated with a given project, rather than to the forum index or to individual forums.

<img width="984" alt="Screen Shot 2019-05-09 at 2 12 11 PM" src="https://user-images.githubusercontent.com/5061248/57476479-ab6f3980-7264-11e9-804c-05a88ac63077.png">
<img width="554" alt="Screen Shot 2019-05-09 at 2 12 24 PM" src="https://user-images.githubusercontent.com/5061248/57476481-ad38fd00-7264-11e9-8432-2fd22ba2ecf8.png">
<img width="1015" alt="Screen Shot 2019-05-09 at 2 12 32 PM" src="https://user-images.githubusercontent.com/5061248/57476489-af9b5700-7264-11e9-8002-8b5924c6b9da.png">

glebm:
This does not seem like a common use case, so maybe as setting (off by default)?

Something like `Thredded.show_messageboard_group_page`?

shelmire:
I'd actually argue that this is closer to the default forum use case. See, for example, https://www.mtgsalvation.com/forums ; each of the large forum groups has a number of subforums under it. Other forums such as https://www.carforums.com/forums/ or https://us.battle.net/forums/en/ similarly have separate pages for groups of forums (these were just random picks of forums I'm aware of). It's also a fairly innocuous change (I can't imagine this breaking anything).

shelmire:
[...]

PR #829, opened 2019-08-21
Show mbs by group

shelmire:
This turns Messageboard Group names into links on the index page, and re-uses that view to display the MessageBoards that belong to that group only when it's clicked. This allows us to do things like link to that group directly, which is a common (dare I say, normal and default) functionality for most forum software. In my use case there's a Project model; each project has a messageboard group associated with it. Messageboards belonging to other projects are irrelevant within this scope, so I need a way to show just the messageboards belonging to that group.

We previously discussed this on https://github.com/thredded/thredded/issues/810

I made an argument, which I hope was successful, that this should be default behavior, and not toggled by a setting. I have been unable to think of any cases where someone would not desire this behavior, or where it could be dangerous. If you still disagree, I can make it a setting, but I'd encourage you to make it default to true.


shelmire:
@glebm Let me know if changes are desired.

app/controllers/thredded/messageboard_groups_controller.rb
```diff
@@ -20,6 +20,14 @@ def create
       end
     end
 
+    def show
+      @groups = Thredded::MessageboardGroupView.grouped(
+        policy_scope(Thredded::Messageboard.where(group: params[:id])),
+        user: thredded_current_user
+      )
+      render template: 'thredded/messageboards/index'
```
glebm: Let's use a separate template. The page should have a different title and view hooks.

app/views/thredded/messageboards/index.html.erb
```diff
@@ -8,7 +8,10 @@
       <%= view_hooks.messageboards_index.list.render self, groups: @groups do %>
         <% @groups.each do |group| %>
           <% if group.name.present? %>
-            <h3 class="thredded--messageboards-group--title"><%= group.name %></h3>
+            <h3 class="thredded--messageboards-group--title">
+              <%= link_to group.name,
+                          show_messageboard_group_path(group.id) %>
```
glebm: We use "slugs" for human-readable URLs throughout.
Let's think what URL format would work best here. Perhaps "/g/<slug>"?

app/views/thredded/messageboards/index.html.erb
```diff
@@ -8,7 +8,10 @@
       <%= view_hooks.messageboards_index.list.render self, groups: @groups do %>
         <% @groups.each do |group| %>
           <% if group.name.present? %>
-            <h3 class="thredded--messageboards-group--title"><%= group.name %></h3>
+            <h3 class="thredded--messageboards-group--title">
+              <%= link_to group.name,
```
glebm: Please add a setting to control this and use `link_to_if Thredded.show_messageboard_group_page`.

I keep track of a dozen forums using Thredded, and very few of them have more than 6 boards, so this link would be confusing.

I'm OK with the setting being `true` by default.

config/routes.rb
```diff
@@ -59,6 +59,7 @@
 
   resource :preferences, only: %i[edit update], as: :global_preferences
   resource :messageboard, path: 'messageboards', only: [:new]
+  get '/messageboard_groups/:id', action: :show, controller: 'messageboard_groups', as: :show_messageboard_group
```
glebm: 1. Change to "messageboard-groups" (Thredded uses hyphens throughout).
2. "messageboard-groups" needs to be added to the list at https://github.com/thredded/thredded/blob/13f4861b978a1450a0852a11f98582b4fb7ebac0/app/models/thredded/messageboard.rb#L9

spec/features/thredded/user_visits_messageboard_group_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.feature 'Creating a messageboard group' do
+  it 'admin can visit particular groups' do
```
glebm: Let's test for a regular user instead? The page is not limited to admins.

app/views/thredded/messageboards/index.html.erb
shelmire: Sounds good.

shelmire:
@glebm It's been a bit (had other work priorities lately), but I believe I've resolved the issues mentioned. I thought there was another request, but I appear to have imagined it. Thanks again for the thorough review and expertise. This gem is doing a lot of things I've never done before, so it's an interesting learning experience.

app/views/thredded/messageboards/index.html.erb
```diff
@@ -9,8 +9,11 @@
         <% @groups.each do |group| %>
           <% if group.name.present? %>
             <h3 class="thredded--messageboards-group--title">
-              <%= link_to group.name,
-                          show_messageboard_group_path(group.id) %>
+              <%= link_to_if Thredded.show_messageboard_group_page,
+                            group.name,
+                            show_messageboard_group_path(group.id) do 
+                              group.name
+                            end%>
```
glebm: You can remove the `do...end` part, since it's the same as the second argument (`group.name`)

app/controllers/thredded/messageboard_groups_controller.rb
```diff
@@ -20,6 +20,15 @@ def create
       end
     end
 
+    def show
+      @group = Thredded::MessageboardGroup.where(id: params[:id])
+      @groups = Thredded::MessageboardGroupView.grouped(
+        policy_scope(Thredded::Messageboard.where(group: params[:id])),
+        user: thredded_current_user
+      )
+      render :show
```
glebm: Remove this line, Rails does `render action_name` by default

shelmire: Corrected.

app/views/thredded/messageboards/index.html.erb
shelmire: Corrected.

PR #837, opened 2019-10-19
sprockets 4 support

ahorek:


spec/dummy/app/assets/config/manifest.js
```diff
@@ -0,0 +1,6 @@
+//= link application.js
+//= link day.css
+//= link night.css
+//= link email.css
+//= link favicon.ico
```
glebm: Let's just link all the images:

```
//= link_tree ../images/
```

app/assets/config/thredded_manifest.js
```diff
@@ -0,0 +1,4 @@
+//= link thredded.js
+//= link thredded.css
+//= link_directory ../images/thredded .svg
```
glebm: There are some png images as well for the onebox:
https://github.com/thredded/thredded/tree/master/app/assets/images/favicons

spec/dummy/app/assets/config/manifest.js
ahorek: done

app/assets/config/thredded_manifest.js
ahorek: thanks, done

Issue #831, opened 2019-10-10
Running Thredded with Rails 6 and webpack


dazzaroonie:
I can't seem to find a way to run Thredded using Rails 6 and webpack.

There is no way to require the javascript.

Any plans to have it on npm? Or is there a way I can have this working with Webpack?

glebm:
Thredded still uses Sprockets, which Rails 6 still supports.

glebm:
> Any plans to have it on npm?

Not currently.

> Or is there a way I can have this working with Webpack?

I found this on including JavaScript from gems directly into the pack:

https://github.com/rails/webpacker/issues/57#issuecomment-491317195

glebm:
Perhaps Thredded can provide a helper for this

glebm:
Released v0.16.4 with Webpack support!
https://github.com/thredded/thredded/blob/v0.16.14/README.md#add-thredded-javascripts-webpack

PR #838, opened 2019-10-20
Experimental Webpack support

glebm:
Refs #831

@ahorek @dazzaroonie Please take a look

thredded_create_app PR: https://github.com/thredded/thredded_create_app/pull/17


Gemfile
```diff
@@ -4,6 +4,7 @@ source 'https://rubygems.org'
 
 gem 'rails', '~> 6.0.0'
 gem 'rails-i18n', '~> 6.0.0'
+gem 'webpacker', '~> 4.0'
```
ahorek: the gem should be also in https://github.com/thredded/thredded/blob/master/spec/gemfiles/rails_6_0.gemfile ?

ideally you should test both variants
1/ rails 6 + webpack
2/ rails 6 + sprockets 4

spec/dummy/config/initializers/assets.rb
```diff
@@ -1,5 +1,13 @@
 # frozen_string_literal: true
 
-if Gem::Version.new(Sprockets::VERSION) < Gem::Version.new('4')
-  Rails.application.config.assets.precompile += %w[manifest.js]
+Rails.application.config.assets.precompile +=
+  if Gem::Version.new(Sprockets::VERSION) < Gem::Version.new('4')
+    %w[manifest.js]
+  else
+    %w[day.css night.css email.css]
```
ahorek: this shouln't be necessary, the manifest file is already included https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/railtie.rb#L107

spec/dummy/config/application.rb
```diff
@@ -29,6 +29,11 @@
 
 require 'web-console' if Rails.env.development?
 
+# If true, use Webpack. Otherwise, use Sprockets.
+THREDDED_TESTAPP_WEBPACK = Rails::VERSION::MAJOR >= 6 && ENV['THREDDED_TESTAPP_SPROCKETS'] != '1'
```
ahorek: you should avoid using constants in a global scope, I would prefer a helper method

glebm: Done. Some duplication but whatever.

spec/dummy/config/initializers/assets.rb
glebm: Done

spec/dummy/config/initializers/assets.rb
```diff
@@ -1,5 +1,10 @@
 # frozen_string_literal: true
 
 if Gem::Version.new(Sprockets::VERSION) < Gem::Version.new('4')
-  Rails.application.config.assets.precompile += %w[manifest.js]
+  Rails.application.config.assets.precompile += %w[day.css night.css email.css]
```
ahorek: just precompile manifest.js for sprockets < 4 (sprockets 4 includes it by default), this way you don't have to define top level assets at two places

Gemfile
glebm: Done

PR #852, opened 2020-01-04
Post caching fixes for Rails 6

### LINKED ISSUE #839: Caching bug: Post content replaced by another post
Really strange (and bad!) bug here on Rails 6 with Thredded 0.16.13

I figured I would first open this issue in case others have ran into it, then investigate cause. 

In some cases, it seems that every single post in a topic has its content replaced by a the last post. 

Perhaps it’s caching related?

65352820-dcd11c00-dbeb-11e9-964c-bc02fcd324bc

See: https://github.com/sudara/alonetone/issues/780

glebm:
This does not seem to fix issue #839

### DIFF
--- a/app/view_models/thredded/post_view.rb
+++ b/app/view_models/thredded/post_view.rb
@@ -13,6 +13,7 @@ module Thredded
              :blocked?,
              :last_moderation_record,
              :cache_key,
+             :cache_version,
              :cache_key_with_version,
              to: :@post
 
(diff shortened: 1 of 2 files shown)


jayroh:
p.s. _"This may or may not fix"_ ... I love the confidence :laughing:

glebm:
Doesn't seem to fix the issue but is an improvement anyway, merging.

PR #859, opened 2020-03-24
Fix label reference for localized values

denzelem:
In case the i18n translation is not the same value as the attribute
itself, the referenced ID in the label will not match the input
field.

Here is an example with the locale set to 'de':

Before this fix:

```
<li>
  <label for="messageboard_Name">Name</label>
  <input required="required" type="text" name="messageboard[name]" id="messageboard_name">
</li>
```

After this fix:

```
<li>
  <label for="messageboard_name">Name</label>
  <input required="required" type="text" name="messageboard[name]" id="messageboard_name">
</li>
```

### DIFF
--- a/app/views/thredded/messageboards/_form.html.erb
+++ b/app/views/thredded/messageboards/_form.html.erb
@@ -1,16 +1,16 @@
 <%= form_for messageboard, html: { class: 'thredded--form' } do |f| %>
     <ul class="thredded--form-list">
       <li>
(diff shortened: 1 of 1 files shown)


denzelem:
@glebm @andreibondarev I wrote this small fix to have a better support for my integration tests, that rely on correct linked label fields. Imho other users will profit from this change, too.

Can you please review the changes? Thanks!

glebm:
Thanks!

PR #861, opened 2020-05-09
Add safe navigation for postable

### LINKED ISSUE #860: Destruction of messsageboard fails
It appears that the destruction of messageboards (either directly via messageboard.destroy or via the gui as introduced in #825) does not work:

```
NoMethodError (undefined method `update_last_user_and_time_from_last_post!' for nil:NilClass):

thredded (0.16.16) app/models/thredded/post.rb:81:in `update_parent_last_user_and_time_from_last_post'

```
The reason seems to be that the callbacks are triggered in an order that destroys topics before posts, resulting in postables that are `nil`. Adding safe navigation to postable seems to fix the problem.

fosterfarrell9:
This fixes #860 .

### DIFF
--- a/app/models/concerns/thredded/post_common.rb
+++ b/app/models/concerns/thredded/post_common.rb
@@ -40,7 +40,8 @@ module Thredded
 
       before_validation :ensure_user_detail, on: :create
 
-      after_commit :update_unread_posts_count, on: %i[create destroy]
+      # postable won't exist when destroying a messageboard because topics are destroyed before the posts
+      after_commit :update_unread_posts_count, on: %i[create destroy], if: :postable
     end
 
     def avatar_url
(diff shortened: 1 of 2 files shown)


app/models/concerns/thredded/post_common.rb
```diff
@@ -81,7 +81,7 @@ def previous_post
     protected
 
     def update_unread_posts_count
-      postable.user_read_states.update_post_counts!
+      postable&.user_read_states&.update_post_counts!
```
glebm, line 84: if the postable exists then so will `user_read_states`
Instead, let's add `, if: :postable` on line 43

glebm:
Perhaps a better way to solve this is to avoid calling this method and a similar one on `Topic` when we're destroying a messageboard.

timdiggins:
Fixed in #914

PR #867, opened 2020-05-28
Remove warning from spec runs

timdiggins:
Before this fix, rspec runs would include the following warning:

> WARNING: `around(:context)` hooks are not supported and behave like `around(:example). Called from /Users/tim/workspaces/thredded/thredded/spec/commands/thredded/autofollow_users_spec.rb:74:in `block (2 levels) in <module:Thredded>'.

### DIFF
--- a/spec/commands/thredded/autofollow_users_spec.rb
+++ b/spec/commands/thredded/autofollow_users_spec.rb
@@ -71,7 +71,7 @@ module Thredded
     end
 
     context 'with thredded_user_preferences.auto_follow_topics default true' do
-      around(:all) do |ex|
+      around do |ex|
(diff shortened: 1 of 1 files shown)


spec/commands/thredded/autofollow_users_spec.rb
```diff
@@ -71,7 +71,7 @@ module Thredded
     end
 
     context 'with thredded_user_preferences.auto_follow_topics default true' do
-      around(:all) do |ex|
+      around do |ex|
```
glebm, line 74: If we add more tests, we should use a `before(:all)` and `after(:all)` hook to minimize the number of schema changes. Since it's just the one test, it doesn't matter here.

Issue #880, opened 2021-04-03
ruby-gravatar and GPL licensing


rickychilcott:
In https://github.com/sn/ruby-gravatar/commit/5a1225c58f9363e66daae01698fcefc7d4f697b3, the author changed the license to GPL (from MIT). If you accidentally update all of your gems, you're likely to get the latest version which could jeopardize your business if Thredded is not used standalone and is used inside of another app.

I made an API-compatible gem (similar but different code) at https://github.com/mission-met/rails_gravatar and should more or less be a drop in. Bonus: I actually wrote tests for my gem and DON'T require using the `raw` tag helper around everything.

I can try to make a PR tomorrow or Sunday, if it will be accepted to this project but it might be faster for @glebm to just try out my gem, make the change, cut a new gem, and call it a day.

rickychilcott:
Actually, it's not 100% API compatabile, but basically is. Steps to update:

1. Drop the `raw` tags
2. Change all `Gravatar.` to `RailsGravatar.`
3. Change `Gravatar.prefetch_dns` to `RailsGravatar.prefetch_dns_tag`

timdiggins:
@rickychilcott the gem sounds cool, but I'm not sure about the risk statement ("GPL ... could jeopardize your business if Thredded is not used standalone and is used inside of another app"). Can you provide some reference that this is the case?

glebm:
Your gem sounds like an improvement, please feel free to send a PR

rickychilcott:
If you haven't recently followed the Rails/Mimemagic issue, https://emmanuelhayford.com/dislodging-mimemagic-and-understanding-mit-gnu-gpl/ appears to be a decent write up, Suez canal jokes aside.

Also, if you're now panicking that you too might have GPL dependencies in your codebase, check out https://github.com/github/licensed It might help you figure out semi-quickly.

I'll send a PR.

rickychilcott:
Thanks @glebm!

timdiggins:
PS @rickychilcott Thanks for the link  -- totally irrelevant of the PR and your great work on the railsgravatar gem and this gem, I just think we need to be really careful of the claims we're making around GPL  -- and avoid spreading panic.

[...]

PR #881, opened 2021-04-05
Switch to RailsGravatar

rickychilcott:
This addresses #880 which is a concern about relying on the rb-gravatar gem, which has been GPLed in https://github.com/sn/ruby-gravatar/commit/5a1225c58f9363e66daae01698fcefc7d4f697b3 We _could_ peg our dependency on an older version of the gem, but it might be better to rely on a gem that is guaranteed to not be GPLed.

As such, I created a gem called [RailsGravatar](https://github.com/mission-met/rails_gravatar) which has a very similar API but changes some of the internals to better output rails tags (using `ActionView` helpers) and wrote a comprehensive test suite. The most important part is that it's MIT licensed and has no concerns about it being GPLed.


glebm:
Thanks!

glebm:
Looks like this isn't working correctly (CI failing)

The 3rd argument to `rb-gravatar` was `'mm'` but RailsGravatar expects a URL?

Can you please send a PR fixing this?

glebm:
Looking at the source, it's not the `fallback_image_url`; not sure what the issue is

glebm:
Example failing job https://travis-ci.org/github/thredded/thredded/jobs/766574511

Error:

```
   Failure/Error: <%= image_tag post.avatar_url, class: 'thredded--post--avatar' if post.user %>

      ActionView::Template::Error:

        nil is not a valid asset source
```

timdiggins:
@rickychilcott we can't reopen this PR because it was merged once, but can you open another PR for this when you're ready?

PR #886, opened 2021-04-10
Revert "Switch to RailsGravatar" to get CI green

timdiggins:
I feel like we want to get CI green again. We can re-merge a new version of #881 when it's green.

Reverts thredded/thredded#881


timdiggins:
@glebm Other failures are due to nodejs requirements of latest autoprefixer-rails. Working on that in #887 ...

But shall we revert this and reopen #881?

timdiggins:
rebased on master which now includes #887 -- so this should now go green 🤞

glebm:
Yeah let's revert and reopen :+1:

PR #887, opened 2021-04-10
use node 12 for autoprefixer-rails compatibility

timdiggins:


timdiggins:
@glebm when combined with #886 this makes travis green: https://travis-ci.org/github/red56/thredded/builds/766586525 🎉 

Once this goes through travis (won't be green yet), I will merge this then rebase #886 on it. I'll then merge #886 (if it's then green) unless you shout.

PR #889, opened 2021-04-11
Switch to RailsGravatar

rickychilcott:


rickychilcott:
Well... hmm. I just cherry-picked the main commit and pulled against the latest master and everything seems OK.

rickychilcott:
Ok. I didn't wait long enough. I'll investigate further.

rickychilcott:
OK. I figured it out. It was fixed in RailsGravatar 1.0.4. I was using the `Comparable#clamp` method to conform the size param, but that wasn't released until Ruby 2.5. Also, I was rescuing all exceptions and just outputting `nil` if there was an error, which didn't surface the underlying issue.

That's all been fixed right now. CI is rerunning after a rebase and force push, but it will pass.

Ricky

PR #893, opened 2021-07-13
Update Onebox

rickychilcott:
To allow me to update the [addressable](https://github.com/sporkmonger/addressable) gem (which is desired because of CVE-2021-32740), I need to update the [onebox](https://github.com/discourse/onebox) gem.

This PR is testing whether a onebox upgrade to 2.x will cause problems

### DIFF
--- a/thredded.gemspec
+++ b/thredded.gemspec
@@ -36,7 +36,7 @@ Thredded works with SQLite, MySQL (v5.6.4+), and PostgreSQL. See the demo at htt
   s.add_dependency 'html-pipeline'
   s.add_dependency 'kramdown', '>= 2.0.0'
   s.add_dependency 'kramdown-parser-gfm'
-  s.add_dependency 'onebox', '~> 1.8', '>= 1.8.99'
+  s.add_dependency 'onebox', '>= 1.8.99'
   # html-pipeline dependencies, see https://github.com/jch/html-pipeline#dependencies
   # for the AutolinkFilter
   s.add_dependency 'rinku'


rickychilcott:
@glebm @timdiggins - CI used to run on PRs, but it appears that Travis no longer has their `.org` builds any longer. See https://travis-ci.org/github/thredded/thredded/branches

Can one of you sign up for a free travis-ci account on travis-ci.com and see if we can get builds going again?

thredded.gemspec
```diff
@@ -36,7 +36,7 @@ Thredded works with SQLite, MySQL (v5.6.4+), and PostgreSQL. See the demo at htt
   s.add_dependency 'html-pipeline'
   s.add_dependency 'kramdown', '>= 2.0.0'
   s.add_dependency 'kramdown-parser-gfm'
-  s.add_dependency 'onebox', '~> 1.8', '>= 1.8.99'
+  s.add_dependency 'onebox', '~> 2.2'
```
timdiggins, line 39: my feeling is, as a library, we should prefer looser dependencies (ie. no ~> or <) unless we know there's a specific issue (e.g. it won't work with that version). Thus we could simplify this to '>= 1.8.99'. This allows users of the library to be specific about which version they want to use.

timdiggins: That said I just read your comment "This PR is testing whether..." -- but you could test that by loosening the version and running it locally I think. (Sadly there doesn't seem to be any indication of what breaking changes  Onebox 2.x brings in, if any: https://github.com/discourse/onebox/blob/master/CHANGELOG.md)

timdiggins: PS I've run the specs locally against rails 6.0 x sqllite x onebox 2.2  and passes locally. Am in communication with travis support.

rickychilcott: Thank you, Tim. I haven't been able to get a local testing environment running locally, unfortunately. I'll try again next time I run into something.

I'm happy to make it `'>= 1.8.99'`, which should get me over the CVE-2021-32740 issue.

I was also bummed to not see any details about what 2.x brings. I'm going to try a version of the app locally with this and deploy to a staging environment before I request a review.

Thank you for your work on this @timdiggins!

timdiggins: If you want to make it >= 1.8.99, it should still run on travis with the latest version (because bundle install without a lockfile should I think install the latest version, even if travis is caching gems). Travis-ci.com is now building I think.

PS running specs locally is pretty straightforward, just a few  steps -- https://github.com/thredded/thredded/#testing gives all the details (though in slightly the backwards order -- how to run first, then how to setup )

timdiggins: FYI @rickychilcott #895

rickychilcott:
I was able to merge in master and change it to `s.add_dependency 'onebox', '>= 1.8.99'`

It successfully passes all test. I also tried a few URLs in a deployment to my staging environment and everything seems fine.

Note that this is still not allowing us to go to addressable 2.8.0 (to resolve CVE-2021-32740), but I will work with the onebox team to cut a release of onebox that will allow that to happen.

@timdiggins  can you merge?

rickychilcott:
Thank you @timdiggins. Onebox has cut a new gem now as well. All is right with the world!

PR #894, opened 2021-07-13
drop rails 4.2 - 5.1 support to simplify ongoing development

### LINKED ISSUE #885: Drop rails 4.2 - 5.1 support
As discussed in #883, supporting old versions of rails is making work on this gem harder, and makes it hard to modernise aspects of it. 

So we're planning to remove rails 4.2 - 5.1 support.

timdiggins:
Outstanding tasks

* [x] remove rails 4.2 - 5.1 runs from travis 
* [x] use relation.pluck directly
* [x] inline true_value (true) and drop method
*  ~~drop database cleaner and use rails system specs~~ see #898

closes #885

### DIFF
--- a/app/controllers/thredded/application_controller.rb
+++ b/app/controllers/thredded/application_controller.rb
@@ -63,15 +63,6 @@ module Thredded
       @is_thredded_moderator = !thredded_current_user.thredded_can_moderate_messageboards.empty?
     end
 
-    if Rails::VERSION::MAJOR < 5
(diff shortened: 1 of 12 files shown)


timdiggins:
FYI @glebm we are waiting on travis OSS credits to start building again (i have just requested them), but the code is ready for review, just let's not merge till we get a green CI

rickychilcott:
Travis is green!

glebm (approved):

PR #909, opened 2022-02-13
fix an issue with the latest roadie

timdiggins:
a change to roadie **seems to** enforce an extra param to the
before_transformation callback. This was always documented
so not a breaking change. It is causing test failures, eg.
https://app.travis-ci.com/github/thredded/thredded/builds/246374669


glebm (approved):
Thanks!

Issue #862, opened 2020-05-15
Database Pool Exhaustion


rickychilcott:
I'm intermittently experiencing `ActionView::Template::Error: could not obtain a connection from the pool within 5.000 seconds (waited 5.000 seconds); all pooled connections were in use` errors.

I'm fairly certain I've tracked it down to `Thread` creation/execution in https://github.com/thredded/thredded/blob/bde4da03f498f8587f706acc90bef34ab644f0f1/lib/thredded/collection_to_strings_with_cache_renderer.rb#L82-L89

I think I understand why the partials are rendered this way -- an attempt at parallelizing view rendering for very expensive/large-collection partials? -- but it also isn't protecting against going over the pool limit and triggering issues like I'm experiencing.

For now, I've set `Thredded::CollectionToStringsWithCacheRenderer.render_threads = 1` which appears to have resolved the issue - albeit with slightly lower performance. It took some digging to figure that out, so I'm not sure if this is well documented yet or if other have experienced this before?

I also wonder if I can somehow increase the number of rendering threads beyond 1, but stop db exhaustion from happening. I spent a few minutes testing the concept of wrapping each serial partial render with `ActiveRecord::Base.connection_pool.with_connection` to grab a connection and return it after each render. Perhaps I'm just not fully testing the system, but it seemed to resolve my issues -- though I'd think that it could still cause pool exhaustion (if you have 10 connections in your pool and you try to render a large collection that generates 25 threads, it's possible it could take too long to get a connection)

```ruby
def render_partials_serial(view_context, collection, opts)
  ActiveRecord::Base.connection_pool.with_connection do
    partial_renderer = ActionView::PartialRenderer.new(@lookup_context)
    collection.map { |object| render_partial(partial_renderer, view_context, opts.merge(object: object)) }
  end
end
```

Any thoughts or ways I can assist? I do recognize that if I increased my database pool to allow more connections, I may never hit this issue, but resources are somewhat constrained and it was surprising to see a `Thread.start` in Thredded without some more documentation around it or issues being raised.

Thanks for everyone's work on this fine gem.

Velora:
Hey @rickychilcott I wanted to chime in to say that we are experiencing the same issue as you. 

I plan on looking into this more in the near future, but I wanted to share what I have found so far in case it is helpful to anyone. 

First, ```Thredded::CollectionToStringsWithCacheRenderer.render_threads = 1``` does resolve the issue for us as well. 

[...]

PR #923, opened 2022-02-22
prep for v1 release

timdiggins:


timdiggins:
@glebm could you check the changelog in particular. Not sure if there's anything else notable to extract from the detailed log.

I'd like to address #862 somehow (even if just something in the README?) before release 1.0 but not sure what's best.

timdiggins:
@glebm last quick look at the updated Changelog? (have added in some thank yous, not sure if that's normal, but seems polite)

glebm (approved):

glebm:
Looks good!

PR #926, opened 2022-02-23
ensure db pool isn't exhausted during parallel rendering.

timdiggins:
~~this is draft for comment. it includes all the commits 
which I used to generate the test data, but I will separate
these out into separate PRs where relevant.~~


rickychilcott:
I deployed this draft branch to production and set THREDDED_RENDER_THREADS to 10.

As suspected, this works like a charm 🥂

PR #938, opened 2022-02-26
update release commands in release checklist [CI SKIP]

timdiggins:


RELEASE-CHECKLIST.md
```diff
@@ -18,6 +18,8 @@ Release checklist:
   ```bash
   gem tag && gem release
   ```
+
+  (alternatively if gem-release isn't installed you can use `rake release`)
```
glebm, line 22: Or just always use `bundle exec rake release`?
`rake release` tags as well.

timdiggins: Ok agreed - in that case can drop reference to installing gem-release gem, which simplifies the readme (latest commit)

glebm (approved):

PR #942, opened 2022-03-09
prevent unsupported domain from generating empty link

### LINKED ISSUE #940: Bold Links are Completely Stripped from Email Notifications
**Expected behavior:** 

1. I write a post with a link that is made bold with markdown.
2. Followers receive a notification email containing all content from the post

**Actual behavior:** 

1. The notification email received completely removes links if they were bold in the original post. 

**Steps to Reproduce:**

1. Write a post containing the following: `**https://mysite.com**`
2. That text/html will be completely missing from the email received. 

Using v0.16.16

timdiggins:
fixes #940

### DIFF
--- a/spec/lib/thredded/html_pipeline/onebox_filter_spec.rb
+++ b/spec/lib/thredded/html_pipeline/onebox_filter_spec.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Thredded::HtmlPipeline::OneboxFilter do
+  subject(:onebox_filter) { described_class.new(document, context) }
+
+  let(:context) { nil }
+
+  def find_singular_a_node(transformed)
+    a_nodes = Nokogiri.HTML5(transformed).css('a[href]')
+    expect(a_nodes.length).to eq(1)
+    a_nodes.first
+  end
+
+  context 'with a link to an unsupported domain' do
+    let(:document) do
+      <<~HTML
+      <!DOCTYPE html>
+      <title>name</title>
+      <body>
+        <a href='#{href}'>#{href}</a>
+      </body>"
+      HTML
+    end
+    let(:href) { 'https://www.example.com' }
+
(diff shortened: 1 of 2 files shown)


timdiggins:
@glebm I can't work out why this is passing on my local machine and failing (on every matrix combination, just on the specific spec:

https://github.com/thredded/thredded/pull/942/files#diff-c8734eed0b22bc15c413b3a3584dfc4999461a5ba1fb86e49bc760134afe1293R73

spec/lib/thredded/html_pipeline/onebox_filter_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  context 'with a strong link to an unsupported domain' do
+    let(:document) do
+      <<~HTML
+      <!DOCTYPE html>
+      <title>name</title>
+      <body>
+        <strong><a href='#{href}'>#{href}</a></strong>
+      </body>"
```
glebm, line 56: This `"` probably shouldn't be here

timdiggins: Agree with you, removed it, but didn't make it pass. 😢

timdiggins:
So it's failing (for the last two runs - before and after the fix to remove the extraneous quote) on the travis "branch" run and succeeding on the travis "pull" run.

There are no extra commits on master, so the code from the repo will be the same. 

Maybe there's something weird in the cache - either the bundler cache or the onebox cache? 

I've tried turning off the onebox cache for this test. Waiting for more credits from travis so the test runs start again

glebm:
Odd, perhaps let's log the entire html to see what it produces

timdiggins:
Well either it fixed itself or forcing the onebox cache to be NullCache did the trick. Now green on both. I think I'll merge and worry about it if it surfaces again.

glebm (approved):

Issue #944, opened 2022-04-11
Incompatible with: html-pipeline version > 2


rakshassa:
HTML::Pipeline::SanitizationFilter::WHITELIST is not available.

NOTE: using gem 'html-pipeline', '~> 1.11' fixes this problem.

timdiggins:
@rakshassa thanks for reporting this. I think this changed was introduced in html-pipeline actually in v2.13.1 (but reverted accidentally in v2.14.0 and reintroduced in v2.14.1). PR #945 (merged) largely addresses this by limiting the version, but a new PR will allow for more recent (> 2.14.1) versions of html-pipeline.

timdiggins:
Again, thanks for the report! This should be fixed in master because of #951 -- please reopen this you have anything to add.

PR #945, opened 2022-05-04
[#944] limit version of html-pipeline for HTML::Pipeline::SanitizationFilter::WHITELIST

joeadcock:
HTML::Pipeline::SanitizationFilter::WHITELIST was renamed to HTML::Pipeline::SanitizationFilter::ALLOWLIST in html-pipeline v2.14.1

### DIFF
--- a/thredded.gemspec
+++ b/thredded.gemspec
@@ -33,7 +33,7 @@ Thredded works with SQLite, MySQL (v5.6.4+), and PostgreSQL. See the demo at htt
   s.add_dependency 'rails_gravatar'
 
   # post rendering
-  s.add_dependency 'html-pipeline'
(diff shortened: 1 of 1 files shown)


timdiggins:
@joeadcock Thanks! I'll get going with accommodating more recent HTML::Pipelines, but this gets master green again. 

FYI according to this https://github.com/gjtorikian/html-pipeline/commit/539a3c2ef5ecf29eb8e247869bef59f56c37bd8e and https://my.diffend.io/gems/html-pipeline/2.13.1/2.14.0 versions HTML-Pipeline has had a bumpy release history and this change (WHITELIST -> ALLOWLIST) occurs in v2.13.1, v2.13.2, and v2.14.1 (but not v2.14.0).

PR #947, opened 2022-05-28
Lock html-pipeline version to 2.14.0 to resolve uninitialized constant error

jasonxp:
Lock html-pipeline version to 2.14.0 due to the fact that ```HTML::Pipeline::SanitizationFilter::WHITELIST``` was renamed to ```HTML::Pipeline::SanitizationFilter::ALLOWLIST``` in ```html-pipeline``` v2.14.1. This caused the error ```thredded-1.0.0/lib/thredded/content_formatter.rb:31:in `<class:ContentFormatter>': uninitialized constant HTML::Pipeline::SanitizationFilter::WHITELIST (NameError)``` in apps using ```thredded``` upon a bundle update.

### DIFF
--- a/thredded.gemspec
+++ b/thredded.gemspec
@@ -33,7 +33,7 @@ Thredded works with SQLite, MySQL (v5.6.4+), and PostgreSQL. See the demo at htt
   s.add_dependency 'rails_gravatar'
 
   # post rendering
-  s.add_dependency 'html-pipeline'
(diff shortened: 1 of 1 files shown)


timdiggins:
@jasonxp thanks for your PR - it was an almost duplicate of #945 which I've merged (and have added in some edge cases in #948).

PR #948, opened 2022-06-03
tweak html-pipeline dependency

timdiggins:
excluding version with renamed ALLOWLIST (for now).


timdiggins:
Dropped in favour of #951

PR #949, opened 2022-06-03
fix error in standalone navigation for sign in

timdiggins:


timdiggins:
Only found this egregous error thanks to i18n-tasks health. Thanks @glebm! FYI will try to deal with the rails 5.2 error separately. It seems to be recurring fairly predictably but only on travis - maybe an issue with ruby 2.6.6 which travis is using.

PR #952, opened 2022-06-03
drop existing syntax for outdated unsupported rubies

timdiggins:


timdiggins:
FYI @glebm there's a follow-on PR to upgrade rubocop

glebm:
I guess we should also bump `required_ruby_version` in the gemspec once this is merged.

glebm (approved):

Issue #686, opened 2018-03-09
Route calls should be called with engine namespace


michelson:
Hello , I just installed this interesting engine into an existing rails app 4.2. 
I've got some deprecation warnings regarding the route generation. After some debugging I found out that in your views you are calling the routes directly. instead of doing `thredded.messageboards_path` you are calling `messageboards_path`. In terms of engine implementation that's a bad practice because the engine could collide with existing routes or just hit my issue: 
```
DEPRECATION WARNING: You are trying to generate the URL for a named route called "private_topics" but no such route was found. In the future, this will result
```
that ends rendering , in rails 4 , routes like `http://localhost:3000/forum/assets?action=new&controller=thredded%2Fmessageboards` but in rails 5 will raise error

glebm:
> In terms of engine implementation that's a bad practice because the engine could collide with existing routes

That's not how route lookup works in an engine. Routes are looked up in the engine namespace when called from within an engine.

Check the backtrace to make sure it's not coming from your app, and also make sure you're on the latest Rails 4.2.x patch release.

PR #961, opened 2022-10-08
fix thredded routes when main app has include url_helpers

### LINKED ISSUE #960: problem generating routes within thredded if application helpers include url_helpers
We've just installed thredded and are seeing an odd behaviour on URLs.  Every link from the forum site, with the exception of the devise/user ones, looks like this:  
`http://localhost:3000/forum/assets?action=index&controller=thredded%2Fmessageboards`
or
`http://localhost:3000/forum/assets?action=new&controller=thredded%2Fmessageboards` 
with `assets` inserted in the url instead of something more like the demo site's
`https://thredded.org/thredded/messageboards/new`

This is thredded-1.0.1, ruby 2.7.2, rails 6.0.3.2

The routes are configured:
```
Fromthepage::Application.routes.draw do
  # TODO make the URL fall under user and collection profile
  mount Thredded::Engine => '/forum'
```

```
                     new_messageboard GET      /messageboards/new(.:format)                                         thredded/messageboards#new
               show_messageboard_group GET      /messageboard-groups/:id(.:format)                                   thredded/messageboard_groups#show
                     edit_messageboard GET      /messageboards/:id/edit(.:format)                                    thredded/messageboards#edit
                          messageboard PATCH    /messageboards/:id(.:format)                                         thredded/messageboards#update
                                       PUT      /messageboards/:id(.:format)                                         thredded/messageboards#update
                                       DELETE   /messageboards/:id(.:format)                                         thredded/messageboards#destroy
         edit_messageboard_preferences GET      /:messageboard_id/preferences/edit(.:format)                         thredded/preferences#edit
....
```

You can see all of our changes here:  https://github.com/benwbrum/fromthepage/pull/3345/files

The only thing we changed in the initializer was `Thredded.user_name_column = :login`

Any suggestions?

timdiggins:
if main app has `include Rails.application.routes.url_helpers` then the 
thredded routes get broken.

In addition we now have a way to run "configuration specs" which mess with the
dummy app in some way, but are run in isolation to each other (each spec file is run
separately in isolation, with a spearate change to the dummy)

This fixes #960
may also address issue underlying #686


timdiggins:
@glebm 👋 
I don't really understand why adding the line in a main app helper causes the original problems, but I think our problems are related to rendering in a renderer - which already has some interesting unresolved issues in rails (https://github.com/rails/rails/issues/34452), but in this case it looks like it might be on us to fix, rather than a rails issue.. 

Let me know what you think of the configuration spec construct (wording / implementation)

glebm (approved):
Looks good overall! I also have no idea why it breaks with url_helpers included or why including them here fixes it.

PR #971, opened 2023-02-17
fix for incorrect post content output when multi-threaded render

### LINKED ISSUE #965: different posts in same topic are rendered the same
I'm getting something similar to #839 happening when caching is enabled (default in production) and multi-threaded rendering is enabled (default in Thredded). 

At the moment it seems deterministic but not predictable (which is fun) - ie. certain topics will have one or two or three duplicated posts (the same ones) and other ones won't.

I haven't worked out yet how this quite works, but I have detected that the cache key being generated in Rails 6.0 and 6.1 (not tested in Rails 7 yet) doesn't use the post's `id`, but instead uses something like a URL for the post - but (I think) because the URL doesn't exist, it uses a url based on the hash of the content of the post : e.g. `thredded/posts/_content:fe038035150f380ed06f9e0ae6b255bf`.

This all happens in `Thredded::CollectionToStringsWithCacheRenderer`: 

* `collection_by_cache_keys` calls `expanded_cache_key` with the `item` (ie. the post) as the first arg.
   https://github.com/thredded/thredded/blob/main/lib/thredded/collection_to_strings_with_cache_renderer.rb#L62

       key = expanded_cache_key(item, view, template, digest_path)

* `expanded_cache_key` calls this arg the `key`
   https://github.com/thredded/thredded/blob/main/lib/thredded/collection_to_strings_with_cache_renderer.rb#L68
   
       def expanded_cache_key(key, view, template, digest_path)

* this then calls `cache_fragment_name` to get the cache name (partial key), which eventually delegates this to the view
  where it leaves Thredded code and calls a flavour of url_for
  

I'm not 100% sure how to test for this or fix this, but it strike me it might be better to call item.to_param before sending it to expanded_cache_key, ie. `expanded_cache_key(item.to_param,...`

@glebm could do with your brainpower on this if you have any spare cycles

timdiggins:
fixes #965

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,6 +110,7 @@ script:
 - SIMPLECOV_NAME="test:js" bundle exec rspec --force-color --format d --tag=js
 - MIGRATION_SPEC=1 SIMPLECOV_NAME="test:migrations" bundle exec rspec spec/migration --force-color --format d
 - spec/configuration/run_all
+- bin/run_specs_repeatedly --tag=threaded_render
 after_script:
 - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
 
--- a/bin/run_specs_repeatedly
+++ b/bin/run_specs_repeatedly
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+# exit on first failure
+set -e
+for i in {1..10}
+do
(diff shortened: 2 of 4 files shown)


glebm (approved):
Nice!

PR #973, opened 2023-02-18
fix when there are duplicate posts to render

### LINKED ISSUE #972: render_posts with duplicate posts (same id)
Potential issue with Thredded::CollectionToStringsWithCacheRenderer,  which would happen if you try to render_posts with a collection with the same post more than once in the list ie. something like: `render_posts([cached_post, new_post_1,new_post_1, new_post_2])`. This is never going to appear in vanilla rails, but I guess we should fix it anyway. 

If correct, should be straightforward to write a test now based on work in #971 but the fix may be a bit more complex.

timdiggins:
fixes #972


timdiggins:
@glebm I'm going to merge this so I can do a release, but please shout if you spot anything that needs tweaking.

Issue #979, opened 2024-03-03
 Cannot load such file -- html/pipeline (LoadError)


kvoshell:
Attempting to add `thredded` to an existing app (Rails 7.0.0, ruby 3.2.0) throws a LoadError

```
➜  rails generate thredded:install
/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require': cannot load such file -- html/pipeline (LoadError)
	from /.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:17:in `require'
	from /.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/zeitwerk-2.6.6/lib/zeitwerk/kernel.rb:38:in `require'
	from /.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/thredded-1.1.0/lib/thredded.rb:8:in `<main>'
	from /.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require'
	from /.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/bootsnap-1.16.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require'
	from /.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/zeitwerk-2.6.6/lib/zeitwerk/kernel.rb:38:in `require'
	from /.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.5/lib/bundler/runtime.rb:60:in `block (2 levels) in require'
	from /.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.5/lib/bundler/runtime.rb:55:in `each'
	from /.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.5/lib/bundler/runtime.rb:55:in `block in require'
[...]

PR #981, opened 2024-04-07
Restrict html_pipeline version to <3

glebm:
html_pipeline v3 changes its underlying parser from Nokogiri to Selma, which means we'd have to rewrite all of the HTML pipeline / sanitization code.

Refs #979


rickychilcott:
LGTM. It appears it'd be a lot of work to switch over. That's a shame, but totally get wanting to embark on that project.

glebm:
Selma appears to be less powerful, e.g. it doesn't know how to sanitize `style` tags. In Thredded we allow `text-align` (and no other CSS attributes) in `style` to fully support markdown tables.

Really they should've released v3 as a different gem.

Issue #986, opened 2024-07-18
Switch to GitHub Actions


rickychilcott:
Given https://github.com/thredded/thredded/pull/985#issuecomment-2234305156, switch to GitHub Actions (from Travis) so we can get proper testing in place again.

rickychilcott:
I'm happy to take a stab.

timdiggins:
Thanks @rickychilcott for this - great work!

PR #987, opened 2024-07-18
Switch to GitHub Actions

rickychilcott:
Use a matrix-based approach to test Thredded. Ideally, this would test the following:

- Latest Ruby versions
- Highest Rails version (including edge)
- With each major database provider (postgres, mysql, and sqlite)
- Node version (currently only testing with version 14)

Addresses #986


rickychilcott:
@glebm you may [need to turn on GHA](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository) on the repository.

rickychilcott:
Nevermind. I think it's running...

glebm:
The actions are turned out but I don't see them running, odd

rickychilcott:
They're running on [my repo](https://github.com/mission-met/thredded/actions) right now. I think once this gets merged, it'll run on this repo

timdiggins:
@rickychilcott brilliant - thanks for starting on this. 

BTW I think we could drop rails 6.0 support as 6.0 has EOL'd (only if that helps!)

.github/workflows/build.yml
```diff
@@ -0,0 +1,165 @@
...
+    #   uses: actions/cache@v2
+    #   with:
+    #     path: |
+    #       tmp/cache/onebox-data
+    #       tmp/cache/onebox-views
+    #       spec/dummy/tmp/cache/onebox-data
+    #       spec/dummy/tmp/cache/onebox-views
+    #     key: onebox-${{ runner.os }}-${{ matrix.ruby }}-${{ matrix.gemfile }}
+    #     restore-keys: |
+    #       onebox-${{ runner.os }}-${{ matrix.ruby }}-${{ matrix.gemfile }}
```
rickychilcott, line 155: It's not clear what this is. Do you think we still need this? It's been converted over from Travis

.github/workflows/build.yml
```diff
@@ -0,0 +1,165 @@
...
+
+    # - name: Notify Slack
+    #   uses: 8398a7/action-slack@v3
+    #   if: always()
+    #   with:
+    #     status: ${{ job.status }}
+    #     author_name: GitHub Actions
+    #     fields: repo,message,commit,author,action,eventName,ref,workflow,job,took
+    #     env:
+    #       SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
```
rickychilcott, line 165: Not sure if you all still maintain a slack for this project. Ok to remove?

.github/workflows/build.yml
```diff
@@ -0,0 +1,165 @@
...
+        spec/configuration/run_all
+
+    - name: "Run Tests: run_spec_repeatedly"
+      run: |
+        bin/run_specs_repeatedly --tag=threaded_render
+
+    # - name: Test & publish code coverage
+    #   uses: paambaati/codeclimate-action@v8.0.0
+    #   env:
+    #     CC_TEST_REPORTER_ID: <code_climate_reporter_id>
```
rickychilcott, line 143: If we want to get CodeClimate working again, then we'll need to determine the CC_TEST_REPORTER_ID: See https://docs.codeclimate.com/docs/finding-your-test-coverage-token

rickychilcott:
Happy Friday @glebm and @timdiggins! This appears to be passing and tests Rails 6.1+, all major ruby versions, and against all three database engines. We also have Rubocop and i18n-tasks checks in place.

I left a few comments that I'm looking for feedback on and then I think we can ship it.

See the CI runs at https://github.com/mission-met/thredded/actions. https://github.com/mission-met/thredded/actions/runs/10012043134 is the latest run that was successful.

I've experienced some flakiness on individual runs, most notably  "Browser did not produce websocket url within 10 seconds, try to increase `:process_timeout`" issues. Which [cac0a20](https://github.com/thredded/thredded/pull/987/commits/cac0a208ae031ad45e1eab3ac1b015fad1b0ca68) may or may not be improving.

.github/workflows/build.yml
glebm: The tests contain onebox embeds to various websites, the renderer fetches that content. If an external website is being slow to respond or simply down, that can lead to test flakiness, so we cached the onebox responses.

This can be addressed in a follow-up.

https://github.com/discourse/onebox

.github/workflows/build.yml
glebm: Don't think there is a slack anymore, OK to remove

glebm:
Fantastic work, thank you! The timeout issues could be due to the lack of onebox cache.

.github/workflows/build.yml
rickychilcott: Got it. I think I've got restore/caching working now.

rickychilcott:
No problem, @glebm. Thank you for all of your work on this over the years. I think we're ready to merge.

PR #988, opened 2024-08-21
Deprecated alias_attribute for a method

Juleffel:
In rails 7.2 there is a bug with `postable` when accessing forums.

Seems to be a simple deprecation of `alias_attribute postable, topic` line, changing this fix the problem.

image

image


timdiggins:
Thanks very much for the PR @Juleffel I'm working on rails 7.2 compatibility now and I can see this problem coming up in the CI

PR #989, opened 2024-08-30
Replace the hard sassc-rails gem dependency with a soft dependency on a Sprockets Sass compiler

botandrose:
Hi, thanks for thredded!

Now that Ruby Sass and SassC are dead, we've been switching our Rails projects over to use the Dart Sass compiler in the form of the `dartsass-sprockets` gem. However, Thredded has a hard dependency on `sassc-rails`, and the two cannot coexist.

Here's my five-minute hack to relax the dependency. It seems to work well enough, but I'm open to other resolution strategies here.

Thoughts?

### DIFF
--- a/lib/thredded.rb
+++ b/lib/thredded.rb
@@ -29,7 +29,6 @@ require 'thredded/users_provider'
 require 'autoprefixer-rails'
 require 'timeago_js'
 require 'sprockets/es6'
-require 'sassc-rails'
 
 require 'thredded/version'
 require 'thredded/compat'
--- a/lib/thredded/engine.rb
+++ b/lib/thredded/engine.rb
@@ -22,8 +22,11 @@ module Thredded
     end
 
     config.after_initialize do |app|
-      if !Rails.env.production? && !app.assets.preprocessors.keys.include?("text/scss")
-        raise %(Thredded requires a Sass compiler to be registered in Sprockets. Please add "sassc-rails" or "dartsass-sprockets" to your application Gemfile.)
+      if !Rails.env.production? && app.assets.preprocessors.keys.exclude?('text/scss')
+        fail [
+          'Thredded requires a Sass compiler to be registered in Sprockets.',
+          %(Please add "sassc-rails" or "dartsass-sprockets" to your application Gemfile.),
+        ].join(' ')
       end
     end
   end
--- a/lib/thredded/engine.rb
+++ b/lib/thredded/engine.rb
@@ -20,5 +20,11 @@ module Thredded
         thredded_manifest.js
       ]
     end
+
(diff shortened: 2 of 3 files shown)


glebm:
Is this sufficient, or do we need to do something like this? https://github.com/twbs/bootstrap-rubygem/blob/f1f92e0cb077d9bbfe0986fd1184a0560d022aa0/lib/bootstrap/engine.rb#L3-L19

botandrose:
Hi @glebm , thanks for taking a look at this. And its very interesting to see an alternative approach. From my personal perspective, either approach would enable me to stop using my fork of the gem, so thumbs up to either.

But thinking at it from a maintenance perspective, there seem to be some interesting tradeoffs and questions.

It strikes me that this PR's strategy is akin to "duck-typing" the sass dependency, while `bootstrap-rubygem` is "static-typing" it.

It seems to me that the latter approach provides stronger guarantees that the dependency actually exists. But also requiring some manual maintenance of exactly which dependencies are permitted as the landscape evolves. We'd probably also need to test the integration with each sass engine, too... I'm not sure they all work the same way with assets in an engine's `app/assets/stylesheets` folder. `dartsass-rails` and `cssbundling-rails`, for example, both require an external process to be manually run to compile the sass (which I personally find objectionable). I'm not exactly sure how that works (or not) with assets provided by an engine.

This PR is a little more conservative in that it merely asserts that the Rails app CAN compile sass via sprockets, broadening support from just `sassc-rails` to at least both `sassc-rails` and `dartsass-sprockets`. I honestly don't know whether or not the other sass gems would pass that assertion! If they don't, then maybe this approach isn't sufficient, because you'd just get another followup PR requesting changes for them. But maybe we cross that bridge if and when we come to it?

glebm:
Sounds good to me. Please address the rubocop warnings and I'll merge.

How come some test suites are passing even though there is no longer a Sass compiler included in the test app?

botandrose:
@glebm Okay, Rubocop has been appeased.

I was also surprised to see the test suites passing (ignoring some unrelated flakiness). It turns out that the dummy app is installing the gem's development dependencies, which includes `rails_email_preview`, which declares a dependency on `sassc-rails`: https://rubygems.org/gems/rails_email_preview .

glebm:
Thanks! We should probably add an explicit dependency on a Sass compiler to the test Gemfiles because the next version of rails_email_preview also no longer depends on a particular Sass compiler https://github.com/glebm/rails_email_preview/commit/a446fc651c09979aa7d40b56e2afc68c22e6f074

timdiggins:
👍 And I guess we should mention this soft dependency on a sprockets sass compiler in the README, with a link to the different options.

PR #993, opened 2024-11-05
support rails edge (8.1)

timdiggins:


romanbsd:
@timdiggins what do you think about this approach? https://github.com/thredded/thredded/pull/1016

timdiggins:
@romanbsd thanks for the PR - it will take me a few days before I can have a look at this.

PR #995, opened 2024-11-07
use caches for yarn in CI

timdiggins:


timdiggins:
Not really possible as we don't generate a yarn.lock in spec/dummy

PR #999, opened 2025-05-07
Ukrainian I18n file

mykbren:
🇺🇦 Added localization file


glebm (approved):

timdiggins:
Thanks for the PR @mykbren

(Fails with Rails 7.0 are nothing to do with this, and I'll address in a PR separately)

PR #1001, opened 2025-05-22
Docker compose fix

A-Yerkeshev:
- Alpine was upgraded to 3.20;
- NodeJS was pinned to v14.21.3, because recent versions are not compatible with package versions in spec/dummy/package.json;
- libressl-dev was removed because it conflicted with openssl, required by other packages;
- yaml-dev added because it's required by psych gem;
- spec/gemfiles are copied;
- node-sass v14.4.1 was explicitly required because it has pre-built binaries for Alpine;
- tmp/pids/server.pid is removed before every run, so that it's possible to restart application with `docker compose up` after Ctrl+C;

### DIFF
--- a/script/docker-dev-start-web.sh
+++ b/script/docker-dev-start-web.sh
@@ -16,4 +16,6 @@ if ! [[ -f .db-seeded ]]; then
   touch .db-seeded
 fi
 
+rm -f tmp/pids/server.pid
+
 foreman start -f Procfile.docker.dev


timdiggins:
Hey thanks for this -- annoying we have to use such out of date versions of NPM to support the dummyapp -- shows the dummyapp needs upgrading. I couldn't get it to run on my silicon Mac (arm) - and I don't have the current lifeforce to address this (though running the tests via docker would be handy for me, but I'd want all database support for that). 

Anyway looks good and I'll merge shortly (hoping to get CI clean first)

timdiggins:
Thanks again @A-Yerkeshev. I've got it working on ARM now - I'll merge yours then merge the changes

PR #1004, opened 2025-05-24
drop active_record_union, with arel replacement

timdiggins:
working for 8.0 with postgresql
not working with sqlite


timdiggins:
@glebm yes exactly. I've extracted the code from active_record_union and simplified

glebm (approved):

PR #1013, opened 2025-07-10
Modify ApplicationHelper#time_ago to display time in server's timezone

A-Yerkeshev:
Hi!

I faced the issue in our application that all times are one hour more than they should be. Perhaps this is related to daylight saving time. Didn't anyone face this issue before?
 
Screenshot from 2025-07-10 08-42-21

### DIFF
--- a/app/helpers/thredded/application_helper.rb
+++ b/app/helpers/thredded/application_helper.rb
@@ -67,6 +67,7 @@ module Thredded
     # @return [String] html_safe datetime presentation
     def time_ago(datetime, default: '-', html_options: {})
       return content_tag :time, default if datetime.nil?
+      datetime = datetime.localtime
       html_options = html_options.dup
       is_current_year = datetime.year == Time.current.year
       if datetime > 4.days.ago


timdiggins:
Hi @A-Yerkeshev thanks for looking at this. It's true that the Thredded dummy app is set to UTC.

But I can't repro your problem in my main app (which is set to London and is on BST - summer time, right now). 

```ruby
class Tester; include ActionView::Helpers; include  Thredded::ApplicationHelper; end

Time.now.utc
=> 2025-07-11 14:42:16.767637 UTC
Time.zone.now
=> Fri, 11 Jul 2025 15:42:21.567723000 BST +01:00
Tester.new.time_ago(1.minute.ago)
=> "<time datetime=\"2025-07-11T15:41:29+01:00\" data-time-ago=\"true\" title=\"Fri, 11 Jul 2025 15:41:29 +0100\">1 minute ago</time>"
```

In general I can't see the purpose of calling localtime on a TimeWithZone (which is what rails mostly uses/returns).
https://api.rubyonrails.org/classes/ActiveSupport/TimeWithZone.html#method-i-localtime - unless you are changing the timezone of a specific instance and want the server's timezone (ie you are doing something a bit exotic)

What time zone is your main app set to (`Rails.application.config.time_zone`)? 
And is your timezone data up to date (either system or https://github.com/tzinfo/tzinfo-data)?

A-Yerkeshev:
My timezone is UTC and up to date. But yes, there is probably something else wrong with my setup if nobody reported this before. Thanks for the explanation!

PR #1016, opened 2025-11-27
Fix for Rails 8.1 counter cache change

romanbsd:
This is my take on the counter cache issue in Rails 8.1

Feedback appreciated!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,17 +56,17 @@ jobs:
       fail-fast: false
       matrix:
         node: ["22"]
-        rails_version: ["7.0", "7.1", "7.2", "8.0"]
+        rails_version: ["7.1", "7.2", "8.0", "8.1"]
         db: [sqlite3, postgresql, mysql2]
         include:
           # one ruby per rails - to cut down noise, we can add variety if there's a reason
-          - rails_version: 7.0
-            ruby: 3.1
           - rails_version: 7.1
             ruby: 3.2
           - rails_version: 7.2
             ruby: 3.2
           - rails_version: 8.0
+            ruby: 3.4
+          - rails_version: 8.1
             ruby: 3.4
 
     env:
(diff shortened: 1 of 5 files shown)


romanbsd:
Removed Rails 7.0, added 8.0 to matrix.

romanbsd:
oh, right, it cannot build because of node-sass...

timdiggins:
> oh, right, it cannot build because of node-sass...

feel free to rebase this one on your other PR #1017 (which should get it green) -- I will review this one once that is merged...

timdiggins:
I'm not 100% sure we should be working round this only in the TopicForm - it leaves it impossible to create a topic in the console. (I think you need to copy in this spec:  https://github.com/thredded/thredded/pull/993/changes#diff-9cf83f58efcb5c0700e5dd5047f98419dd8f188bd73cd7038605f56580886a92 and then fix it one way or another. 

I'm thinking maybe the best route is to drop the unnecessary extra foreign key (as mentioned in the linked spec) -- we should be able to just allow post → messageboard via a `through: :postable` (it's just the code is quite confusing because of some over-DRYing of post/private post)

----

If we go ahead with this one, it needs a rebase now on main, but also the rails 8.1 gemfile isn't quite right. Here are my tweaks.

```
gem 'rails', '~> 8.1.1'
gem 'rails-i18n', '~> 8.1.0'

# TODO: create a new dummy using import-maps ?
gem 'webpacker', '~> 5.0'

# https://github.com/rails/rails/blob/v8.1.1/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L14
gem 'sqlite3', '>= 2.1'

# https://github.com/rails/rails/blob/v8.1.1/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L3
gem 'pg', '~> 1.1'

# https://github.com/rails/rails/blob/v8.1.1/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb#L6
gem 'mysql2', '~> 0.5'
```

romanbsd:
I rebased meanwhile and fixed another issue

PR #1017, opened 2025-11-27
Migrate from Turbolinks and UJS to Hotwire Turbo

romanbsd:


timdiggins:
@romanbsd Great work. Haven't reviewed in depth, but this is looking good. It breaks the tests, because the package.json and the js config in dummy / github is wildly out of date.

* [x] Update node (in github build) to something more recent
* [ ] ~~Update rails and other packages in dummy package.json (not sure how we manage this with testing against multiple rails - but hopefully using the latest will still work)~~
* [ ] ~~Update the js build mechanism: Simplest thing is probably to upgrade from webpacker to shakapacker as that requires the least changes~~

These could be in a separate PR from this change, but this PR would depend on these changes to run the spec.

spec/dummy/config/initializers/assets.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+Rails.application.config.assets.precompile += %w[email.css day.css night.css favicon.ico]
```
timdiggins, line 3: interested in why this was necessary - these should be already ok - because linked from manifest.js

spec/dummy/package.json
```diff
@@ -2,14 +2,13 @@
   "name": "dummy",
   "private": true,
   "dependencies": {
-    "@rails/actioncable": "^6.0.0-alpha",
-    "@rails/activestorage": "^6.0.0-alpha",
-    "@rails/ujs": "^6.0.0-alpha",
-    "@rails/webpacker": "^4.0.7",
+    "@hotwired/turbo-rails": "^8.0.20",
+    "@rails/actioncable": "^6.1.710",
+    "@rails/activestorage": "^6.1.710",
```
timdiggins: Given we're not using actioncable or activestorage in Thredded, I think we can remove these dependencies from the dummy

spec/dummy/package.json
```diff
@@ -2,14 +2,13 @@
...
+    "@rails/webpacker": "^5.4.4",
```
timdiggins, line 6: I don't think this works any more. But should be straightforward to upgrade to Shakapacker.

spec/dummy/package.json
```diff
@@ -2,14 +2,13 @@
...
     "rails-erb-loader": "^5.5.2",
-    "turbolinks": "^5.2.0",
-    "webpack": "^4.0.0",
-    "node-sass": "^4.14.1"
+    "sass-embedded": "^1.79.4",
```
timdiggins: Think `sass` might be more universally compatible (I've heard of cross compilation with sass-embedded issues on macos?) - and performance benefit is negligible in specs

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 
 # Unreleased (plans for 1.3.0)
 
+* Remove the remaining Rails UJS integration in favor of Turbo-powered helpers.
+* Migrate from Turbolinks to Hotwire Turbo
```
timdiggins, line 9: This will be a breaking change for users who are on Turbolinks so we should release as 2.0.

timdiggins: Mention also "drop support for jquery-ujs"

timdiggins (changes requested):
Great work. A few changes required, and above all need to get the specs running again. 

(though to be fair, they were broken already)

spec/dummy/config/initializers/assets.rb
romanbsd: I'm not sure myself, perhaps some gem change made this necessary, but the dummy app wasn't finding them otherwise.

spec/dummy/package.json
romanbsd: it actually does. The specs pass.

spec/dummy/package.json
romanbsd: I changed to sass-embedded not because of performance, but because it was impossible to build node-sass anymore (modern node and python are incompatible).

CHANGELOG.md
romanbsd: it's not that the support is dropped, it's not needed anymore.

spec/dummy/package.json
romanbsd: removed

spec/dummy/package.json
timdiggins: Totally agree - but `sass` is a separate package that relies on the standard dart-sass implementation. `sass-embedded` is a recompilation of dart-sass into js.

CHANGELOG.md
timdiggins: that's true, but we drop the configuration to allow people to use jquery-ujs instead of rails-ujs (some existing users may be still using jquery-ujs rather than rails-ujs - potentially)

spec/dummy/package.json
timdiggins: Great to hear it's working on your machine, maybe we can get it to work on CI as is when upgrade node

romanbsd:
All specs green. I rebeased the code.

spec/dummy/package.json
romanbsd: all specs pass now

spec/dummy/package.json
romanbsd: changed to sass

CHANGELOG.md
romanbsd: added note of it being removed

app/views/thredded/topics/show.html.erb
```diff
@@ -39,7 +39,7 @@
         <%= button_to t('thredded.topics.delete_topic'), topic.destroy_path, method: :delete,
                       form_class: 'thredded--topic-delete-form',
                       class: 'thredded--button',
-                      'data-confirm' => t('thredded.topics.delete_confirm') %>
+                      'data-turbo-confirm' => t('thredded.topics.delete_confirm') %>
```
timdiggins, line 42: I finally get what you were saying about the "no need for rails-ujs". This PR if I read it correctly, makes `Turbo` mandatory and `rails-ujs` isn't a valid replacement. 

If we're going to do that, then we need to drop the !Turbo parts and also make that requirement (`turbo-rails`?) mandatory in the README. 

Alternatively we allow a configuration (if needed - perhaps it can be auto-detected) for rails-ujs vs Turbo (and output `data-turbo-confirm` vs `data-confirm` etc differentially). And then announce that we're not going to support rails-ujs for ever (maybe until we drop rails 7.2). 

I'm leaning towards the second option.  But we could test with turbo only (as we're going to drop rails-ujs and so we don't have to build two spec/dummy uses don't need to be tested)

* v2.0: add support for turbo, drop support for turbolinks, jquery-ujs (and drop rails 7.0, 7.1)
* v3.0: drop support for rails-ujs and make turbo mandatory (and drop rails 7.2) - some time in August 2026 https://endoflife.date/rails

thoughts? @romanbsd, @glebm, @rickychilcott? (Also anyone else who happens to see this).

lib/thredded/webpack_assets.rb
```diff
@@ -100,8 +100,6 @@ def dep_to_path(dep)
           path = File.join(Gem.loaded_specs['timeago_js'].full_gem_path, 'assets', 'javascripts', "#{dep}.js")
           fail "Failed to find #{dep}" unless File.file?(path)
           return path
-        elsif dep == 'rails-ujs'
-          return '@rails/ujs'
```
timdiggins, line 104: see above

CHANGELOG.md
timdiggins: Sorry I only just got what you meant - see longer comment above

timdiggins (commented):
Not sure how we handle making turbo-rails required for end users. See comment in the PR.

app/views/thredded/topics/show.html.erb
glebm: I think the gem has been stable for long enough that we can simply require Rails 8 for the next version.
People on Rails 7 can still use the older version, it's not like features are being added actively.

As for Turbo vs rails-ujs, it looks like rails-ujs is a part of Rails and so should be supported?
https://github.com/rails/rails/blob/main/actionview/app/assets/javascripts/rails-ujs.js
Looks like a separate gem is not required and it's simply part of actionview these days

Disclaimer: I don't do Rails anymore so don't know what the current state of the ecosystem is.

glebm: I now see that Turbo is recommended over UJS by the official Rails guide, so no objections to dropping support for UJS https://guides.rubyonrails.org/working_with_javascript_in_rails.html#replacements-for-rails-ujs-functionality

romanbsd: Just note, that rails_ujs is long time gone from Rails, so data-confirm won't work unless you require users to use the obsolete rails_ujs. Rails 7 removed rails-ujs from new application defaults because Rails 7 switched from Turbolinks + rails-ujs to Hotwire (Turbo + Stimulus). Rails 7.0 is EOL already BTW.

timdiggins: Agreed, good points.  The gem is stable enough, no major new features -- so leave v1.2.1 with support for rails 7.0, 7.1, 8.0 with rails-ujs, jquery-ujs, and/or turbolinks. v2.x will support v8.0, v8.1 and turbo-rails is mandatory. I think we could include turbo-rails as a gem dependency without any version requirements (let turbo-rails handle the version requirements itself). Nice plan. 

I'd like to get this running on my machine and do a manual test before merging.

timdiggins:
@romanbsd just checking in to let you know I haven't forgotten this - it's just been very busy. Hope to do this next weekend.

romanbsd:
ok, ping me if anything needs changing, etc

timdiggins (approved):
Great work - (finally got every spec passing on my machine -- some kind of local caching issue). I'll merge this and make a new PR for README changes.

